PR #35, opened 2021-04-18
Fixed the `dt` in the burgers data generation script

scaomath:
@zongyi-li 

https://github.com/zongyi-li/fourier_neural_operator/blob/73854fa9bcf75229f2871e0d38b80ac30586859a/data_generation/burgers/burgers1.m#L1-L7

In the Burgers data generation file, `spin` is called with a fixed `dt = 1e-4`. However, it actually should be the time step given by the `tspan`. I know that you are working on cleaning the code, but this PR's change is so minimal that merging with other branches should be no problem after rebasing.

### DIFF
--- a/data_generation/burgers/burgers1.m
+++ b/data_generation/burgers/burgers1.m
@@ -1,8 +1,9 @@
 function u = burgers1(init, tspan, s, visc)
 
 S = spinop([0 1], tspan);
+dt = tspan(2) - tspan(1);
 S.lin = @(u) + visc*diff(u,2);
 S.nonlin = @(u) - 0.5*diff(u.^2);
 S.init = init;
-u = spin(S,s,1e-4,'plot','off'); 
+u = spin(S,s,dt,'plot','off'); 
 


zongyi-li:
Thanks for the catch!

zongyi-li:
Oh, I guess in the original code, `dt` is the internal timestep, while `tspan` is the sample rate, so they can be different. You can have a coarser/ finer output compared to the `dt` that solves the equation.

scaomath:
> Oh, I guess in the original code, `dt` is the internal timestep, while `tspan` is the sample rate, so they can be different. You can have a coarser/ finer output compared to the `dt` that solves the equation.

Oh that makes sense. I found this when I tried to dissect the code to generate solutions other than FFT expansion of GRFs.
Viscosity set to `0.1` seems way too big for solution not to decay to 0, and never developed a shock. Please check the following notes.



If the `dt` was default to `1e-4` and `nu=0.1`, matching toward `t=1` seeming kills the wave every time.
[https://sites.wustl.edu/scao/files/2021/04/burgers_v1e-1.webm](https://sites.wustl.edu/scao/files/2021/04/burgers_v1e-1.webm)

zongyi-li:
Yes, the viscosity `nu` control the dissipation of energy. For `nu=0.1,` the energy decays quite fast. If you want shocks, you can try with `nu=0.001`. With `N=1000` training samples, the FNO will get round <1% error. Hope I understand your question correctly.

scaomath:
> Yes, the viscosity `nu` control the dissipation of energy. For `nu=0.1,` the energy decays quite fast. If you want shocks, you can try with `nu=0.001`. With `N=1000` training samples, the FNO will get round <1% error. Hope I understand your question correctly.

Thanks for the reply. If the params are changed to `visc=1e-1` and `gamma = 2`, `tau = 5`, `sigma = 25`, the solution at time 1 will be almost zero. 

So I suspect that this block might return `u(x, 1e-4*(steps+1))`:
https://github.com/zongyi-li/fourier_neural_operator/blob/a7c8fa3fe00c5fc377c7ac52454b076e0e204197/data_generation/burgers/gen_burgers1.m#L36-L38

I do not mean to undermine your result, the 1% error is so impressive in that it beats previous methods (with this "collocation point" cheating method) by magnitudes which puts me in awe. Just I found this weirdness when I tried to regenerate the dataset `R10` using a fixed seed.

zongyi-li:
Thanks for the kind words. Sorry I may not understand the question very well. For `nu=0.1`, the energy will decay quite fast and therefore it's kind of expected that the solution will be zero at time 1. Let me know if I missed anything.

zongyi-li:
Hi Shumao, 
I realized the reason you may get 0 at t=1 using these parameters `visc=1e-1 and gamma = 2, tau = 5, sigma = 25.`

The original data is generated on domain (0, 2pi) but the script has domain (0, 1). To fix that you may need to 
`change x = linspace(0,1,s+1); `
To 
`x = linspace(0,2*pi,s+1); `

Or equivalently you can scale `visc = 1e-1` to `visc = 1e-1/(2*pi)`
Sorry I didn't understand the question eariler!

PR #37, opened 2021-05-04
Fixed a tiny bug in SpectralConv2d

scaomath:
https://github.com/zongyi-li/fourier_neural_operator/blob/70f469fef815bad0e07ea4b487b10cdcf3c50b30/fourier_2d.py#L56

The `out_ft` variable should have `out_channels` as the dimension as the complex matmul's output dim is `(B, C_out, W, H)`.

This would pose no change on the current FNO2d net but potentially gives people freedom to use a bottleneck structure.


zongyi-li:
Thanks, that's correct.

PR #43, opened 2021-05-22
Fixed tensor dimension issue during 1d Burger's prediction

j-griffiths:
Fixed the prediction stage of `fourier_1d.py` to return a tensor with all dimensions of size `1` eliminated (using the `squeeze()` method). Without this, it currently returns an error immediately after training and before prediction has started.


zongyi-li (approved):

zongyi-li:
Thanks, this bug is known, but I haven't got a chance to fix it.

PR #79, opened 2022-08-30
Fixed the spatial random sampling in UnitGaussianNormalizer

scaomath:
In the original repo, `sample_idx` was supposed to be a spatial masking to enable PINN-like loss. However, there is a tiny bug as the mean/std shape can be either in `(T, n, n)` or `(n, n, T)` (default behavior in FNO repo). So an extra switch `self.time_last` is added.

Currently no routine that has `decode` in the current repo uses this arg so it should not affect existing codes.

Also a `.to()` method is added to match the behavior of torch.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,126 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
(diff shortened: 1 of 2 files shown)


zongyi-li (approved):
thanks, it looks good to me!

PR #106, opened 2023-03-23
Uno

ashiq24:


neuralop/models/fno_block.py
```diff
@@ -1,10 +1,39 @@
 from torch import nn
 import torch.nn.functional as F
-
+import torch
 from .spectral_convolution import FactorizedSpectralConv
 from .skip_connections import skip_connection
 from .mlp import MLP
 
+def resample(x, res_scale, axis):
```
JeanKossaifi, line 8: Can you add a docstring?

neuralop/models/fno_block.py
```diff
@@ -1,10 +1,39 @@
...
+    if isinstance(axis, list) and isinstance(res_scale, (float, int)):
```
JeanKossaifi, line 9: Could be a tuple -- you'd want to check if it's iterable instead. 
Here, let's just check that they are both int or not:
```python
if isinstance(res_scale, (float, int)):
   if isinstance(axis, (float, int)):
       axis = [axis]
       res_scale = [res_scale]
    else:
        res_scale = [res_scale]*len(axis)
else:
    assert len(res_scale) == len(axis), error_msg
```

neuralop/models/fno_block.py
```diff
@@ -1,10 +1,39 @@
...
+
+    Nx = x.shape[axis]
+    X = torch.fft.rfft(x, dim=axis, norm = 'forward')    
+    newshape = list(x.shape)
+    new_res = int(res_scale*newshape[axis])
+    newshape[axis] = new_res // 2 + 1
+
+    Y = torch.zeros(newshape, dtype=X.dtype, device=x.device)
+
+    N = min(new_res, Nx)
```
JeanKossaifi, line 31: It's good to try to name variable with informative names (e.g. avoid Nx etc) and be consistent (e.g. `new_shape` vs `newshape`)

neuralop/models/padding.py
```diff
@@ -16,10 +16,11 @@ class DomainPadding(nn.Module):
     This class works for any input resolution, as long as it is in the form
     `(batch-size, channels, d1, ...., dN)`
     """
-    def __init__(self, domain_padding, padding_mode='one-sided'):
+    def __init__(self, domain_padding, padding_mode='one-sided', output_scale_factor = 1):
```
JeanKossaifi, line 19:
```suggestion
    def __init__(self, domain_padding, padding_mode='one-sided', output_scale_factor=None):
```

neuralop/models/fno_block.py
JeanKossaifi: We also add a unit-test for it.

neuralop/models/tfno.py
```diff
@@ -39,6 +39,208 @@ def forward(self, x):
         x = self.fc2(x)
         return x
 
+class UNO(nn.Module):
```
JeanKossaifi, line 42: Let's put UNO in its own file, in `models.uno/py`

JeanKossaifi (changes requested):
Thanks @ashiq24! Left comments in the code.

neuralop/models/uno.py
```diff
@@ -0,0 +1,245 @@
...
+                skip_outputs[layer_idx] = self.horizontal_skips[str(layer_idx)](x)
+
+        if self.domain_padding is not None:
+            x = self.domain_padding.unpad(x)
+
+        x = self.projection(x)
+
+        return x
+
+def partialclass(new_name, cls, *args, **kwargs):
```
JeanKossaifi, line 221: Let's just import this from the other file.

neuralop/models/tests/test_uno.py
```diff
@@ -0,0 +1,31 @@
+import time
+from torchsummary import summary
```
JeanKossaifi, line 2: Let's not add additional requirements, we don't need to print the summary.

neuralop/models/padding.py
```diff
@@ -16,10 +16,11 @@ class DomainPadding(nn.Module):
...
         super().__init__()
         self.domain_padding = domain_padding
         self.padding_mode = padding_mode.lower()
+        self.output_scale_factor = output_scale_factor
```
JeanKossaifi, line 23: I would rather have the option to use None rather than 1: we avoid any rounding issue and just skip the resampling all together if it's None

neuralop/models/fno_block.py
```diff
@@ -1,10 +1,51 @@
...
+    if isinstance(res_scale, (float, int)):
+        if isinstance(axis, (float, int)):
+            axis = [axis]
+            res_scale = [res_scale]
+        else:
+              res_scale = [res_scale]*len(axis)
+    else:
+        assert len(res_scale) == len(axis), "leght of res_scale and axis are not same"
+    index = np.argsort(axis)
+    axis = [axis[i] for i in index]
```
JeanKossaifi, line 32: What is the goal of these list comprehensions?

JeanKossaifi (commented):
Thanks @ashiq24 - it's much cleaner, added a few more comments. As discussed, let's merge and iterate!

neuralop/models/fno_block.py
ashiq24: this is to sort the axis and their corresponding scaling factors. For example, the parameters passed are
axis = [-1,-3,-2]
res_scale = [1,2,3]
After these lines, it will be 
axis  = [-3,-2,-1]
res_scale = [2,3,1]

PR #109, opened 2023-04-01
solved resampling issue

ashiq24:


neuralop/models/tests/test_uno.py
```diff
@@ -2,28 +2,33 @@
...
-def test_UNO():
-    layer_configs = [{"out_channels":20, "n_modes" : [5,5], "res_scaling" :[0.5,0.5] },\
-                    {"out_channels":20, "n_modes" : [5,5], "res_scaling" :[1,1] },\
-                    {"out_channels":20, "n_modes" : [5,5], "res_scaling" :[1,1] },\
-                    {"out_channels":20, "n_modes" : [5,5], "res_scaling" :[2,2] },\
-                    {"out_channels":10, "n_modes" : [5,5], "res_scaling" :[2,2] },\
-                            ]
-    horizontal_skips_map ={4:0,3:1}
-    model = UNO(3,3,5,layer_configs = layer_configs, horizontal_skips_map = horizontal_skips_map, n_layers = 5, domain_padding = 0.2, output_scale_factor = 2)
+#def test_UNO():
```
JeanKossaifi, line 5: I am guessing reverting to this was accidental?

neuralop/models/padding.py
```diff
@@ -54,23 +54,37 @@ def pad(self, x):
             
             print(f'Padding inputs of {resolution=} with {padding=}, {self.padding_mode}')
 
+            out_put_pad = padding
```
JeanKossaifi, line 57:
```suggestion
            output_pad = padding
```
Let's just call it `output_pad`

neuralop/models/resample.py
```diff
@@ -55,7 +63,38 @@ def resample(x, res_scale, axis=None):
         idx_tuple = [slice(None), slice(None)] + [slice(*b) for b in boundaries]
 
         out_fft[idx_tuple] = X[idx_tuple]
-    y = torch.fft.irfftn(out_fft, norm='forward', dim = axis)
+    y = torch.fft.irfftn(out_fft, s =  new_size ,norm='forward', dim = axis)
+
+    return y
+
+
+def iterative_resample(x, res_scale, axis):
```
JeanKossaifi, line 71: There are quite a few pep8 issues, let's fix at least the obvious ones.

neuralop/models/resample.py
```diff
@@ -55,7 +63,38 @@ def resample(x, res_scale, axis=None):
...
+      raise Exception("Axis and Scal factor are in different sizes")
+
+    if isinstance(axis, list):
+        for i in range(len(res_scale)):
+            rs = res_scale[i]
+            a = axis[i]
+            x = resample(x, rs, a)
+        return x
+
+    Nx = x.shape[axis]
```
JeanKossaifi, line 86: It is better to use informative names,e.g `input_shape`

neuralop/models/tests/test_padding.py
```diff
@@ -1,5 +1,5 @@
 import torch
-from ..padding import DomainPadding
+from neuralop.models.padding import DomainPadding
```
JeanKossaifi, line 2: why did you change this? Let's revert to relative import

neuralop/models/tests/test_fno_block.py
```diff
@@ -1,5 +1,5 @@
 import torch
-from ..fno_block import FNOBlocks
+from neuralop.models.fno_block import FNOBlocks
```
JeanKossaifi, line 2: Let's revert this to relative import

JeanKossaifi: In general it's best to focus PR changes to just the feature we're adding

neuralop/models/uno.py
```diff
@@ -144,7 +143,8 @@ def __init__(self,
         
 
         if domain_padding is not None and domain_padding > 0:
-            self.domain_padding = DomainPadding(domain_padding=domain_padding, padding_mode=domain_padding_mode, output_scale_factor = output_scale_factor)
+            self.domain_padding = DomainPadding(domain_padding=domain_padding, padding_mode=domain_padding_mode\
+            , output_scale_factor = [i['res_scaling'] for i in layer_configs])
```
JeanKossaifi, line 147: Why the list comprehension here? We should accept (and use) a single float for all the spatial dimensions. Otherwise we should probably just have a list directly passed in config.

JeanKossaifi (changes requested):
Thanks @ashiq24, this fixes the issues -- the logic of the code is good, let's just clean it up a little before merging!

neuralop/models/uno.py
ashiq24: as each layer of UNO has a different (possibly fractional) scaling factor, using a fixed output scale factor creates issues due to rounding for float value. So the actual output resolution may vary depending on the input resolution. 
So, for UNO, the Padding class is provided with the domain scaling factor along each dimension for all the layers.
Added support for output_scale_factor as a scaler and None type.

neuralop/models/padding.py
```diff
@@ -54,23 +58,37 @@ def pad(self, x):
             
             print(f'Padding inputs of {resolution=} with {padding=}, {self.padding_mode}')
 
+            output_pad = padding
+
+            for scale_factor in self.output_scale_factor:
+                if isinstance(scale_factor, (float, int)):
+                    scale_factor = [scale_factor]*len(resolution)
+                output_pad = [int(round(i*j)) for (i,j) in zip(scale_factor,output_pad)]
```
JeanKossaifi, line 66: Why the for loop? You're just overwriting output_pad at each iteration here. 
If scale_factor is a list shouldn't we be using one per mode?
We should also document the parameter

neuralop/models/padding.py
```diff
@@ -54,23 +58,37 @@ def pad(self, x):
...
+                unpad_indices = (Ellipsis, ) + tuple([slice(None, -p, None) for p in output_pad])
                 padding = [i for p in padding for i in (0, p)]
             else:
                 raise ValueError(f'Got {self.padding_mode=}')
             
             self._padding[f'{resolution}'] = padding
 
             padded = F.pad(x, padding, mode='constant')
+
+            out_put_shape = padded.shape[2:]
```
JeanKossaifi, line 84: Let's not call the variable out_put_shape, should be just `output_shape`.

neuralop/models/uno.py
JeanKossaifi: But each layer should have a list, this looks like each layer has an int? 
i.e. layer_config['res_scaling'] should be a list directly?

neuralop/models/tests/test_uno.py
```diff
@@ -10,16 +10,21 @@ def test_UNO():
...
                             ]
     horizontal_skips_map ={4:0,3:1}
-    model = UNO(3,3,5,layer_configs = layer_configs, horizontal_skips_map = horizontal_skips_map, n_layers = 5, domain_padding = 0.2, output_scale_factor = 2)
+    model = UNO(3,3,5,layer_configs = layer_configs, horizontal_skips_map = horizontal_skips_map, n_layers = 5, domain_padding = 0.2)
 
     t1 = time.time()
-    in_data = torch.randn(10,3,20,20)
+    in_data = torch.randn(32,3,64,64)
+    out = model(in_data)
+    out = model(in_data)
```
JeanKossaifi, line 18: The model is queried 3 times on the same input

JeanKossaifi (commented):
It looks much better overall. THere are still a few issues, let's address them in a follow up PR so we can already merge this.
Mainly: let's try to make the code as clear and readable as possible (clear logic, clear variable names and more docstrings), add a test for updates to the domain padding, and clarify how we want to pass the layers config.

PR #116, opened 2023-04-15
Uno

ashiq24:


neuralop/models/uno.py
```diff
@@ -139,12 +146,25 @@ def __init__(self,
...
-        assert len(self.layer_configs) == n_layers
+        
+
+        self.layer_configs = []
+        for l in range(n_layers):
+            l_config = {}
+            l_config['out_channels'] = uno_out_channels[l]
+            l_config['n_modes'] = uno_n_modes[l]
+            l_config['res_scaling'] = uno_scalings[l]
+            self.layer_configs.append(l_config)
```
JeanKossaifi, line 157: We don't need to create this `layer_configs` dict, let's directly store the `uno_foo`

neuralop/models/uno.py
```diff
@@ -139,12 +146,25 @@ def __init__(self,
...
+        
+        if self.horizontal_skips_map is None:
+            self.horizontal_skips_map = {}
+            for i in range(n_layers//2,0,):
+                self.horizontal_skips_map[n_layers - i -1] = i
```
JeanKossaifi, line 162: I think we could just remove the horizontal skips map and just assume a regular UNO architecture. That's 99% of the usecases.

neuralop/models/uno.py
```diff
@@ -139,12 +146,25 @@ def __init__(self,
...
         
 
         if domain_padding is not None and domain_padding > 0:
             self.domain_padding = DomainPadding(domain_padding=domain_padding, padding_mode=domain_padding_mode\
-            , output_scale_factor = [i['res_scaling'] for i in layer_configs])
+            , output_scale_factor = [i['res_scaling'] for i in self.layer_configs])
```
JeanKossaifi, line 167: Let's directly use the res_scaling from the list

neuralop/models/tests/test_uno.py
```diff
@@ -3,14 +3,36 @@
...
-    layer_configs = [{"out_channels":20, "n_modes" : [5,5], "res_scaling" :[0.5,0.5] },\
-                    {"out_channels":20, "n_modes" : [5,5], "res_scaling" :[1,1] },\
-                    {"out_channels":20, "n_modes" : [5,5], "res_scaling" :[1,1] },\
-                    {"out_channels":20, "n_modes" : [5,5], "res_scaling" :[2,2] },\
-                    {"out_channels":10, "n_modes" : [5,5], "res_scaling" :[2,2] },\
-                            ]
+    uno_layers = {"out_channels":[32,64,64,64,32],\
+                "n_modes" : [[5,5],[5,5],[5,5],[5,5],[5,5]],\
+                "res_scaling" : [[1.0,1.0],[0.5,0.5],[1,1],[1,1],[2,2]]\
+                        }
```
JeanKossaifi, line 9: We no longer need the dict, we can just use the lists you used below

JeanKossaifi (commented):
Thanks @ashiq24, looks good! Left some comments - not sure what is happening with that test, it seems to only be here -- the tests in the main branch and tensorly all pass.

neuralop/models/uno.py
ashiq24: It makes the forward call cleaner. Please note that **horizontal_skips_map** can be set to **None**, and then it fills up the **horizontal_skips_map** with a regular skip connection.

JeanKossaifi:
Subsumed by #125

PR #152, opened 2023-06-07
Finodev

ashiq24:


neuralop/models/uno.py
```diff
@@ -153,19 +153,25 @@ def __init__(self,
...
+            self.output_scaling_factor = np.multiply(self.output_scaling_factor, k)
+        self.output_scaling_factor = self.output_scaling_factor.tolist()
+        if len(self.output_scaling_factor) == 1:
+            self.output_scaling_factor = self.output_scaling_factor[0]
+
         if isinstance(self.output_scaling_factor, (float, int)):
             self.output_scaling_factor = [self.output_scaling_factor]*self.n_dim
         
-
+        print("calculated out factor", self.output_scaling_factor)
```
JeanKossaifi, line 174: Let's add an attribute `verbose` (pass it in init) and only print `if self.verbose:`

neuralop/models/uno.py
```diff
@@ -195,7 +201,7 @@ def __init__(self,
                                             use_mlp=use_mlp, 
                                             mlp_dropout=mlp_dropout, 
                                             mlp_expansion=mlp_expansion,
-                                            output_scaling_factor = self.uno_scalings[i],
+                                            output_scaling_factor = [self.uno_scalings[i]],
```
JeanKossaifi, line 204: Why is it a list?

neuralop/models/uno.py
```diff
@@ -186,7 +192,7 @@ def __init__(self,
 
             if i in self.horizontal_skips_map.keys():
                 prev_out = prev_out + self.uno_out_channels[self.horizontal_skips_map[i]]
-            print([self.uno_scalings[i]])
+            print(self.uno_scalings[i])
```
JeanKossaifi, line 195: I don't think this print is necessary

neuralop/models/uno.py
```diff
@@ -225,6 +231,7 @@ def forward(self, x):
 
         if self.domain_padding is not None:
             x = self.domain_padding.pad(x)
+        print(x.shape)
```
JeanKossaifi, line 234: Let's remove this and the next print too

neuralop/models/uno.py
```diff
@@ -153,19 +153,25 @@ def __init__(self,
...
         self.operator_block = operator_block
         self.integral_operator = integral_operator
         
         if self.horizontal_skips_map is None:
             self.horizontal_skips_map = {}
             for i in range(n_layers//2,0,):
                 self.horizontal_skips_map[n_layers - i -1] = i
-                
+
+        self.output_scaling_factor = np.ones_like(self.uno_scalings[0])
```
JeanKossaifi, line 164: Let's explicitly make this a list, not a tensor

neuralop/models/uno.py
```diff
@@ -153,19 +153,25 @@ def __init__(self,
...
+        for k in self.uno_scalings:
```
JeanKossaifi, line 165: Can you add some comment here, it's hard to read and understand the logic.

neuralop/models/uno.py
```diff
@@ -153,19 +153,25 @@ def __init__(self,
...
+            self.output_scaling_factor = np.multiply(self.output_scaling_factor, k)
+        self.output_scaling_factor = self.output_scaling_factor.tolist()
```
JeanKossaifi, line 167: Let's make it a list and not add additional `tolist`.
You can replace the np.multiply (which should work on a list anyway) with math.prod.
Actually we don't need the list, you can directly take the prod I think

neuralop/models/spectral_convolution.py
```diff
@@ -480,8 +480,8 @@ def __init__(self, in_channels, out_channels, n_modes, incremental_n_modes=None,
         
         hm1, hm2 = self.half_n_modes[0], self.half_n_modes[1]
         if self.factorization is None:
-            self.W1 = nn.Parameter(torch.empty(self.in_channels, hm1, hm2, hm1, hm2, dtype = torch.cfloat))
-            self.W2 = nn.Parameter(torch.empty(self.in_channels, hm1, hm2, hm1, hm2, dtype = torch.cfloat))
+            self.W1 = nn.Parameter(torch.empty(hm1, hm2, hm1, hm2, dtype = torch.cfloat))
```
JeanKossaifi, line 483: Let's add this in a separate PR

neuralop/models/padding.py
```diff
@@ -54,36 +54,31 @@ def pad(self, x):
...
-
-#             for scale_factor in self.output_scale_factor:
-#                 if isinstance(scale_factor, (float, int)):
-#                     scale_factor = [scale_factor]*len(resolution)
             output_pad = [int(round(i*j)) for (i,j) in zip(self.output_scale_factor,output_pad)]
 
             if self.padding_mode == 'symmetric':
                 # Pad both sides
-                unpad_indices = (Ellipsis, ) + tuple([slice(p, -p, None) for p in output_pad ])
+                unpad_indices = (Ellipsis, ) + tuple([slice(p, -p, None) for p in output_pad[::-1] ])
```
JeanKossaifi, line 64: Why did you reverse here?

neuralop/models/resample.py
```diff
@@ -47,10 +47,10 @@ def resample(x, res_scale, axis, output_shape = None):
...
         new_size = output_shape
 
-    if len(axis) == 1:
-        return F.interpolate(x, size = new_size[0], mode = 'linear', align_corners = True)
-    if len(axis) == 2:
-        return F.interpolate(x, size = new_size, mode = 'bicubic', align_corners = True, antialias = True)
+#     if len(axis) == 1:
+#         return F.interpolate(x, size = new_size[0], mode = 'linear', align_corners = True)
+#     if len(axis) == 2:
+#         return F.interpolate(x, size = new_size, mode = 'bicubic', align_corners = True, antialias = True)
```
JeanKossaifi, line 42: Why did you comment these?

JeanKossaifi: Can we update to still use them, they are faster

JeanKossaifi (commented):
Thanks @ashiq24, left some comments

neuralop/models/uno.py
ashiq24: The Fno blocks takes scaling_factors which are 2d list
Scaling factor for each layer and scaling factor along each dims. 
It can also take a scaler and also 1D list with scaling factor along each layer.

neuralop/models/uno.py
ashiq24: removed it.

neuralop/models/uno.py
ashiq24: done

neuralop/models/uno.py
ashiq24: math.prod multiply numbers in a list. but I need to multiply 2 lists. Each list specifies the scaling factors along each of the dims.

neuralop/models/padding.py
ashiq24: The first element of the output_pad pads around the last dimension. So, the first element should in the last slice.

neuralop/models/resample.py
ashiq24: I can update this. But then the neural operator can only be used with PyTorch version later than 1.9.1. Because, " antialias" is not available in the previous distribution of PyTorch.

neuralop/models/padding.py
JeanKossaifi: Why is that the case? Shouldn't it be in the same order?

neuralop/models/uno.py
JeanKossaifi: Hmm this code is becoming very confusing - can we try to think and simplify this scaling API?

neuralop/models/padding.py
ashiq24: As per my understanding from the documentation, it is (left, right, top, botton, front, back) for 3D tensor... Left/right padding pads along the dim = -1. Top/bottom pads along the dim = -2... and so on.

neuralop/models/uno.py
ashiq24: To do the scaling we need two things ---
one -- scaling factor for each layer.
two -- scaling factor along each of the dimension.
We need a two array in general.
But in case, we scale each of the dimensions with same scaling factor, then it can be a 1d array, where each element is the scaling factor for each layer.

neuralop/models/padding.py
JeanKossaifi: Great point @ashiq24 - can we make a quick comment and point to the doc for this?

neuralop/models/uno.py
JeanKossaifi: Alright, let's process it to always be a nested list, at the highest level and we can use zip to do the element-wise multiplication. 

Do we **need** to have both output_scaling_factor and uno_scaling?

neuralop/models/padding.py
ashiq24: You meant in the code as a doc string?

JeanKossaifi: Yes, in the code just with `# comment` :)

neuralop/models/uno.py
ashiq24: done

.gitignore
```diff
@@ -79,3 +79,5 @@ target/
 
 #Ipython Notebook
 .ipynb_checkpoints
+
+setup_git.sh
```
JeanKossaifi, line 83: I don't think this needs to be in the main repo

JeanKossaifi:
Thanks @ashiq24 looks good overall but looks like the latest PR merged created a conflict, can we resolve?

ashiq24:
I think it now suitable for merging.

neuralop/models/padding.py
JeanKossaifi: @ashiq24  can you add a short comment here?

.gitignore
JeanKossaifi: Let's remove this one.

neuralop/models/resample.py
JeanKossaifi: Makes sense, let's open an issue to address this in a separate PR.

neuralop/models/uno.py
JeanKossaifi: @ashiq24 -- a list would be better here

neuralop/models/uno.py
JeanKossaifi: Let's add some comments before merging.

neuralop/models/uno.py
JeanKossaifi: ?

JeanKossaifi (commented):
Looks good overall, let's just address the comments to make the code a little clearer before merging.

JeanKossaifi:
Thanks @ashiq24, it looks good - just one last thing, can you have a look at [pep8](https://peps.python.org/pep-0008/)? 
In particular, can you remove the spaces around the ` = `, at least in the function calls/signatures?
Good to merge aside from that.

JeanKossaifi:
Thanks @ashiq24, merging.

PR #155, opened 2023-06-09
GINO project

kovachki:


ahmed_test.py
```diff
@@ -0,0 +1,98 @@
+import torch
```
JeanKossaifi, line 1: Should we move this to `scripts` ?

neuralop/datasets/data_utils.py
```diff
@@ -0,0 +1,65 @@
+import torch
```
JeanKossaifi, line 1: Should we make this one the default?

neuralop/datasets/dict_dataset.py
```diff
@@ -0,0 +1,21 @@
+from torch.utils.data import Dataset
+
+class DictDataset(Dataset):
```
JeanKossaifi, line 3: Would be good to add a docstring here.

neuralop/models/mlp.py
```diff
@@ -52,3 +52,29 @@ def forward(self, x):
                 x = self.dropout(x)
 
         return x
+
+#Reimplementation of the MLP class using Linear instead of Conv 
+class MLPLinear(torch.nn.Module):
```
JeanKossaifi, line 57: Nice  - Same we probably need to document it and add a unit-test

neuralop/models/positional_embedding.py
```diff
@@ -0,0 +1,19 @@
+import torch
+import torch.nn as nn
+
+class PositionalEmbedding(nn.Module):
```
JeanKossaifi, line 4: Should we merge that with the old one we have?

JeanKossaifi:
Subsumed by #181

Issue #103, opened 2023-03-20
"grad can be implicitly created only for scalar outputs"


itskobold:
Hi zongyi, thanks for your work.

I'm receiving this error "grad can be implicitly created only for scalar outputs" when I increase the number of in and out channels to 32.  No such error when in/out channels both = 1.

I'm preparing my data by running 128 FDTD simulations for 64 steps each, then splitting these simulations in the time domain.  Spatial domains are both 64 samples long.  The first 32 samples should be fed into FNO as the input (x) with the remaining 32 predicted (y).  These blocks of data are then converted to 128x32x64x64 tensors and passed to the network.  Do I have to prepare the data in some other way first?

Thanks in advance!

JeanKossaifi:
Hi @itskobold 
Can you share a minimal snippet of code to reproduce your error?

itskobold:
Hi again, below is a snippet from my code...

```
x_train, y_train = load_data(f"{proj_dir}/fdtd/{fdtd_dir}/{block_id}.pkl")
x_test, y_test = load_data(f"{proj_dir}/fdtd/{fdtd_dir}_test/{block_id}.pkl")
train_resolution = x_train.shape[-1]

if encode_output:
    reduce_dims = list(range(y_train.ndim))
    output_encoder = UnitGaussianNormalizer(y_train, reduce_dim=reduce_dims)
    y_train = output_encoder.encode(y_train)
else:
    output_encoder = None

train_db = TensorDataset(x_train, y_train,
                         transform_x=PositionalEmbedding(grid_boundaries, 0) if positional_encoding else None)
train_loader = torch.utils.data.DataLoader(train_db,
                                           batch_size=batch_size, shuffle=True,
                                           num_workers=0, pin_memory=True, persistent_workers=False)

test_db = TensorDataset(x_test, y_test,
                        transform_x=PositionalEmbedding(grid_boundaries, 0) if positional_encoding else None)
test_loader = torch.utils.data.DataLoader(test_db,
                                          batch_size=test_batch_size, shuffle=False,
                                          num_workers=0, pin_memory=True, persistent_workers=False)
test_loaders = {train_resolution: test_loader}

model = TFNO(n_modes=(32, 32),
             in_channels=32,
             out_channels=32,
             hidden_channels=32,
             projection_channels=64,
             factorization='tucker',
             rank=0.42)
model = model.to(device)

# Create the optimizer
optimizer = torch.optim.Adam(model.parameters(),
[...]

PR #160, opened 2023-06-13
h1loss output sum, fix for issue #103

itskobold:
h1loss output is summed to produce a scalar, training would previously break if number of output channels > 1

https://github.com/neuraloperator/neuraloperator/issues/103


kovachki:
For the loss to produce a scalar, the channel dimension must be added to the `reduce_dims` otherwise it will only reduce over the first dimension (this is the default behavior). Here is a 2d example:
```
loss = H1Loss(d=2, reduce_dims=[0,1])
x = torch.randn(5,10,128,128)
y = torch.randn(5,10,128,128)
loss(x,y)
```
This will be scalar.

dhpitt:
Closing, @kovachki's explanation of we compute loss this way is sufficient

PR #172, opened 2023-06-29
Add low-precision to TFNO

crwhite14:
This pull request adds low precision options for TFNO. Here is a brief summary.

 1. `opt.amp_autocast` was previously in the yaml file but was a no-op. This PR gets `opt.amp_autocast` working
- False (default): run the model in full precision
- True: turn on torch.amp.autocast, torch's built-in half-precision method. This turns many operations to half precision, with a few notable exceptions: reduction operations, weight updates (due to instability) and complex-valued operations (due to lack of implementation).

2. Add new parameter to the yaml file, `fno_block_precision`
- 'full' (default): standard full precision
- 'half': the FFT, contraction, and inverse FFT run in half precision
- 'mixed': the FFT runs in full-precision, and the contraction and inverse FFT run in half precision

3. Add new parameter to the yaml file, `stabilizer`
- None (default)
- 'tanh': adds a tanh just before the FFT in the fno-block. Typically this needs to be set to stabilize the FFT, when `fno_block_precision='half'`.

Running
```bash
python train_navier_stokes.py --opt.amp_autocast=True --tfno2d.fno_block_precision='half' --tfno2d.stabilizer='tanh'
python train_navier_darcy.py --opt.amp_autocast=True --tfno2d.fno_block_precision='half' --tfno2d.stabilizer='tanh'
```

Improves runtime and memory by up to 30%, depending on the GPU used, the resolution of the data (greater speedups for 64x64 resolution or higher), and other hyperparameters such as factorization and rank.

### DIFF
--- a/neuralop/models/spectral_convolution.py
+++ b/neuralop/models/spectral_convolution.py
@@ -339,8 +339,6 @@ class FactorizedSpectralConv(nn.Module):
 
         if self.fno_block_precision == 'half':
             x = x.half()
-        else:
-            x = x.float()
 
         x = torch.fft.rfftn(x, norm=self.fft_norm, dim=fft_dims)
 
--- a/neuralop/models/spectral_convolution.py
+++ b/neuralop/models/spectral_convolution.py
@@ -1,3 +1,4 @@
+from .einsum_utils import einsum_complexhalf
 from torch import nn
 import torch
 import itertools
@@ -33,7 +34,11 @@ def _contract_dense(x, weight, separable=False):
     if not torch.is_tensor(weight):
         weight = weight.to_tensor()
 
-    return tl.einsum(eq, x, weight)
+    if x.dtype == torch.complex32:
+        # if x is half precision, run a specialized einsum
+        return einsum_complexhalf(eq, x, weight)
+    else:
+        return tl.einsum(eq, x, weight)
 
 def _contract_dense_separable(x, weight, separable=True):
     if separable == False:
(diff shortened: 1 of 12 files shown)


neuralop/models/spectral_convolution.py
```diff
@@ -321,9 +336,22 @@ def forward(self, x, indices=0, output_shape = None):
         
         #Compute Fourier coeffcients
         fft_dims = list(range(-self.order, 0))
-        x = torch.fft.rfftn(x.float(), norm=self.fft_norm, dim=fft_dims)
 
-        out_fft = torch.zeros([batchsize, self.out_channels, *fft_size], device=x.device, dtype=torch.cfloat)
+        if self.fno_block_precision == 'half':
+            x = x.half()
+        else:
+            x = x.float()
```
JeanKossaifi, line 343: Should we remove this one or do you think we need to always explicitly cast here @crwhite14 @rtu715 ?

JeanKossaifi (commented):
This looks great, thanks!

neuralop/models/spectral_convolution.py
crwhite14: We can remove the "else: x = x.float()"
I just did that in this commit: https://github.com/neuraloperator/neuraloperator/pull/172/commits/bcc52e768c921ee68f3e6620df27367d63928b0d#diff-5ae1e49af12ed16c75135c0043a08575110fd03d4c722a837b60aa0950b31e32L342-L343

JeanKossaifi:
Thanks, great PR @crwhite14 @rtu715, merging!

PR #177, opened 2023-07-11
fix SFNO example

crwhite14:
Quick fix for plot_SFNO_swe:
- Add `fno_block_precision` as an argument to FactorizedSphericalConv.
- Make plot_SFNO_swe work on a GPU as well as CPU.


JeanKossaifi:
Thanks @crwhite14, merging!

PR #181, opened 2023-07-20
add gno and fnogno

kovachki:


neuralop/models/integral_transform.py
```diff
@@ -6,7 +6,47 @@
...
+        should be dim x + dim y or dim x + dim y + dim f
+    mlp_layers : list, default None
+        List of layers sizes speficing a MLP which 
+        parametrizes the kernel k. The MLP will be 
+        instansiated by the MLPLinear class  
+    mlp_non_linearity : callable, default torch.nn.functional.gelu
+        Non-linear function used to be used by the 
+        MLPLinear class. Only used if mlp_layers is 
+        given and mlp is None
+    transform_type : int, default 0
```
JeanKossaifi, line 42: Do you think a string might be less ambiguous? But even then naming is not obvious

neuralop/models/integral_transform.py
```diff
@@ -6,7 +6,47 @@
 from .mlp import MLPLinear
 
 class IntegralTransform(nn.Module):
-    def __init__(self, mlp=None,
+    """Integral Kernel Transform (GNO)
+    Computes one of the following:
+        (a) \int_{A(x)} k(x, y) dy
```
JeanKossaifi, line 11: here dy corresponds to the actual measure (ie this is always the weighted integral)?

neuralop/models/integral_transform.py
```diff
@@ -26,21 +66,56 @@ def __init__(self, mlp=None,
...
+        neighbors : dict
+            The sets A(x) given in CRS format. The 
+            dict must contain the keys "neighbors_index"
+            and "neighbors_row_splits." For descriptions
+            of the two, see NeighborSearch.
+        x : torch.Tensor of size [m, d2], default None
+            m points of dimension d2 over which the 
+            output function is defined. If None,
+            x = y.
+        f_y : torch.Tensor of size [n, d3], default None
```
JeanKossaifi, line 100: Can we use shape instead of size to be unambiguous? Currently not sure if it (n, ) or 3rd order, or (n, d3)

neuralop/models/tfnogno.py
```diff
@@ -91,23 +112,18 @@ def __init__(
         )
         del self.fno.projection
```
JeanKossaifi, line 113: This feels weird to me

JeanKossaifi: What do you think of just using a Spectral Conv rather than a full FNO so we don't have to meddle with internals?

JeanKossaifi: Other option (which I think I like more) is to first provide a GNO layer out of the box and then just have a trivial concat of GNO and FNO?

JeanKossaifi (commented):
This looks great to me overall. Some very minor comments on the doc.

Should we add a unit-test for the models to just check forward/backward? https://github.com/neuraloperator/neuraloperator/blob/main/neuralop/layers/tests/test_fno_block.py

And maybe move the kernel integral itself to layers rather than models?

PR #182, opened 2023-07-21
Liftproj mod to mlp

btolooshams:
Replacing listing and projection modules with mlp.

In this version, with n_layers=1, mlp out_channels with be hidden_channels (hence, hidden_channel must be set when n_layers=1. Fix is needed.


JeanKossaifi:
Thanks @btolooshams, looks good, merging!

PR #183, opened 2023-07-22
MLP additional statement to check for scenario where n_layers=1

btolooshams:


JeanKossaifi:
Great, thank you @btolooshams!

PR #185, opened 2023-07-23
adding flag option to only pad the last dim

btolooshams:
Added a pad_only_last_dim in DomainPadding class (default False). Updated FNO class to have the flag domain_padding_only_last_dim (default False). If set True, then padding is only applied to the last dim of the input.

### DIFF
--- a/neuralop/layers/padding.py
+++ b/neuralop/layers/padding.py
@@ -16,11 +16,12 @@ class DomainPadding(nn.Module):
     This class works for any input resolution, as long as it is in the form
     `(batch-size, channels, d1, ...., dN)`
     """
-    def __init__(self, domain_padding, padding_mode='one-sided', output_scaling_factor=None):
+    def __init__(self, domain_padding, padding_mode='one-sided', output_scaling_factor=None, pad_only_last_dim=False):
         super().__init__()
         self.domain_padding = domain_padding
         self.padding_mode = padding_mode.lower()
         self.output_scaling_factor = output_scaling_factor
+        self.pad_only_last_dim = pad_only_last_dim
 
         # dict(f'{resolution}'=padding) such that padded = F.pad(x, indices)
         self._padding = dict()
@@ -39,6 +40,8 @@ class DomainPadding(nn.Module):
         """
         resolution = x.shape[2:]
 
+
+
         if isinstance(self.domain_padding, (float, int)):
             self.domain_padding = [float(self.domain_padding)]*len(resolution)
         if self.output_scaling_factor is None:
@@ -52,7 +55,10 @@ class DomainPadding(nn.Module):
 
         except KeyError:
             padding = [int(round(p*r)) for (p, r) in zip(self.domain_padding, resolution)]
-            
+
+            if self.pad_only_last_dim:
+                padding = padding[-1:]
+
             print(f'Padding inputs of {resolution=} with {padding=}, {self.padding_mode}')
             
             
@@ -76,9 +82,8 @@ class DomainPadding(nn.Module):
                 padding = [i for p in padding for i in (0, p)]
             else:
                 raise ValueError(f'Got {self.padding_mode=}')
-            
+  
             self._padding[f'{resolution}'] = padding
-            
 
             padded = F.pad(x, padding, mode='constant')
 
--- a/neuralop/layers/padding.py
+++ b/neuralop/layers/padding.py
@@ -6,26 +6,22 @@ class DomainPadding(nn.Module):
 
     Parameters
     ----------
-    domain_padding : float
+    domain_padding : float or list
         typically, between zero and one, percentage of padding to use
+        if a list, make sure if matches the dim of (d1, ..., dN)
     padding_mode : {'symmetric', 'one-sided'}, optional
         whether to pad on both sides, by default 'one-sided'
-    padding_dim : bool or list
-        set True to pad all dimension, or provide a list of dimensions you want to pad, by default 'True'
-        the indices of dimensions to pad start with spatial dimension (excluding batch and channel).
-        first spatial dimension is 0, second spatial dimension is 1, ...
         
     Notes
     -----
     This class works for any input resolution, as long as it is in the form
     `(batch-size, channels, d1, ...., dN)`
     """
-    def __init__(self, domain_padding, padding_mode='one-sided', output_scaling_factor=None, padding_dim=True):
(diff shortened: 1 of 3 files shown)


JeanKossaifi:
Thanks @btolooshams. Let's make it more general, how about having two use cases:
- padding is True, in that case set padding = range(...) such that we pad all dimensions
- Padding is a list of dimensions (for your use case would be [4], in that case we just pad those dimensions.

The only question is whether we account batch-size and channels in the dimensions (dim 0 is then batch size). Otherwise 0 would be the first spatial dim. 
  

What do you think? Could you make the change for the padding and update the docstrings in the models?

btolooshams:
I have modified the previous pool request. The "pad_only_last_dim" is now changed to "padding_dim". If True, all dimensions (0,1,2) excluding batch and channel with be padded, if "padding_dim" is a list, then those dimensions in the list will only be padded.

An additional fix --> the padding in F is applied in reverse order, so I have added a line to reverse the padding. Padding was being wrongly applied before.

btolooshams:
The current implementation excludes batch and ch dim (0,1) in counting with dimension to pad. 0 is the first spatial dimension.

JeanKossaifi:
Thank you @btolooshams, this is great! Could we call it just `padding`? 
Would be great to add a quick test case in the unit-tests for sanity and to make sure it remains correct in the future?

neuralop/layers/tests/test_padding.py
```diff
@@ -3,16 +3,44 @@
...
+
     data = torch.randn((2, 3, 10, 10))
-    padder = DomainPadding(0.2, mode)
+    padder = DomainPadding(padding_dim, mode)
     padded = padder.pad(data)
 
     target_shape = list(padded.shape)
-    target_shape[-1] = target_shape[-2] = out_size[mode]
+    for ctr in range(1,3):
+        target_shape[-ctr] = out_size[mode][-ctr]
```
JeanKossaifi, line 19: Can we just write this out to make it easier to read with a small inline comment? ALso I don't know what `ctr` means, good to use clear names

neuralop/layers/tests/test_padding.py
```diff
@@ -3,16 +3,44 @@
...
+@pytest.mark.parametrize('padding_dim', [0.2, [0.1, 0.2]])
+def test_DomainPadding_2d(mode, padding_dim):
+    if isinstance(padding_dim, float):
+        out_size =  {'one-sided': [12, 12], 'symmetric': [14, 14]}
+    else:
+        out_size =  {'one-sided': [11, 12], 'symmetric': [12, 14]} 
+
     data = torch.randn((2, 3, 10, 10))
-    padder = DomainPadding(0.2, mode)
+    padder = DomainPadding(padding_dim, mode)
```
JeanKossaifi, line 14: Why do we call it padding_dim here?

neuralop/layers/padding.py
```diff
@@ -52,33 +57,50 @@ def pad(self, x):
...
             # padded by the amount mention at the first position of the 'padding' vector.
             # The details about F.pad can be found here : https://pytorch.org/docs/stable/generated/torch.nn.functional.pad.html
 
             if self.padding_mode == 'symmetric':
                 # Pad both sides
-                unpad_indices = (Ellipsis, ) + tuple([slice(p, -p, None) for p in output_pad[::-1] ])
+                unpad_list = list()
+                for p in output_pad[::-1]:
+                    if p == 0:
+                        unpad_amount_pos = None
```
JeanKossaifi, line 79: Why not leave 0 in this case?

neuralop/layers/padding.py
```diff
@@ -52,33 +57,50 @@ def pad(self, x):
...
+                        unpad_amount_neg = None
+                    else:
+                        unpad_amount_pos = p
+                        unpad_amount_neg = -p
+                    unpad_list.append(slice(unpad_amount_pos, unpad_amount_neg, None))
```
JeanKossaifi, line 84: I actually don't mind spelling it out but can we call it padding_start and padding_end or something clearer? :)

neuralop/layers/padding.py
```diff
@@ -39,8 +40,12 @@ def pad(self, x):
         """
         resolution = x.shape[2:]
 
+        # if domain_padding is list, then to pass on
         if isinstance(self.domain_padding, (float, int)):
             self.domain_padding = [float(self.domain_padding)]*len(resolution)
+
+        assert len(self.domain_padding) == len(resolution)
```
JeanKossaifi, line 47: Good check to have! Can you add a message if that's not the case to explain the issue to users?

neuralop/layers/tests/test_padding.py
```diff
@@ -3,16 +3,44 @@
...
+        out_size =  {'one-sided': [12, 12, 12], 'symmetric': [14, 14, 14]}
+    else:
+        out_size =  {'one-sided': [11, 10, 12], 'symmetric': [12, 10, 14]} 
+
+    data = torch.randn((2, 3, 10, 10, 10))
+    padder = DomainPadding(padding_dim, mode)
+    padded = padder.pad(data)
+
+    target_shape = list(padded.shape)
+    for ctr in range(1,4):
```
JeanKossaifi, line 39: Same comment as above - ideally tests should be self explanatory and easy to get by just reading the code!

JeanKossaifi (commented):
Thanks @btolooshams! I like the API like this, I left a few comments, mostly just style.

neuralop/layers/tests/test_padding.py
btolooshams: "ctr" is changed to "pad_dim", and comment added.

neuralop/layers/tests/test_padding.py
btolooshams: padding_dim is renamed to "padding"

neuralop/layers/padding.py
btolooshams: slice gives an error for p=0. So if padding is 0, p is set to None, hence no slicing is done to unpad list.

neuralop/layers/padding.py
btolooshams: name modified :)

neuralop/layers/padding.py
btolooshams: Explanation is added. "domain_padding length must match the number of spatial/time dimensions (excluding batch, ch)"

JeanKossaifi:
Thank you @btolooshams, this is great, merging!

PR #186, opened 2023-07-24
Add example IPython notebook for 1+1 dim Burgers problem

m4e7:
Migrate the example file [`fourier_1d.py`](https://github.com/neuraloperator/neuraloperator/blob/master/fourier_1d.py) from the old `master` branch to [`main:examples`](https://github.com/neuraloperator/neuraloperator/tree/main/examples) as `fourier_1d.ipynb` [1]. Reshape the old data for the newest version of the library, and add support to `layers/` for `float64` and `complex128` numbers.

Additionally, reformat touched files with `autopep8` and update some comments.

my commits: https://github.com/neuraloperator/neuraloperator/pull/186/files/e5812a4ef0940f4e0b3e78dfb65e29d74c2017df..95f965d8a20939159e923aa2f1f6e5bf41b6cbeb

[1] [Notebook](https://github.com/m4e7/neuraloperator/blob/mogab/add_eg_notebook_burgers/examples/fourier_1d.ipynb)


neuralop/layers/spectral_convolution.py
```diff
@@ -237,61 +301,89 @@ def __init__(self, in_channels, out_channels, n_modes, incremental_n_modes=None,
...
-                output_scaling_factor = [[s]*len(self.n_modes) for s in output_scaling_factor]
+                output_scaling_factor = [
+                    [s] * len(self.n_modes) for s in output_scaling_factor]
         self.output_scaling_factor = output_scaling_factor
 
         if init_std == 'auto':
             init_std = (1 / (in_channels * out_channels))
         else:
-            init_std = 0.02
+            init_std = init_std
```
m4e7, line 314: Assumption. If we want the default to be `0.02`, then that renders the input arg `init_std` unused (if it's a `float` / not "auto").

m4e7:
@JeanKossaifi @zongyi-li This is mostly ready for review, but I'm debugging clamping behavior after rebasing onto the latest commit in `neuraloperator/neuraloperator:main`.

neuralop/layers/spectral_convolution.py
JeanKossaifi: Good catch!

m4e7:
> @JeanKossaifi @zongyi-li This is mostly ready for review, but I'm debugging clamping behavior after rebasing onto the latest commit in `neuraloperator/neuraloperator:main`.

Fixed clamping! (now fixing tests...)

neuralop/layers/mlp.py
```diff
@@ -22,33 +17,81 @@ class MLP(nn.Module):
...
+                                else hidden_channels)
         self.non_linearity = non_linearity
-        self.dropout = nn.ModuleList([nn.Dropout(dropout) for _ in range(n_layers)]) if dropout > 0. else None
-        
+        self.dropout = nn.ModuleList(
+            [nn.Dropout(dropout) for _ in range(n_layers)]
+        ) if dropout > 0. else None
+
         Conv = getattr(nn, f'Conv{n_dim}d')
+        dtype = kwargs.get('dtype', None)
```
JeanKossaifi, line 51: We can explicitly add these to the init with default None

neuralop/layers/mlp.py
```diff
@@ -22,33 +17,81 @@ class MLP(nn.Module):
...
-        for i in range(n_layers):
-            if i == 0 and i == (n_layers - 1):
-                self.fcs.append(Conv(self.in_channels, self.out_channels, 1))
-            elif i == 0:
-                self.fcs.append(Conv(self.in_channels, self.hidden_channels, 1))
-            elif i == (n_layers - 1):
-                self.fcs.append(Conv(self.hidden_channels, self.out_channels, 1))
-            else:
-                self.fcs.append(Conv(self.hidden_channels, self.hidden_channels, 1))
+        if n_layers == 1:
```
JeanKossaifi, line 54: Why not keep the previous loop and just break if there's only one layer?

neuralop/layers/mlp.py
```diff
@@ -22,33 +17,81 @@ class MLP(nn.Module):
...
+            dtype=dtype,
+            device=device,
+        ))
 
     def forward(self, x):
         for i, fc in enumerate(self.fcs):
             x = fc(x)
-            if i < self.n_layers:
+            # Don't apply non-linearity on the last layer:
+            if i < (self.n_layers - 1):
```
JeanKossaifi, line 94: Nice catch!

neuralop/layers/padding.py
```diff
@@ -84,7 +93,8 @@ def pad(self, x):
 
             out_put_shape = padded.shape[2:]
 
-            out_put_shape = [int(round(i*j)) for (i,j) in zip(self.output_scaling_factor,out_put_shape)]
+            out_put_shape = [round(i * j) for (i, j)
```
JeanKossaifi, line 96: I know this isn't from this PR by we shouldn't be calling this out_put_shape, just `output_shape` - let's try to remove convoluted names and code from the library

neuralop/layers/skip_connections.py
```diff
@@ -14,25 +23,43 @@ def skip_connection(in_features, out_features, n_dim=2, bias=False, type="soft-g
...
+            n_dim=n_dim,
+        )
+    elif skip_type.lower() == 'linear':
+        Conv = getattr(nn, f'Conv{n_dim}d')
+        return Conv(
+            in_channels=in_features,
+            out_channels=out_features,
+            kernel_size=1,
+            bias=bias,
+            device=kwargs.get('device'),
```
JeanKossaifi, line 54: Same here let's make it explicit, the kwargs is more meant to silently absorb arguments not-common to all the skips but not the best pattern either.

neuralop/layers/mlp.py
```diff
@@ -22,33 +17,81 @@ class MLP(nn.Module):
...
+                                if hidden_channels is None
+                                else hidden_channels)
         self.non_linearity = non_linearity
-        self.dropout = nn.ModuleList([nn.Dropout(dropout) for _ in range(n_layers)]) if dropout > 0. else None
-        
+        self.dropout = nn.ModuleList(
+            [nn.Dropout(dropout) for _ in range(n_layers)]
+        ) if dropout > 0. else None
+
         Conv = getattr(nn, f'Conv{n_dim}d')
```
JeanKossaifi, line 50: As I read this I realize we'll need to handle the case for d>3 manually as pytorch doesn't have ConvNd

JeanKossaifi (commented):
Thanks @m4e7! I started a review, will try to go through rest soon. Let's try to split the main changes from the rest for the future so it's easier to review - now the diff is quite large so it's easy to miss important changes.

neuralop/layers/spectral_convolution.py
```diff
@@ -38,33 +45,37 @@ def _contract_dense(x, weight, separable=False):
         # if x is half precision, run a specialized einsum
         return einsum_complexhalf(eq, x, weight)
     else:
-        return tl.einsum(eq, x, weight)
+        return tl.backend.einsum(eq, x, weight)
```
m4e7, line 48: @JeanKossaifi I had to add `.backend.` in order to be able to call `einsum()` properly (otherwise `einsum()` was not found in `tensorly`). Is there a better way to get around that problem?

neuralop/layers/mlp.py
m4e7: Done in https://github.com/m4e7/neuraloperator/pull/6

neuralop/layers/spectral_convolution.py
m4e7: Never mind, I found the fix here: https://stackoverflow.com/questions/76368751/how-to-resolve-attributeerror-module-tensorly-has-no-attribute-svd-error

m4e7:
Closed in favor of https://github.com/m4e7/neuraloperator/pull/6

PR #193, opened 2023-08-09
Spectrum analysis of datasets.

Robertboy18:
I have pushed a script for analyzing the spectrum of a given dataset. This script performs a comprehensive analysis of the frequency components present in the data, providing valuable insights into understanding the difficulty of the dataset by analyzing its frequency spectrum.

For example, the script below calculates the spectrum of the Darcy flow dataset. Comments have been added for a more general setup as well.

Thank you!

### DIFF
--- a/examples/plot_darcy_flow_spectrum.py
+++ b/examples/plot_darcy_flow_spectrum.py
@@ -14,6 +14,7 @@ import torch
 import matplotlib
 import matplotlib.pyplot as plt
 from neuralop.utils import spectrum2
+from neuralop.datasets import load_darcy_flow_small
 
 font = {'size'   : 28}
 matplotlib.rc('font', **font)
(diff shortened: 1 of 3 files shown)


examples/plot_darcy_flow_spectrum.py
```diff
@@ -0,0 +1,104 @@
...
+S = s
+
+# additional paramaters for the dataset
+Re = 5000
+index = 1
+T = 100
+dataset_name = "Darcy Flow"
+
+# %%
+HOME_PATH = '/home/user/'
```
JeanKossaifi, line 40: @Robertboy18 this will not work for users -  better to just use the builtin dataset for this, as in https://github.com/neuraloperator/neuraloperator/blob/main/examples/plot_FNO_darcy.py

neuralop/utils.py
```diff
@@ -91,3 +91,47 @@ def get_wandb_api_key(api_key_file='../config/wandb_api_key.txt'):
...
+        The spatial size of the signal.
+        
+    Returns
+    --------
+    spectrum: a numpy array 
+        A 1D numpy array of shape (s,) representing the computed spectrum.
+    """
+    T = u.shape[0]
+    u = u.reshape(T, s, s)
+    # u = torch.rfft(u, 2, normalized=False, onesided=False) - depending on your choice of normalization and such
```
JeanKossaifi, line 113: Can we make these options in the function rather than a comment here?

neuralop/utils.py
```diff
@@ -91,3 +91,47 @@ def get_wandb_api_key(api_key_file='../config/wandb_api_key.txt'):
...
+    u = torch.fft.fft2(u)
+
+    # 2d wavenumbers following Pytorch fft convention
+    k_max = s // 2
+    wavenumers = torch.cat((torch.arange(start=0, end=k_max, step=1), \
```
JeanKossaifi, line 118: Can we do everything in pytorch instead of going back and forth pytorch <-> numpy?

JeanKossaifi (commented):
Thank you @Robertboy18, great to have! Added a few comments.

neuralop/utils.py
Robertboy18: Sure!

examples/plot_darcy_flow_spectrum.py
Robertboy18: That makes sense, will load the dataset directly then!

Robertboy18: Actually, the issue is that we need to load the dataset in its raw format (so .pt, .mat, .zarr etc) to get the full spectrum; we cannot use the data loaders! I can leave a comment asking the user to change the HOME_PATH as 'neuraloperator/neuralop/datasets/data/darcy_test_16.pt' this path is the direct link to the raw format dataset.

neuralop/utils.py
Robertboy18: Sure, that is a good suggestion.

examples/plot_darcy_flow_spectrum.py
JeanKossaifi: Why?

Robertboy18: Changed it!

JeanKossaifi:
Thank you @Robertboy18, looks great, merging!

Issue #162, opened 2023-06-15
lifting_channels passed to UNO is not used


acse-cz220:
Thank you for maintaining the library. I noticed that the hyperparameter **lifting_channels** is passed to the UNO model but never used. Only **hidden_channels** is passed to the **Lifting** class. Can you please clarify that?

JeanKossaifi:
That's a good point. This is something we need to do, make lifting and projection more generic and make them both MLPs by default, currently the lifting is just a single linear layer.

JeanKossaifi:
Fixed.

PR #196, opened 2023-08-14
Use lifting channels in FNO.lifting if it is passed

dhpitt:
Simple fix. Corresponding to Issue #162. Updated the FNO class so that when `lifting_channels` is passed to the FNO, `self.lifting` is an MLP with a hidden layer of size `lifting_channels` instead of a linear layer.


JeanKossaifi:
Looks good to me, I like that it doesn't change the interface! Thanks @dhpitt, merging!

PR #197, opened 2023-08-16
Fix mlp nonlinearity

ziqi-ma:
Fix nonlinearity of final MLP - otherwise output is lower-bounded by GELU.

### DIFF
--- a/neuralop/layers/mlp.py
+++ b/neuralop/layers/mlp.py
@@ -48,7 +48,7 @@ class MLP(nn.Module):
     def forward(self, x):
         for i, fc in enumerate(self.fcs):
             x = fc(x)
-            if i < self.n_layers:
+            if i < self.n_layers - 1:
                 x = self.non_linearity(x)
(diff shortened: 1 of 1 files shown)


JeanKossaifi:
Thanks @ziqi-ma! @m4e7 this fixes the issue you found too right?

m4e7:
> Thanks @ziqi-ma! @m4e7 this fixes the issue you found too right?

Yeah, this fixes it! We should also fix `MLPLinear.forward()`, too: https://github.com/neuraloperator/neuraloperator/blob/a371d0890698eb2617dc805b527449b7acac8581/neuralop/layers/mlp.py#L77C34-L77C34

JeanKossaifi:
@ziqi-ma would you be able to make that change too?

JeanKossaifi:
I'll merge this @ziqi-ma as this is an important fix and opened #200 to track the other issue if you want to open a PR for it too! Thanks for your contribution!

PR #198, opened 2023-08-16
Fix MLP to not apply non-linearity in last layer

m4e7:
Lifting/projection operators were clamping model outputs because all MLPs were
erroneously applying their non-linearity at every layer. Fix this off-by-one
error, as well as various lint/pep8 bits.


m4e7:
This problem has already been addressed (https://github.com/neuraloperator/neuraloperator/pull/197), so I'm closing this PR.

PR #199, opened 2023-08-18
Reformat `layers/` directory with `black`

m4e7:
Reformat select files (in particular, those changes in https://github.com/neuraloperator/neuraloperator/pull/186) in directory `neuralop/layers` with `black`.

Additionally, add a "Contributing" section to the README, and state `black` as a formatting tool.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-wandb 
+wandb
 ruamel.yaml
 configmypy
 tensorly
@@ -8,3 +8,4 @@ matplotlib
 opt-einsum
 h5py
 zarr
+torch
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,4 +8,3 @@ matplotlib
 opt-einsum
 h5py
 zarr
-torch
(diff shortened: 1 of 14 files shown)


m4e7:
@JeanKossaifi I've split formatting changes out from earlier commits here to reduce bloat in those PRs. I'll rebase #186 on top of this to clean it up.

requirements.txt
```diff
@@ -7,4 +7,5 @@ torch-harmonics
 matplotlib
 opt-einsum
 h5py
-zarr
\ No newline at end of file
+zarr
+torch
```
JeanKossaifi, line 11: @m4e7 the reason I hadn't included torch in the requirements is that typically users don't want to install it with pip (conda, Docker etc typically are better to handle CUDA installs) and this can cause issues particularly with `pip -U` where it may override existing install with wrong version.

JeanKossaifi (commented):
Thanks @m4e7 looks good to me, just one comment about reqs.

requirements.txt
m4e7: Aha, got it. I didn't realize that was a common user workflow. I've removed that line from `requirements.txt`

PR #201, opened 2023-08-21
CFD dataset and load_cfd

chrischoy:
Data loader and utility functions for  the CFD data from http://www.nobuyuki-umetani.com/publication/mlcfd_data.zip


JeanKossaifi:
The tests are broken - we had also tried to get open3d to work in the CI @kovachki but weren't able to.

dhpitt:
I downloaded the zip file from the provided URL. The directory structure does not match the CFD dataset code. From inside `mlcfd_data`, what is the `data_dir` supposed to be?

PR #206, opened 2023-08-29
Functionality for complex-valued spatial data

dhpitt:
This PR implements basic functionality to allow the Fourier Neural Operator to operate on complex-valued data. 

Changes:

- Apply the full FFT to complex-valued data
- Change logic for contraction and double weights, because the FFT of complex data isn't skew-symmetric
- Complex extension of GELU activation, complex-valued wrapper for simple Conv and MLP layers
- `CFNO` (1d,2d,3d) subclasses

Bahareh and I talked this idea through last week, and this is still very much a WIP.

### DIFF
--- a/neuralop/layers/complex_utils.py
+++ b/neuralop/layers/complex_utils.py
@@ -1,0 +1,49 @@
+"""
+Functionality for handling complex-valued spatial data
+"""
+
+from copy import deepcopy
+
+import torch
+from torch import nn
+import torch.nn.functional as F
+
+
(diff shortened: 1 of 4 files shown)


dhpitt:
Now linked to Issue #208

dhpitt:
> Can we also add some unit-tests for this?

Will do. This isn't quite ready for merging though, I still want to do some end-to-end tests with @btolooshams to make sure this can handle her data. 

It'll also need a slightly different testing structure, because the only way to verify its correctness is to run complex data through it

PR #209, opened 2023-08-30
 Fix Python 3.6 f-string compatibility and condense documentation for FNO classes

m4e7:
There are several instances of `f"{x=}"` in the repo, which causes issues when trying to import the library in to Python 3.6. Drop the `=` in these f-string to fix this problem.

Additionally, the repeated docstrings between base class `FNO` and subclasses `FNOnd` make maintaining documentation more work than it needs to be. The parameters are used identically in the base- and sub-class, and in some cases the subclass documentation is already out of date. Delete the repeated parameter comments in favor of referring users back up to the full documentation of `FNO` (tested in a local `make` of `sphinx-gallery`).

Lastly, run `black` on touched files.

Reviewer(s): @zongyi-li @JeanKossaifi

### DIFF
--- a/neuralop/training/patching.py
+++ b/neuralop/training/patching.py
@@ -30,7 +30,12 @@ class MultigridPatching2D(nn.Module):
         self.stitching = stitching
         
         if levels > 0:
-            print(f'MGPatching({self.n_patches=}, {self.padding_fraction=}, {self.levels=}, {use_distributed=}, {stitching=})')
+            print('MGPatching('
+                  f'n_patches={self.n_patches}, '
+                  f'padding_fraction={self.padding_fraction}, '
+                  f'levels={self.levels}, '
+                  f'use_distributed={use_distributed}, '
+                  f'stitching={stitching})')
         
         #If distributed patches are stiched, re-scale gradients to revert DDP averaging 
         if self.use_distributed and self.stitching:
--- a/neuralop/training/patching.py
+++ b/neuralop/training/patching.py
@@ -1,79 +1,93 @@
+import math
+
 import torch
-import math 
 from torch import nn
-from neuralop.mpu.mappings import gather_from_model_parallel_region, scatter_to_model_parallel_region
+
 import neuralop.mpu.comm as comm
+from neuralop.mpu.mappings import (
+    gather_from_model_parallel_region,
+    scatter_to_model_parallel_region,
+)
 
 
 class MultigridPatching2D(nn.Module):
-    def __init__(self, model, levels=0, padding_fraction=0, use_distributed=False, stitching=True):
-        """Wraps a model inside a multi-grid patching
-        """
+    def __init__(
+        self,
+        model,
+        levels=0,
+        padding_fraction=0,
+        use_distributed=False,
+        stitching=True,
+    ):
+        """Wraps a model inside a multi-grid patching"""
         super().__init__()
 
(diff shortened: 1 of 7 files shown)


JeanKossaifi:
I don't think we need to abandon new features like f-strings that make the code easier to read just to support Python 3.6: it was released in 2016 and has reached [end of life](https://www.python.org/downloads/release/python-360/).

JeanKossaifi:
Other changes are great!

neuralop/training/patching.py
```diff
@@ -1,74 +1,93 @@
...
+        """Wraps a model inside a multi-grid patching"""
         super().__init__()
 
         self.skip_padding = (padding_fraction is None) or (padding_fraction <= 0)
 
-        self.levels=levels
+        self.levels = levels
 
-        if isinstance(padding_fraction, float) or isinstance(padding_fraction, int):
+        if isinstance(padding_fraction, (float, int)):
```
JeanKossaifi, line 29: Thanks, that's much better

neuralop/training/patching.py
```diff
@@ -78,125 +97,169 @@ def _stitch(self, x):
...
 
-        #Size with padding removed    
+        # Size with padding removed
         size = x.size()
 
         # if self.mode == "batch-wise":
-        B = size[0]//(self.n_patches[0]*self.n_patches[1])
-        W = size[3]*self.n_patches[1]
+        B = size[0] // (self.n_patches[0] * self.n_patches[1])
+        W = size[3] * self.n_patches[1]
```
JeanKossaifi, line 111: I really don't like the way `black` formats operations - it actually doesn't follow pep8..

neuralop/training/patching.py
```diff
@@ -78,125 +97,169 @@ def _stitch(self, x):
...
+                width * self.padding_fraction[1],
+            ]
+        ]
         self.padding_height = padding[0]
         self.padding_width = padding[1]
 
         patched = make_patches(x, n=2**self.levels, p=padding)
 
-        s1_patched = patched.size(-2) - 2*padding[0]
-        s2_patched = patched.size(-1) - 2*padding[1]
```
JeanKossaifi, line 118: Same, is there a way to force black to follow pep8 for those, or ignore them?

neuralop/training/patching.py
```diff
@@ -78,125 +97,169 @@ def _stitch(self, x):
...
-            
+                x_sub = torch.nn.functional.pad(
+                    x_sub, pad=[s2_pad, s2_pad, 0, 0], mode="circular"
+                )
+
             if s1_pad > x_sub.size(-2):
                 diff = s1_pad - x_sub.size(-2)
-                x_sub = torch.nn.functional.pad(x_sub, pad=[0, 0, x_sub.size(-2), x_sub.size(-2)], mode='circular')
-                x_sub = torch.nn.functional.pad(x_sub, pad=[0, 0, diff, diff], mode='circular')
+                x_sub = torch.nn.functional.pad(
```
JeanKossaifi, line 180: That's another case where I find the formatted version actually marginally less readable

m4e7:
> I don't think we need to abandon new features like f-strings that make the code easier to read just to support Python 3.6: it was released in 2016 and has reached [end of life](https://www.python.org/downloads/release/python-360/).

Done. I've dropped that commit.

JeanKossaifi:
Great, thanks @m4e7, merging!

PR #212, opened 2023-09-06
Support for a generator input type in GaussianNormalizer

chrischoy:


neuralop/utils.py
```diff
@@ -64,78 +102,110 @@ def to(self, device):
...
     for j in range(1, n_observations + 1):
         ind = torch.where(index == j)
         spectrum[:, j - 1] = (signal[:, ind[0], ind[1]].sum(dim=1)).abs() ** 2
-    
+
     spectrum = spectrum.mean(dim=0)
     return spectrum
 
+
+if __name__ == "__main__":
```
JeanKossaifi, line 195: Can we put this in neuralop/test/test_utils.py and just make it a unit-test instead? It will also automatically be run by the CI this way.

neuralop/utils.py
```diff
@@ -1,42 +1,80 @@
...
+    ) -> Tuple[int, Tensor, Tensor]:
+        """
+        Compute mean and standard deviaion
+        """
+        if isinstance(x, Tensor):
+            mean = torch.mean(x, reduce_dim, keepdim=True).squeeze(0)
+            std = torch.std(x, reduce_dim, keepdim=True).squeeze(0)
+            count = len(x)
+        elif isinstance(x, Generator):
+            first_item = next(x)
```
JeanKossaifi, line 44: As a minor point I would just have the for loop:
```python
for index, element in enumerate(x):
    if index:
        total_sum += item
        total_sum_square += item**2
    else:
        total_sum = element
        total_sum_square = element**2
mean = total_sum / (index + 1)
variance = ( total_sum_square - (total_sum**2)/(index + 1) )/index
```

JeanKossaifi: I think we can also make it a standalone function e.g.
incremental_mean_variance(data_iterator)

neuralop/utils.py
```diff
@@ -1,42 +1,80 @@
...
+            )
+            print(f"   Mean and std of shape {self.mean.shape}, eps={eps}")
+
+    def compute_statistics(
+        self, x: Union[Tensor, Generator], reduce_dim: List[int] = [0]
+    ) -> Tuple[int, Tensor, Tensor]:
+        """
+        Compute mean and standard deviaion
+        """
+        if isinstance(x, Tensor):
```
JeanKossaifi, line 39: might be more robust to use torch.is_tensor(x)
Also we could just always use the second version?

neuralop/utils.py
JeanKossaifi: Lastly maybe we should support batch - this way we can directly call the normalizer on a (lazy) data loader?

JeanKossaifi: Oh also shouldn't we also use the reduce dim when computing mean / variance?

JeanKossaifi (commented):
Thanks @chrischoy, I left some comments!

neuralop/utils.py
chrischoy: is_tensor is a wrapper for isinstance(x, tensor).

neuralop/utils.py
chrischoy: I will add the support for the reduction. The test is already done on a batch.

neuralop/utils.py
chrischoy: Okay sounds good.

JeanKossaifi:
Support for Gaussian Normalizer was added in https://github.com/neuraloperator/neuraloperator/blob/main/neuralop/datasets/output_encoder.py

PR #213, opened 2023-09-07
Changed MLP to nn.Linear to allow 4D prediction

gegewen:
This PR changed the lifting and projection in FNO to allow 4D prediction

### DIFF
--- a/neuralop/layers/mlp.py
+++ b/neuralop/layers/mlp.py
@@ -32,6 +32,7 @@ class MLP(nn.Module):
         **kwargs,
     ):
         super().__init__()
+        self.n_dim = n_dim
         self.n_layers = n_layers
         self.in_channels = in_channels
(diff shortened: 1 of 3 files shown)


JeanKossaifi:
Thanks @gegewen - I think we should directly modify the MLP to support 4D instead, do you agree?

gegewen:
There is no Conv4D if we use MLP - are there any suggestions?

JeanKossaifi:
Thanks @gegewen -- could we add the nested-fno in a separate PR? I want to merge your changes to support 4D MLP asap!

gegewen:
Hey, sorry @JeanKossaifi, the nested no was automatically pushed here when I pushed it to my branch - do you want me to revert this one to without NestedFNO first?

JeanKossaifi:
That would be great, thanks @gegewen! Looks like there is a conflict with the latest version as well.

PR #214, opened 2023-09-08
Split `preactivation` from `FNOBlock.forward()`

m4e7:
Mainly, refactor `FNOBlock.forward()` for readability:

* Create a separate method `FNOBlock.forward_with_preactivation()` where it is assumed `preactivation=True` for clarity of code flow. Likewise, assume `preactivation=False` in the new method `FNOBlock.forward_with_postactivation()`. The old method `forward()` now dispatches to one of these.
* In multiple places, we were checking `self.output_scaling_factors` and calling `resample` like boilerplate. Refactor this check and call (with a no-op return if `scaling_factors is None and output_shape is None`).
* Additionally, refactor out `output_scaling_factors` validation, which was being duplicated across 4 files.

Reviewer(s): @JeanKossaifi @zongyi-li

### DIFF
--- a/neuralop/layers/fno_block.py
+++ b/neuralop/layers/fno_block.py
@@ -1,3 +1,5 @@
+from typing import List, Optional, Union
+
 import torch
 from torch import nn
 import torch.nn.functional as F
@@ -7,7 +9,10 @@ from .normalization_layers import AdaIN
 from .resample import resample
 from .skip_connections import skip_connection
 from .spectral_convolution import SpectralConv
-from ..utils import validate_output_scaling_factor
+from ..utils import validate_scaling_factor
+
+
+Number = Union[int, float]
 
 
 class FNOBlocks(nn.Module):
@@ -16,7 +21,7 @@ class FNOBlocks(nn.Module):
         in_channels,
         out_channels,
         n_modes,
-        output_scaling_factor=None,
+        output_scaling_factor: Optional[Union[Number, List[Number]]] = None,
         n_layers=1,
         incremental_n_modes=None,
         fno_block_precision="full",
@@ -47,9 +52,9 @@ class FNOBlocks(nn.Module):
         self.n_modes = n_modes
         self.n_dim = len(n_modes)
 
-        self.output_scaling_factor = validate_output_scaling_factor(
-            output_scaling_factor, self.n_dim, n_layers
-        )
+        self.output_scaling_factor: Union[
+            None, List[List[float]]
+        ] = validate_scaling_factor(output_scaling_factor, self.n_dim, n_layers)
 
         self._incremental_n_modes = incremental_n_modes
         self.fno_block_precision = fno_block_precision
@@ -210,7 +215,7 @@ class FNOBlocks(nn.Module):
 
         x = x_fno + x_skip_fno
 
-        if (self.mlp is not None) or (index < (self.n_layers - index)):
+        if (self.mlp is not None) or (index < (self.n_layers - 1)):
             x = self.non_linearity(x)
 
         if self.mlp is not None:
--- a/neuralop/layers/fno_block.py
+++ b/neuralop/layers/fno_block.py
@@ -7,7 +7,7 @@ from .normalization_layers import AdaIN
 from .resample import resample
 from .skip_connections import skip_connection
 from .spectral_convolution import SpectralConv
-from ..utils import validate_output_scaling_factor_2d
+from ..utils import validate_output_scaling_factor
 
 
 class FNOBlocks(nn.Module):
@@ -47,7 +47,7 @@ class FNOBlocks(nn.Module):
         self.n_modes = n_modes
         self.n_dim = len(n_modes)
 
-        self.output_scaling_factor = validate_output_scaling_factor_2d(
+        self.output_scaling_factor = validate_output_scaling_factor(
             output_scaling_factor, self.n_dim, n_layers
         )
 
--- a/neuralop/layers/fno_block.py
+++ b/neuralop/layers/fno_block.py
@@ -261,11 +261,15 @@ class FNOBlocks(nn.Module):
 
         return x
 
-    def resample(self, x, index, output_shape):
+    def resample(self, x, index, output_shape=None):
         """Resamples input if scaling factors are available for this block."""
-        if self.output_scaling_factor is None:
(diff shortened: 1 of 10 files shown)


neuralop/utils.py
```diff
@@ -64,78 +69,128 @@ def to(self, device):
...
+
     spectrum = spectrum.mean(dim=0)
     return spectrum
 
+
+Number = Union[float, int]
+
+def validate_output_scaling_factor_1d(
+    output_scaling_factor: Optional[Union[Number, List[Number]]],
+    order: int,
```
JeanKossaifi, line 170: Why do you call them 1d and 2d? They are both N-D, where order = n_dim = N. 
Given this I think we should merge them and just have n_layers=1 by default, what do you think?

neuralop/layers/fno_block.py
```diff
@@ -173,6 +170,9 @@ def __init__(
                 "[instance_norm, group_norm, layer_norm]"
             )
 
+        if self.preactivation:
+            self.forward = self.forward_with_preactivation
```
JeanKossaifi, line 174: Doesn't overriding the other function make it tricky if the user wanted to switch back? 
I guess they can just re-instantiate

neuralop/layers/spectral_convolution.py
```diff
@@ -285,21 +287,14 @@ def __init__(
...
-        self.output_scaling_factor = output_scaling_factor
+        self.output_scaling_factor = validate_output_scaling_factor_2d(
+            output_scaling_factor, self.order, n_layers
+        )
 
         if init_std == "auto":
             init_std = 1 / (in_channels * out_channels)
         else:
-            init_std = 0.02
+            init_std = init_std
```
JeanKossaifi, line 297: Thanks, I kept forgetting to update this!

JeanKossaifi (commented):
Thanks, it's clearer! Just found the 1d/2d helpers a little confusing but I like the changes otherwise.

neuralop/layers/fno_block.py
m4e7: Is it likely that users will want to alternate between pre-activation and "post-" activation?

Regardless, I suppose a better control structure would be to let `forward()` dispatch to separate implementations. I'll change that.

JeanKossaifi: I'm not sure on this one either : I'm torn between the readability of that VS the extra cost of that check every single call. What do you think?

neuralop/utils.py
m4e7: > Why do you call them 1d and 2d?
That's just based on their return value. `validate_*_2d` returns a list of shapes, while `validate_*_1d` returns just a shape.

> I think we should merge them and just have n_layers=1 by default
That sounds fine, as long as a singleton "list" of shapes doesn't confuse any of the models.

neuralop/layers/fno_block.py
m4e7: Given forward goes on to call a lot of linear algebra, I think the cost of one conditional at the start of each `FNOBlock` is negligible (unless GPU acceleration makes that cost significantly cheaper somehow).

JeanKossaifi: Agreed

neuralop/utils.py
JeanKossaifi: I see - how about we call them validate_output_scaling_factor_single_layer and validate_output_scaling_factor_multiple_layers? 

A mouthful and a pain to write but self-explanatory and never used by end users most likely?

JeanKossaifi: Ideally we should never have to do both, makes me think there is a deficiency in the code - we should either check at model level or layer level, probably not both

neuralop/layers/fno_block.py
```diff
@@ -190,74 +190,82 @@ def set_ada_in_embeddings(self, *embeddings):
...
             x_fno = self.norm[self.n_norms * index](x_fno)
 
         x = x_fno + x_skip_fno
 
-        if (
-            not self.preactivation
-            and (self.mlp is not None)
-            or (index < (self.n_layers - index))
-        ):
+        if (self.mlp is not None) or (index < (self.n_layers - index)):
```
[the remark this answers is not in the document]
JeanKossaifi: Yes should be -1, good catch!

neuralop/utils.py
```diff
@@ -64,78 +69,128 @@ def to(self, device):
...
+    if output_scaling_factor is None:
+        return None
+    if isinstance(output_scaling_factor, (float, int)):
+        return [[float(output_scaling_factor)] * n_dim] * n_layers
+    if isinstance(output_scaling_factor, list) and isinstance(
+        output_scaling_factor[0], (float, int)
+    ):
+        return [[float(s)] * n_dim for s in output_scaling_factor]
+
+    # TODO raise ValueError
```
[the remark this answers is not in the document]
JeanKossaifi: Yes, should be None by default I think, i.e. no rescaling

neuralop/utils.py
m4e7: I made an allowance for non-positive values of `n_layers` to return a single layer instead of a list of layers (for a whole module). This is only used in `DomainPadding`.

neuralop/layers/fno_block.py
```diff
@@ -190,74 +192,88 @@ def set_ada_in_embeddings(self, *embeddings):
...
-            if not self.preactivation:
-                if index < (self.n_layers - 1):
-                    x = self.non_linearity(x)
+            x = self.mlp[index](x) + x_skip_mlp
+
         return x
 
+    def resample(self, x, index, output_shape):
+        """Resamples input if scaling factors are available for this block."""
+        if self.output_scaling_factor is None:
```
JeanKossaifi, line 266: Shouldn't you check both self.output_scaling_factor and output_shape vs x.shape?

m4e7: Good point. I've updated `FNOBlock.resample()` to defer first to arg `output_shape` to be passed to `resample`. AFAICT, when `output_shape is not None` in `resample.py:resample()` the other args are ignored (I passed dummy args `res_scale=1, axis=None`, so it should be fine).

JeanKossaifi:
This is great, thanks for all the changes @m4e7! Merging.

JeanKossaifi (approved):

PR #218, opened 2023-09-20
Padding correction

ashiq24:


JeanKossaifi:
Thanks @ashiq24! Could you add a test case with different padding per dimensions to make sure this is fixed in the future?

JeanKossaifi:
Thank you @ashiq24, merged!

PR #219, opened 2023-09-22
fix example Trainer API calls

dhpitt:
My last PR passes all tests but broke `make doc` because I didn't update the API in the `examples` folder. This PR fixes that.


JeanKossaifi:
Excellent, thanks @dhpitt, merging.

PR #224, opened 2023-09-25
Fix loss signatures to build doc

dhpitt:
Doc now builds successfully.

**Changes**
- change default loss signature to take arguments `y_pred, y`
- change `Trainer` to only send required arguments to `model.forward()`
- change SFNO training example to cpu-only (temp fix)


JeanKossaifi:
Thanks @dhpitt! Merged.

PR #232, opened 2023-10-09
fix small bug in the WandB logger callback

dhpitt:


JeanKossaifi:
Thanks @dhpitt, merging

PR #233, opened 2023-10-09
Remove torch_scatter and torch_cluster from CI pipeline's dependencies

dhpitt:
We currently use the module `torch_scatter` on main for one function called `segment_csr`, which aggregates features across each neighborhood for the GINO model in `neuralop.layers.integral_transform.py`. This implements a simple python version of the same function to remove our dependence on `torch_scatter` and fix that annoying OSError in our CI.

**Second note**: the same error occurs with the CPU version of `torch_cluster`, so I wrote a neighborhood search for CPU runners only.


dhpitt:
Small empirical tests show that my neighbor search function runs in about 1/3 the time of `torch_cluster.radius`.

neuralop/layers/neighbor_search.py
```diff
@@ -61,31 +69,34 @@ def forward(self, data, queries, radius):
...
-                bins[-1] -= 1
-            
-            neighbors_row_splits = torch.cumsum(bins, dim=0)
-            neighbors_row_splits = torch.cat((torch.tensor([0], 
-                                              dtype=neighbors_row_splits.dtype,
-                                              device=neighbors_row_splits.device),
-                                              neighbors_row_splits), dim=0)
-            
+                neighbors_count, neighbors_index = self.search_fn(data, queries, radius,
+                                                max_num_neighbors=data.shape[0])
```
JeanKossaifi, line 76: Should we wrap that code blow in a small, tested function?

neuralop/layers/neighbor_search.py
```diff
@@ -61,31 +69,34 @@ def forward(self, data, queries, radius):
...
-            
-            if neighbors_count[-1] != queries.shape[0] - 1:
-                add_max_element = True
-                neighbors_count = torch.cat((neighbors_count,
-                                             torch.tensor([queries.shape[0] - 1], 
-                                                          dtype=neighbors_count.dtype,
-                                                          device=neighbors_count.device)), 
-                                                          dim=0)
+            if not self.use_torch_cluster:
+                return_dict = self.search_fn(data, queries, radius)
```
JeanKossaifi, line 73: @kovachki Should we just keep the manual implementation and just choose between open3d (fast) and fallback (slow)?

neuralop/layers/segment_csr.py
```diff
@@ -0,0 +1,57 @@
...
+        reduce by taking the average of all neighbors.
+        Otherwise take the sum. 
+    """
+    if reduce not in ['mean', 'sum']:
+        raise ValueError("reduce must be one of \'mean\', \'sum\'")
+    
+    if torch.backends.cuda.is_built():
+        """only import torch_scatter when cuda is available"""
+        import torch_scatter.segment_csr as scatter_segment_csr
+        return scatter_segment_csr(src, indptr, reduce)
```
JeanKossaifi, line 33: Hmm cuda being built doesn't necessarily mean torch_scatter will be available? Shouldn't we check for both?

JeanKossaifi (commented):
Thanks for the changes @dhpitt  -- should we have a small unit-test for the fallbacks?

neuralop/layers/neighbor_search.py
dhpitt: Added a unit test for fallback search

dhpitt: Torch_cluster is slower than my implementation anyway. What if we just axed that coed

neuralop/layers/neighbor_search.py
dhpitt: Implemented.

neuralop/layers/segment_csr.py
dhpitt: Logic is changed.

neuralop/layers/neighbor_search.py
kovachki: Yes, I think this is best. Let's just remove cluster altogether.

neuralop/layers/segment_csr.py
kovachki: @dhpitt could you please check that your segment_csr works with backwards. You're doing some inplace operations so I'm not sure.

JeanKossaifi:
Awesome, thanks @dhpitt !

PR #234, opened 2023-10-11
Model checkpointing

dhpitt:
This PR implements two model checkpointing callbacks. One simply saves a model every N epochs, and the other monitors a metric from `eval_losses` to determine when to save a model.

### DIFF
--- a/neuralop/training/callbacks.py
+++ b/neuralop/training/callbacks.py
@@ -349,23 +349,23 @@ class ModelCheckpointCallback(Callback):
     Implements basic model checkpointing by saving a model every N epochs
     """
 
-    def __init__(self, ckpt_dir: Union[Path, str] = Path('./checkpoints'), interval: int = 1):
+    def __init__(self, checkpoint_dir: Union[Path, str] = Path('./checkpoints'), interval: int = 1):
         """
         Parameters
         ----------
-        ckpt_dir : str | pathlib.Path
+        checkpoint_dir : str | pathlib.Path
             folder in which to save checkpoints
         interval : int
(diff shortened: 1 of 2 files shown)


neuralop/training/trainer.py
```diff
@@ -69,6 +70,8 @@ def __init__(self, *,
                  use_distributed=use_distributed, 
                  verbose=verbose)
 
+        if load_from_checkpoint:
+            self.model = torch.load()
```
JeanKossaifi, line 74: We should be updating the state dict of the model here not overwriting it

neuralop/training/callbacks.py
```diff
@@ -346,4 +343,84 @@ def on_before_loss(self, out):
...
+            ckpt_path = self.ckpt_dir / f"{self.state_dict['epoch']}"
+            torch.save(self.state_dict['model'], ckpt_path)
+        
+
+class MonitorMetricCheckpointCallback(ModelCheckpointCallback):
+    """
+    Implements model checkpointing with the addition of monitoring a metric
+    """
+
+    def __init__(self, monitor: str, ckpt_dir: str = './checkpoints'):
```
JeanKossaifi, line 388: can we weite explicitly checkpoint_path?

JeanKossaifi: And monitor could be e.g. loss_key

JeanKossaifi: (same for above method for regular checkpoint)

JeanKossaifi (commented):
Looks good overall @dhpitt !

PR #241, opened 2023-10-18
Fix typo in checkpoint init

rybchuk:
There's a typo in the new checkpoint Callbacks, `super().__init()` instead of `super().__init__()`. When the Callbacks are run, they produce an error: `AttributeError: 'super' object has no attribute '_ModelCheckpointCallback__init'`. After this fix, I successfully ran a job and generated checkpoint files.


JeanKossaifi:
Thanks @rybchuk! @dhpitt let's add some unit-tests for the callbacks to make sure they work as expected. We can use pytest to mock the save file.

PR #242, opened 2023-10-18
Bug fixes and unit testing for Callbacks

dhpitt:
Simple change. Model loads checkpoint in `Trainer.train()` instead of `Trainer.__init__()`

### DIFF
--- a/examples/checkpoint_FNO_darcy.py
+++ b/examples/checkpoint_FNO_darcy.py
@@ -13,7 +13,7 @@ import matplotlib.pyplot as plt
 import sys
 from neuralop.models import TFNO
 from neuralop import Trainer
-from neuralop.training import OutputEncoderCallback, PauseTrainingOnEpochCallback, ResumeTrainingFromCheckpointCallback
+from neuralop.training import OutputEncoderCallback, StateCheckpointCallback
 from neuralop.datasets import load_darcy_flow_small
 from neuralop.utils import count_params
 from neuralop import LpLoss, H1Loss
@@ -76,7 +76,9 @@ trainer = Trainer(model=model, n_epochs=20,
                   device=device,
                   callbacks=[
                     OutputEncoderCallback(output_encoder),
-                    PauseTrainingOnEpochCallback(pause_epoch=10)
+                    StateCheckpointCallback(save_step=10,
+                                            save_optimizer=True,
+                                            save_scheduler=True)
                         ],             
                   wandb_log=False,
                   log_test_interval=3,
@@ -101,7 +103,8 @@ trainer = Trainer(model=model, n_epochs=20,
                   device=device,
                   callbacks=[
                     OutputEncoderCallback(output_encoder),
-                    ResumeTrainingFromCheckpointCallback(checkpoint_dir='./checkpoints/ep_10')
+                    StateCheckpointCallback(save_dir = './new_checkpoints',
+                                            resume_from_dir='./checkpoints/ep_10')
                         ],             
                   wandb_log=False,
                   log_test_interval=3,
--- a/examples/checkpoint_FNO_darcy.py
+++ b/examples/checkpoint_FNO_darcy.py
@@ -103,7 +103,7 @@ trainer = Trainer(model=model, n_epochs=20,
                   device=device,
                   callbacks=[
                     OutputEncoderCallback(output_encoder),
-                    StateCheckpointCallback(save_dir = './new_checkpoints',
+                    StateCheckpointCallback(save_dir='./new_checkpoints',
                                             resume_from_dir='./checkpoints/ep_10')
                         ],             
                   wandb_log=False,
--- a/examples/checkpoint_FNO_darcy.py
+++ b/examples/checkpoint_FNO_darcy.py
@@ -98,27 +98,16 @@ trainer.train(train_loader=train_loader,
               regularizer=False, 
               training_loss=train_loss)
 
-# %%
-# Grab the model from the 15th epoch checkpoint
-checkpointed_trainer = Trainer(model=model, n_epochs=5,
-                  device=device,
-                  callbacks=[
-                    OutputEncoderCallback(output_encoder),
-                        ],             
-                  wandb_log=False,
-                  log_test_interval=3,
-                  use_distributed=False,
-                  verbose=True,
-                  checkpoint_to_load='./checkpoints/ep_15.pt')
 
-# train and evaluate
+# train and evaluate on checkpointed model
 trainer.train(train_loader=train_loader,
               test_loaders=test_loaders,
               optimizer=optimizer,
               scheduler=scheduler, 
               regularizer=False, 
               training_loss=train_loss,
-              eval_losses=eval_losses)
+              eval_losses=eval_losses,
+              checkpoint_to_load='./checkpoints/ep_15.pt')
 
 
 # %%
--- a/examples/checkpoint_FNO_darcy.py
+++ b/examples/checkpoint_FNO_darcy.py
@@ -97,7 +97,7 @@ trainer.train(train_loader=train_loader,
               training_loss=train_loss)
 
 
-# resume training from checkpoint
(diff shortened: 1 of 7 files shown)


neuralop/training/callbacks.py
```diff
@@ -375,82 +376,123 @@ def on_before_loss(self, out):
...
     def on_before_val_loss(self, **kwargs):
         return self.on_before_loss(**kwargs)
-
-class ModelCheckpointCallback(Callback):
-
-    def __init__(self, checkpoint_dir: Union[Path, str] = Path('./checkpoints'), interval: int = 1):
-        """
-        Implements basic model checkpointing by saving a model every N epochs.
         
+class StateCheckpointCallback(Callback):
```
JeanKossaifi, line 380: Maybe we should just rename it CheckpointCallback?

examples/checkpoint_FNO_darcy.py
```diff
@@ -78,9 +76,9 @@
                   device=device,
                   callbacks=[
                     OutputEncoderCallback(output_encoder),
-                    ModelCheckpointCallback(
-                        checkpoint_dir='./checkpoints',
-                        interval='5')
+                    StateCheckpointCallback(save_step=10,
```
JeanKossaifi, line 79: save_interval?

examples/checkpoint_FNO_darcy.py
```diff
@@ -98,29 +96,27 @@
...
-# Grab the model from the 15th epoch checkpoint
-checkpointed_trainer = Trainer(model=model, n_epochs=5,
+
+# resume training from checkpoint
+
+trainer = Trainer(model=model, n_epochs=20,
                   device=device,
                   callbacks=[
                     OutputEncoderCallback(output_encoder),
+                    StateCheckpointCallback(save_dir = './new_checkpoints',
```
JeanKossaifi, line 106: there should be no spaces around =

examples/checkpoint_FNO_darcy.py
```diff
@@ -98,29 +96,27 @@
...
-              test_loaders=test_loaders,
+              test_loaders={},
               optimizer=optimizer,
               scheduler=scheduler, 
               regularizer=False, 
-              training_loss=train_loss,
-              eval_losses=eval_losses)
-
-
+              training_loss=train_loss)
```
JeanKossaifi, line 119: Perhaps we should also show how to resume training in this example. e.g. train for 2/3 epochs, save, and then instantiate a new trainer and resume from saved state

neuralop/training/__init__.py
```diff
@@ -1,4 +1,6 @@
 from .losses import LpLoss, H1Loss
 from .trainer import Trainer
 from .torch_setup import setup
-from .callbacks import Callback, MGPatchingCallback, OutputEncoderCallback, SimpleWandBLoggerCallback, ModelCheckpointCallback
\ No newline at end of file
+from .callbacks import Callback, MGPatchingCallback, \
```
JeanKossaifi, line 4: Just use parenthesis instead of `\` :
```python
from .callbacks import (Callback, MGPatchingCallback, 
                                        ....)
````

File: neuralop/training/callbacks.py [hunk 6]
```diff
@@ -375,82 +376,123 @@ def on_before_loss(self, out):
...
+        """StateCheckpointCallback handles saving and resuming 
+        training state from checkpoint .pt save files. 
+
         Parameters
         ----------
-        checkpoint_dir : str | pathlib.Path
-            folder in which to save checkpoints
-        interval : int
-            interval at which to check metric
+        save_dir : Union[Path, str], optional
```
JeanKossaifi, line 396: No need to use format syntax in docstring, no need for Union, Path or str is fine

neuralop/training/callbacks.py
```diff
@@ -375,82 +376,123 @@ def on_before_loss(self, out):
...
-
-    def __init__(self, checkpoint_dir: Union[Path, str] = Path('./checkpoints'), interval: int = 1):
-        """
-        Implements basic model checkpointing by saving a model every N epochs.
         
+class StateCheckpointCallback(Callback):
+    
+    def __init__(self, 
+                 save_dir: Union[Path, str] = './checkpoints', 
+                 save_best : str = None,
```
JeanKossaifi, line 384: How about 
```suggestion
                 monitor_metric : str = None,
```

neuralop/training/callbacks.py
```diff
@@ -375,82 +376,123 @@ def on_before_loss(self, out):
...
+                 save_step : int = 1,
```
JeanKossaifi, line 385:
```suggestion
                 save_interval : int = 1,
```

neuralop/training/callbacks.py
```diff
@@ -375,82 +376,123 @@ def on_before_loss(self, out):
...
-class ModelCheckpointCallback(Callback):
-
-    def __init__(self, checkpoint_dir: Union[Path, str] = Path('./checkpoints'), interval: int = 1):
-        """
-        Implements basic model checkpointing by saving a model every N epochs.
         
+class StateCheckpointCallback(Callback):
+    
+    def __init__(self, 
+                 save_dir: Union[Path, str] = './checkpoints',
```
JeanKossaifi, line 383: I'm not sure it's a good idea to have this default, we could just make it a mandatory arg.

neuralop/training/callbacks.py
```diff
@@ -375,82 +376,123 @@ def on_before_loss(self, out):
...
+            whether to save optimizer state, by default False
+        save_scheduler : bool, optional
+            whether to save scheduler state, by default False
+        save_regularizer : bool, optional
+            whether to save regularizer state, by default False
+        resume_from_dir : Union[Path, str], optional
+            folder from which to resume training state. 
+            Expects saved states in the form: (all but model optional)
+                model.pt, optimizer.pt, scheduler.pt, regularizer.pt
+            All state files present will be loaded.
```
JeanKossaifi, line 412: I'm wondering whether we should 
i) try to load best_model.pt (if monitor_metric is not None this will exist)
ii) if that fails, load model.pt (which will be the model from the latest save_interval)

neuralop/training/callbacks.py
```diff
@@ -375,82 +376,123 @@ def on_before_loss(self, out):
...
+        if self.save_best:
+            assert self.state_dict['eval_losses'], "Error: cannot monitor a metric if no validation metrics exist."
+            assert self.save_best in self.state_dict['eval_losses'].keys(), "Error: cannot monitor a metric outside of eval_losses."
+            self.best_metric_value = float['inf']
+        else:
+            self.best_metric_value = None
         
+        # load state dict if resume_from_dir is given
+        if self.resume_from_dir:
+            save_fnames = os.listdir(self.resume_from_dir)
```
JeanKossaifi, line 452: Let's not use os, it is not portable. Let's used pathlib.Path and directly glob or better just joinpath with the files we need

neuralop/training/callbacks.py
```diff
@@ -375,82 +376,123 @@ def on_before_loss(self, out):
...
+        
+            for fname in save_fnames:
+                # Strip the file extensions from .pt save files
+                state = fname.split('.')[0]
```
JeanKossaifi, line 456: pathlib provides utils to do this more elegantly, e.g. path.stem

JeanKossaifi (commented):
Looks good, left some comments

neuralop/training/callbacks.py
dhpitt: Currently isn't the way I have things set up - at each save_interval I create a new folder that saves the state. Same for best_model. Do you think it would be better to continually overwrite the save in one folder?

JeanKossaifi: I think so - to avoid blowing up memory?

dhpitt: Done

JeanKossaifi:
Awesome, thank you @dhpitt !

PR #244, opened 2023-10-23
Refactors SpectralConv for simpler FNO

JeanKossaifi:


neuralop/layers/spectral_convolution.py
```diff
@@ -390,6 +383,21 @@ def transform(self, x, layer_index=0, output_shape=None):
...
+    @property
+    def n_modes(self):
+        return self._n_modes
+    
+    @n_modes.setter
+    def n_modes(self, n_modes):
+        if isinstance(n_modes, int): # Should happen for 1D FNO only
+            n_modes = [n_modes]
+        else:
+            n_modes = list(n_modes)
```
m4e7, line 396: Should we have more explicit type checking here? Something like: 
```suggestion
        if isinstance(n_modes, int): # Should happen for 1D FNO only
            n_modes = [n_modes]
        elif all([isinstance(m, int) for m in n_modes]):  # Assumes n_modes is a populated iterable
            n_modes = list(n_modes)
        else:
            raise ValueError(msg)
```
Otherwise a model choking on `n_modes` further in execution could be pretty confusing.

neuralop/layers/legacy_spectral_convolution.py
```diff
@@ -0,0 +1,722 @@
...
+    else:
+        raise ValueError(
+            f'Got implementation={implementation}, expected "reconstructed" or "factorized"'
+        )
+
+
+Number = Union[int, float]
+
+
+class SpectralConv(BaseSpectralConv):
```
m4e7, line 191: Why do we want to keep this legacy code?

neuralop/layers/spectral_convolution.py
```diff
@@ -268,24 +268,22 @@ def __init__(
...
         # n_modes is the total number of modes kept along each dimension
-        # half_n_modes is half of that except in the last mode, correponding to
-        # the number of modes to keep in *each* quadrant for each dim
-        if isinstance(n_modes, int):
+        if isinstance(n_modes, int): # Should happen for 1D FNO only
             n_modes = [n_modes]
+        else:
+            n_modes = list(n_modes)
+        # The last mode has a redundacy as we use real FFT
+        # As a design choice we do the operation here to avoid users dealing with the +1
```
m4e7: I don't think this is documented in the class docstring, though that's where users would readily see it to know about this design. Consider adding this in the docstring.

neuralop/layers/spectral_convolution.py
```diff
@@ -443,7 +410,7 @@ def forward(
         batchsize, channels, *mode_sizes = x.shape
 
         fft_size = list(mode_sizes)
-        fft_size[-1] = fft_size[-1] // 2 + 1  # Redundant last coefficient
+        fft_size[-1] = fft_size[-1] // 2 + 1# Redundant last coefficient
```
m4e7, line 421: nit
```suggestion
        fft_size[-1] = fft_size[-1] // 2 + 1  # Redundant last coefficient
```

neuralop/layers/spectral_convolution.py
```diff
@@ -570,10 +526,10 @@ def forward(self, x, indices=0):
         slices = (
             slice(None),  # Equivalent to: [:,
             slice(None),  # ............... :,
-            slice(self.half_n_modes[0]),  # :half_n_modes[0]]
+            slice(None, self.n_modes[0]),  # :half_n_modes[0]]
```
m4e7, line 534: nit: alignment
```suggestion
            slice(None),  # Equivalent to:  [:,
            slice(None),  # ................ :,
            slice(None, self.n_modes[0]),  # :n_modes[0]]
```

m4e7 (commented):
IIUC, in the new API users should just have to set `n_modes` (instead of `incremental_n_modes`) to dynamically change the number of Fourier modes , right? Then what's `max_n_modes`, and how did we fulfill that functionality before this refactoring?

neuralop/layers/spectral_convolution.py
JeanKossaifi: I assumed it was clear modes should be int, it's also in the docstring: I don't check if it's an int for type check - just to allow users to call FNO(mode, ...) for the 1D FNO rather than FNO( (mode,), ...)

neuralop/layers/legacy_spectral_convolution.py
JeanKossaifi: In case we want to revert - for specific use cases. That version is harder to read but actually faster in some cases. Some users might want it for that reason

neuralop/layers/spectral_convolution.py
JeanKossaifi: Yeah I want to polish the PR a little more

JeanKossaifi:
The goal was to make the role of the variable clear from their names: `max_n_modes` is the max possible number of mode, needs to be specified before we create the weights. `n_modes` is variable and can be changed during training.

JeanKossaifi:
> IIUC, in the new API users should just have to set `n_modes` (instead of `incremental_n_modes`) to dynamically change the number of Fourier modes , right? Then what's `max_n_modes`, and how did we fulfill that functionality before this refactoring?


The goal was to make the role of the variable clear from their names: max_n_modes is the max possible number of mode, needs to be specified before we create the weights. n_modes is variable and can be changed during training.

PR #248, opened 2023-10-25
RNO and dissipative loss (MNO)

mliuschi:


neuralop/layers/recurrent_layers.py
```diff
@@ -0,0 +1,70 @@
+import torch
+import numpy as np
+import torch.nn as nn
+import torch.nn.functional as F
+
+from ..layers.spectral_convolution import SpectralConv
+from ..layers.fno_block import FNOBlocks
+
+class RNO_cell(nn.Module):
```
JeanKossaifi, line 9: Can we add docstrings here and short description of what it does

neuralop/layers/recurrent_layers.py
```diff
@@ -0,0 +1,70 @@
...
+    def forward(self, x, h):
+        z = torch.sigmoid(self.f1(x) + self.f2(h) + self.b1)
+        r = torch.sigmoid(self.f3(x) + self.f4(h) + self.b2)
+        h_hat = F.selu(self.f5(x) + self.f6(r * h) + self.b3) # selu for regression problem
+
+        h_next = (1. - z) * h + z * h_hat
+
+        return h_next
+
+class RNO_layer(nn.Module):
```
JeanKossaifi, line 40: Can we add docstrings here, also a reference for the paper so users can cite and short description of what it does

neuralop/models/rno.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class RNO(nn.Module):
+    """
+    N-Dimensional Recurrent Neural Operator.
+    
+    The RNO has an identical architecture to the finite-dimensional GRU, with 
+    the exception that linear matrix-vector multiplications are replaced by  
+    Fourier layers (Li et al., 2021), and for regression problems, the output 
+    nonlinearity is replaced by a SELU activation.
```
JeanKossaifi, line 18: Can add a short explanation (using the terms h, r etc) and a reference for users to cit

neuralop/models/rno.py
```diff
@@ -0,0 +1,164 @@
...
+                final_hidden_states.append(x[:, -1])
+            else:
+                x = pred_x
+                final_hidden_states.append(x)
+        h = final_hidden_states[-1]
+
+        if self.domain_padding:
+            h = h.unsqueeze(1) # add dim for padding compatibility
+            h = self.domain_padding.unpad(h)
+            h = h[:,0] # remove extraneous dim
```
JeanKossaifi, line 149: Why not just squeeze?

neuralop/models/rno.py
```diff
@@ -0,0 +1,164 @@
...
+                else:
+                    x = pred_x
+                final_hidden_states.append(x[:, -1])
+            else:
+                x = pred_x
+                final_hidden_states.append(x)
+        h = final_hidden_states[-1]
+
+        if self.domain_padding:
+            h = h.unsqueeze(1) # add dim for padding compatibility
```
JeanKossaifi, line 147: Why do you need the unsqueezing? Can you not specify the dims on which to pad?

neuralop/models/rno.py
```diff
@@ -0,0 +1,164 @@
...
+
+        if self.domain_padding:
+            x = self.domain_padding.pad(x)
+
+        final_hidden_states = []
+        for i in range(self.n_layers):
+            pred_x = self.layers[i](x, init_hidden_states[i])
+            if i < self.n_layers - 1:
+                if self.residual:
+                    x = x + pred_x
```
JeanKossaifi, line 137: I think you need to apply self.layers[i].transform to x in case of output_scaling factor passed by user

neuralop/training/losses.py
```diff
@@ -286,6 +286,51 @@ def __call__(self, y_pred, y, h=None, **kwargs):
         return self.rel(y_pred, y, h=h)
 
 
+class DissipativeLoss(object):
```
JeanKossaifi, line 289: Can you add docstring?

neuralop/models/rno.py
mliuschi: The input shape is `(batch, timesteps, dim, dom_size1, dom_size2, ..., dom_sizen)`, whereas the output shape is `(batch, dim, dom_size1, dom_size2, ..., dom_sizen)`. So if I want to use the same `DomainPadding` object (which was instantiated to ignore `timesteps`), then I would need to `unsqueeze(1)` right?

Another option is two have two different `DomainPadding` objects? But that feels wasteful as well. What are your thoughts?

neuralop/models/rno.py
mliuschi: Are you referring to `squeeze` for the last line? In general, if I were to `squeeze` here, I would be wary of accidentally deleting the `out_channels` dimension if it is 1.

neuralop/models/rno.py
mliuschi: I don't think I have implemented a `transform()` function for the recurrent layers. Do I need to do this? I thought any `output_scale_factor` differences were handled by the Fourier blocks under the hood.

neuralop/losses/data_losses.py
```diff
@@ -286,6 +286,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+        self.out_dim = out_dim
+        self.dissloss = LpLoss(d=out_dim, reduce_dims=self.reduce_dims, reductions=self.reductions)
+
+        if domain_shape is not None:
+            self.domain_shape = torch.tensor(domain_shape) # for PDE problems, this is a tuple of the domain size (dim1, ..., dimk)
+            self.domain_ndims = torch.prod(self.domain_shape)
+        else:
+            self.domain_shape = None
+    
+    def sample_uniform_spherical_shell(self, shape: tuple): # shape is (batch_size, dim1, ..., dimk, out_dim)
```
JeanKossaifi, line 333: Can we add a docstring here to explain what the shape correspond to, what the function does and how it should be used?

neuralop/losses/data_losses.py
```diff
@@ -286,6 +286,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+        npoints = shape[0]
+        shape = shape[1:]
+
+        ndim = torch.prod(torch.tensor(shape))
+        inner_radius, outer_radius = self.radii
+        
+        samp_radii = torch.zeros(npoints, 1).uniform_(inner_radius, outer_radius)
```
JeanKossaifi, line 340: sample_radii? I like to spell out the variables.

neuralop/losses/data_losses.py
```diff
@@ -286,6 +286,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+    loss_weight : float
+        Weighting between the dissipative loss and data loss
+    diss_radii : float tuple
+        Inner and outer radii of shell to sample from
+    out_dim : int
+        Dimension of input/output space (i.e., channels)
+    domain_shape : int tuple
+        Shape of PDE domain (resolution)
+    """
+    def __init__(self, data_loss, diss_y_rule, loss_weight: float, diss_radii: tuple, out_dim: int, domain_shape: tuple):
```
JeanKossaifi, line 316: Why not explicitly pass the inner and outer radius for clarity instead of a tuple?
Should we call out_dim `out_channels`? And domain shape, is it just the spatial domain? If so we can say in the docstring, excludes channels.

neuralop/losses/data_losses.py
```diff
@@ -286,6 +286,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
         return self.rel(y_pred, y, h=h)
 
 
+class DissipativeLoss(object):
+    """Dissipative loss is used to avoid model blow-up and to encourage the model
+    to learn dissipative dynamics (particularly in chaotic systems). The dissipative
+    loss samples points uniformly and randomly in a spherical shell around the
+    attractor of a chaotic system (at user-specified radii) and encourages
```
JeanKossaifi, line 293: Csn you expand on the radii -let's also describe each radius individually?

neuralop/losses/data_losses.py
```diff
@@ -286,6 +286,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+    that a learned model's predictions on this shell scale inwards by some
+    user-specified dissipative rule.
```
JeanKossaifi, line 295: Can you have some higher, less technical level intuition here? Readers can refer to the paper for details.

neuralop/losses/data_losses.py
```diff
@@ -286,6 +286,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+    
+    Introduced in "Learning Dissipative Dynamics in Chaotic Systems," NeurIPS 2022.
```
JeanKossaifi, line 297: Can you make it into a reference using the ResT syntax?

neuralop/losses/data_losses.py
```diff
@@ -286,6 +286,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+        self.dissloss = LpLoss(d=out_dim, reduce_dims=self.reduce_dims, reductions=self.reductions)
+
+        if domain_shape is not None:
+            self.domain_shape = torch.tensor(domain_shape) # for PDE problems, this is a tuple of the domain size (dim1, ..., dimk)
+            self.domain_ndims = torch.prod(self.domain_shape)
+        else:
+            self.domain_shape = None
+    
+    def sample_uniform_spherical_shell(self, shape: tuple): # shape is (batch_size, dim1, ..., dimk, out_dim)
+        npoints = shape[0]
```
JeanKossaifi, line 334: nitpick: `n_points` but also, should it not be `n_samples` or `batch_size`?

neuralop/losses/data_losses.py
```diff
@@ -286,6 +286,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+        shape = shape[1:]
```
JeanKossaifi, line 335: This is confusing, you are changing the meaning of shape half way through, best to avoid this. Why are the channels at the end? 
You could just do 
batch_size, *spatial_shape, n_channels = shape

neuralop/losses/data_losses.py
```diff
@@ -286,6 +286,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+
+        ndim = torch.prod(torch.tensor(shape))
```
JeanKossaifi, line 337: This isn't ndim, that would be len(shape). 
Instead of casting the shape to a tensor just use `math.prod(shape)`. 
We can call it e.g. n_elements

neuralop/losses/data_losses.py
```diff
@@ -286,6 +286,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+        inner_radius, outer_radius = self.radii
+        
+        samp_radii = torch.zeros(npoints, 1).uniform_(inner_radius, outer_radius)
+        vec = torch.randn(npoints, ndim) # ref: https://mathworld.wolfram.com/SpherePointPicking.html
+        vec = torch.nn.functional.normalize(vec, p=2.0, dim=1)
```
JeanKossaifi, line 342: Can you add comment one why this is done and also explain the breadcasting below?

neuralop/losses/data_losses.py
```diff
@@ -286,6 +286,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+        samp_radii = torch.broadcast_to(samp_radii, vec.shape)
```
JeanKossaifi, line 343: Same, try not to change the semantic meaning of variables half way through as it makes the code quite hard to read.

neuralop/losses/data_losses.py
```diff
@@ -286,6 +286,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+
+        return (samp_radii * vec).reshape((npoints, *shape))
+
+    def __call__(self, x, y, x_diss, pred_diss, **kwargs):
```
JeanKossaifi, line 347: Add a short docstring as the arguments are non-standard. 
Usually our losses take as input y_pred, y.

neuralop/models/rno.py
JeanKossaifi: Yes but what happens if self.layers[i] downscales (or upscales) the input? Or does this never happen?

neuralop/models/rno.py
JeanKossaifi: Yes, just squeeze that particular dimension, ie. h = h.squeeze(1)

neuralop/models/rno.py
JeanKossaifi: I see, let's keep it this way then

JeanKossaifi (commented):
Thank you @mliuschi, it looks much better! I left some comments, mostly just style and clarity.

neuralop/losses/data_losses.py
mliuschi: As I have it set up, the inner and outer radii are saved as fields in the `DissipativeLoss` object, but I don't think this is the best idea. What do you think?

neuralop/losses/data_losses.py
mliuschi: I've added the reference. Let me know if it's the wrong syntax!

neuralop/losses/data_losses.py
mliuschi: Yes, perhaps `batch_size` is clearer!

neuralop/losses/data_losses.py
mliuschi: I added a few comments for clarity. Let me know if they are not clear!

neuralop/losses/data_losses.py
mliuschi: Done! I have also added a "Usage example" in the main docstring for the class as well

neuralop/models/rno.py
mliuschi: If I understand correctly, we need `transform` after each spectral convolution in case we upscale/downscale the input? In `recurrent_layers.py`, I operate directly on FNO blocks, so I think any changes in resolution are dealt with inside the blocks. What do you think?

JeanKossaifi:
@mliuschi following on this based on our discussion, let us know when it is ready to review again.

vduruiss:
Adding @rwhan

mliuschi:
@vduruiss I think since it's written we may as well merge it. Just from a quick look, how far do you think we are from merging?

PR #260, opened 2023-11-02
Update tensor_dataset.py

slanthaler:
Fix bug in self.transform_y:

self.transform_y was applied to x instead of y


JeanKossaifi:
Thank you for spotting and fixing @slanthaler !

PR #263, opened 2023-11-09
fix domain_padding to accept list (e.g., [0,0,1]) in addition to sc…

btolooshams:
fix domain_padding in uno to accept list (e.g., [0,0,1] for 2d case).


JeanKossaifi:
Thank you @btolooshams, merging.

PR #267, opened 2023-12-04
fixing the horizontal_skips_map construction, it was not going through

btolooshams:


btolooshams:
There two fixed here. One for horizontal_skips_map construction as default, another for uno_scaling.

JeanKossaifi:
Thanks @btolooshams !

PR #272, opened 2023-12-20
Small fix in  examples/plot_darcy_flow.py

helgehr:
While going through an example I noticed some small issue. The fix does not change the result but only what gets printed, before the resolution was hard-coded to 'res' and the `test_data` was actually set to the `train_dataset[0]`. Now actual batches from the test_loaders and their resolution get printed.


JeanKossaifi:
Thanks for spotting the issue and fixing it @helgehr, this is great! Merging.

PR #274, opened 2023-12-24
Incremental Resolution and Frequency - Final Updates + Documentation fixes

Robertboy18:
I'm excited to share with you the final update for the Incremental version of Neural Operators. This pull request includes the latest changes and improvements to the Incremental module, bringing it to its completed state.

Here's a brief overview of the updates included in this pull request:

Algorithm 1: Incremental
Added a new method called loss_gap(loss) to handle the incremental mode increase based on the loss gap. The method tracks the loss values and increases the incremental modes if the difference between consecutive losses falls below a predefined threshold.

Algorithm 2: Gradient-based Explained Ratio
Implemented the grad_explained() method to handle the incremental mode increase based on the explained variance of gradients. The method accumulates gradients and checks the explained variance of each mode. The incremental mode is increased if the explained variance is below a specified threshold.

Algorithm 3: Regularize Input Resolution
Added the incremental_resolution_regularize(x, y) method to regulate the input resolution based on the dataset being used.
The method applies sub-sampling to the input data to reduce the resolution, depending on the current index and dataset.

Other improvements
Added dataprocessors and the IncrementalCallback and a test example for the users to understand. Added necessary checks to ensure that the incremental_grad and incremental_loss_gap flags cannot be used together. Updated comments throughout the code to provide clarity and improve readability. 

Also added some fixes to the doc.

Looking forward to your feedback and suggestions.


dhpitt (commented):
Thanks @Robertboy18! The logic looks good. 

1. It would be useful to add some unit tests to make sure the behavior is what we'd expect. 

2 is something I can help with - If I remember correctly, modifying the model via the state dict might produce weird behavior. It may create a copy instead of modifying the actual model from the training loop in place. Unit tests will help to verify.

Robertboy18:
> Thanks @Robertboy18! The logic looks good.
> 
> 1. It would be useful to add some unit tests to make sure the behavior is what we'd expect.
> 
> 2 is something I can help with - If I remember correctly, modifying the model via the state dict might produce weird behavior. It may create a copy instead of modifying the actual model from the training loop in place. Unit tests will help to verify.

Thanks for the review David, I added two simple unit tests for both the incremental resolution and the incremental frequency modes. The state_dict does get updated instead as the unit test shows clearly. Let me know if there is anything else to change: )

dhpitt (approved):
Thanks @Robertboy18! Nothing more to add from me.

Robertboy18:
Added some document fixes as well.

neuralop/training/callbacks.py
```diff
@@ -513,7 +515,12 @@ def on_epoch_end(self, epoch, **kwargs):
     
     def on_batch_start(self, idx, **kwargs):
         self._update_state_dict(idx=idx)
-
+        self.mode = "Train"
+        self.data = kwargs['data_processor']
```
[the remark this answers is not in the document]
Robertboy18: Yup good point!

neuralop/training/trainer.py
```diff
@@ -79,7 +80,12 @@ def __init__(self, *,
         self.device = device
         self.amp_autocast = amp_autocast
         self.data_processor = data_processor
-
+        self.incremental_resolution = False
+        
+        # If the data_processor is an IncrementalDataProcessor, then we need to do curriculum learning - Increase the resolution of the samples incrementally
```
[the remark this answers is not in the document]
Robertboy18: This isn't there anymore!

neuralop/datasets/data_transforms.py
```diff
@@ -64,6 +64,138 @@ def forward(self, **data_dict):
...
+        y = data_dict['y'].to(self.device)
+
+        if self.in_normalizer is not None:
+            x = self.in_normalizer.transform(x)
+        if self.positional_encoding is not None:
+            x = self.positional_encoding(x, batched=batched)
+        if self.out_normalizer is not None and self.train:
+            y = self.out_normalizer.transform(y)
+        
+        if mode == "Train":
```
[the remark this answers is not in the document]
Robertboy18: That makes sense, but the preprocess is where we are actually changing the samples, so that requires info of what mode we are using (the mode is actually changed in on_train_batch_start and on_val_batch_start).

dhpitt (approved):
Looks good, thanks for the edits @Robertboy18! @JeanKossaifi what do you think of the new structure?

doc/source/user_guide/fno.rst
```diff
@@ -7,9 +7,7 @@ This page (which takes about 10 minutes to read), introduces the Fourier neural
 It the first work that can learn resolution-invariant solution operators on Navier-Stokes equation,
 achieving state-of-the-art accuracy among all existing deep learning methods and
 up to 1000x faster than traditional solvers.
-Also check out the paper https://arxiv.org/abs/2010.08895 and article
-https://www.technologyreview.com/2020/10/30/1011435/ai-fourier-neural-network-cracks-navier-stokes-and-partial-differential-equations/
-
+Also check out the paper [1]_ and article [2]_.
```
dhpitt, line 10: nit: maybe "for further reading..."

neuralop/datasets/data_transforms.py
```diff
@@ -109,7 +109,139 @@ def forward(self, **data_dict):
         output = self.postprocess(output)
         return output, data_dict
 
+class IncrementalDataProcessor(torch.nn.Module):
+    def __init__(self, 
+                 in_normalizer=None, out_normalizer=None, 
+                 positional_encoding=None, device = 'cpu',
+                 dataset_sublist=[2, 1], dataset_resolution=16, dataset_indices=[2,3], epoch_gap=10, verbose=False):
```
dhpitt, line 116: nit: `subsampling_rates` might be a clearer name for this parameter

neuralop/datasets/tests/test_data_processor.py
```diff
@@ -31,4 +31,37 @@ def test_DefaultDataProcessor():
     
     _, inv_xform_data = pipeline.postprocess(out, xform_data)
 
-    assert_close(inv_xform_data['y'].cpu(), data['y'])
\ No newline at end of file
+    assert_close(inv_xform_data['y'].cpu(), data['y'])
+    
+
+# enure that the data processor incrementally increases the resolution
```
dhpitt, line 37: nit: ensure

neuralop/training/tests/test_callbacks.py
```diff
@@ -194,4 +196,69 @@ def test_load_from_checkpoint():
 
     # clean up dummy checkpoint directory after testing
     shutil.rmtree('./full_states')
+    
+    
+# enure that the model incrementally increases in frequency modes
+def test_incremental():
+    # Loading the Darcy flow dataset
+    train_loader, test_loaders, output_encoder = load_darcy_flow_small(
```
dhpitt, line 204: nit: thinking about test runtime. Might there be a way to verify that the behavior is as expected without loading such a big dataset?

neuralop/utils.py
```diff
@@ -247,4 +247,25 @@ def validate_scaling_factor(
...
+    rank = torch.matrix_rank(tensor).cpu()
+    return rank
+
+def compute_stable_rank(tensor):
+    # Compute the stable rank of a tensor
+    tensor = tensor.detach()
+    fro_norm = torch.linalg.norm(tensor, ord='fro')**2
+    l2_norm = torch.linalg.norm(tensor, ord=2)**2
+    rank = fro_norm / l2_norm
+    rank = rank.cpu()
```
dhpitt, line 263: nit: why does this need to be on the cpu?

dhpitt (commented):
Looks ready to go overall. A few tiny nitpicks, that's all

neuralop/datasets/burgers.py
```diff
@@ -114,4 +114,31 @@ def load_burgers_1dtime(
     output_encoder = None
     test_loaders = {'test':test_loader}
 
-    return train_loader, test_loaders, output_encoder
\ No newline at end of file
+    return train_loader, test_loaders, output_encoder
+
+def load_burgers_mat(data_path, n_train, n_test, batch_train=32, batch_test=100,
```
dhpitt, line 119: Is this function necessary for this PR? The less clutter perhaps the better

JeanKossaifi:
Thank you for all the changes @Robertboy18 and @dhpitt, it looks much better now! Great addition to the library!

PR #277, opened 2024-01-17
Add batched and variable mesh options to FNOGNO and IntegralTransform

dhpitt:
This PR adds batched mode to the FNOGNO and IntegralTransform for cases when the mesh stays constant across a batch. Tests for the FNOGNO have been updated to add this case.]

I envision the following use cases:

1. **Batched**: The information that is inconsistent across a batch is represented by the tensor input `f` to the FNOGNO. All other info (`in_p` = latent geometry, `out_p` = output geometry) stays constant. The model outputs a tensor of shape `[batch, n_output_points, n_out_channels]`.
2. **Variable**: No batch dimension, the forward pass is as before.


neuralop/layers/integral_transform.py
```diff
@@ -90,46 +89,49 @@ def __init__(
...
             (b) or (d). If None, (a) is computed.
-        weights : torch.Tensor of shape [n,], default None
+        weights : torch.Tensor of shape [batch, n] or [n,], default None
             Weights for each point y proprtional to the
             volume around f(y) being integrated. For example,
             suppose d1=1 and let y_1 < y_2 < ... < y_{n+1}
             be some points. Then, for a Riemann sum,
             the weights are y_{j+1} - y_j. If None,
             1/|A(x)| is used.
+        batched: bool, default True
```
juliusberner, line 125: Why don't we infer this option from the input?

neuralop/layers/integral_transform.py
```diff
@@ -165,7 +182,11 @@ def forward(
         else:
             reduction = "mean"
 
-        out_features = segment_csr(
-            rep_features, neighbors["neighbors_row_splits"], reduce=reduction
-        )
+        splits = neighbors["neighbors_row_splits"]
+        if batched:
+            # splits = splits.view(1,-1) # expand along batch dim if batched
```
juliusberner, line 187: I guess we can delete this comment?

neuralop/layers/segment_csr.py
```diff
@@ -3,50 +3,79 @@
...
+        and use_scatter
+    ):
         """only import torch_scatter when cuda is available"""
         import torch_scatter.segment_csr as scatter_segment_csr
+
         return scatter_segment_csr(src, indptr, reduce=reduce)
 
     else:
-        n_nbrs = indptr[1:] - indptr[:-1] # end indices - start indices
+        # if batched, shape [b, n_reps, channels]
```
juliusberner, line 45: Should we add a comment that this is a naive implementation of segment_csr? Also, should we warn the user in case `use_scatter=True` but the above `if` fails?

neuralop/layers/integral_transform.py
dhpitt: That sounds good. Updated

neuralop/layers/integral_transform.py
dhpitt: Done

neuralop/layers/integral_transform.py
```diff
@@ -138,21 +134,37 @@ def forward(
             x = y
 
         rep_features = y[neighbors["neighbors_index"]]
+
+        # batching only matters if latent embedding values are provided
```
juliusberner, line 138: It is unclear what a latent embedding is in this general function. Perhaps `f_y`?

neuralop/layers/integral_transform.py
```diff
@@ -90,36 +89,33 @@ def __init__(
...
             output function is defined. If None,
             x = y.
-        f_y : torch.Tensor of shape [n, d3], default None
+        f_y : torch.Tensor of shape [batch, n, d3] or [n, d3], default None
             Function to integrate the kernel against defined
             on the points y. The kernel is assumed diagonal
             hence its output shape must be d3 for the transforms
             (b) or (d). If None, (a) is computed.
-        weights : torch.Tensor of shape [n,], default None
+        weights : torch.Tensor of shape [batch, n] or [n,], default None
```
juliusberner, line 118: What happens if weights have a batch-dim., but there is no `f_y` or it does not have a batch-dim.? It is also unclear whether there is a use-case where we need different weights for different samples in the batch (since they have the same mesh).

neuralop/layers/segment_csr.py
juliusberner: Thanks! Are we not using the `logging` module so users can suppress the warning?

neuralop/models/tests/test_fnogno.py
```diff
@@ -41,7 +47,10 @@ def test_fnogno(gno_transform_type, fno_n_modes):
...
 
     # Check output size
-    assert list(out.shape) == [100, out_channels]
+    if gno_batched:
+        assert list(out.shape) == [batch_size, 100, out_channels]
+    else:
+        assert list(out.shape) == [100, out_channels]
 
     # Check backward pass
     loss = out.sum()
```
juliusberner, line 56: We could also set the grad of `f` to True and do the gradient computation only on one sample from the output batch, i.e., `out[0]`. Then we can additionally check that the other samples in the input batch `f[1], ...` are not getting grads.

neuralop/layers/integral_transform.py
dhpitt: That's a good point about the batch. Let me add a test to verify the correct repeat shape for unbatched

dhpitt: On further thought, this PR doesn't use the GNO weights at all - might be better as a separate issue

neuralop/layers/segment_csr.py
juliusberner: Agree! We are also calling the logging module in `mpu/comm.py`

juliusberner (approved):

PR #278, opened 2024-01-19
Fixes a bug in plotting train error

suyashbire1:
Train error was being divided for the second time by n_train in BasicLoggerCallback. This lead to smaller values in plots of train error on wandb plots.


JeanKossaifi:
Thank you @suyashbire1, merging!

PR #291, opened 2024-02-22
sum losses before backward

Luca-DTU:
When the number of output channels is higher than one, the loss.backward() will return an error.
This small fix makes sure that this doesn't happen by calling loss.sum().backward() to make it compatible to multiple outputs

### DIFF
--- a/neuralop/training/trainer.py
+++ b/neuralop/training/trainer.py
@@ -186,7 +186,7 @@ class Trainer:
                 if regularizer:
                     loss += regularizer.loss
                 
-                loss.backward()
+                loss.sum().backward()
                 del out
 
                 optimizer.step()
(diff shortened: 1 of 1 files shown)


Luca-DTU:
This change relates to issue #292

dhpitt (commented):
Hi @Luca-DTU, thanks for the contribution! This issue may arise from the fact that you're not using a loss function that takes the specific shape of your outputs into account. Loss should return a scalar value - what shape are your model outputs?

dhpitt:
It depends on the loss function you're using. If you're using one of the losses from `neuralop.losses`, e.g. `LpLoss`, you need to provide the loss with dimensions over which to reduce. This is controlled by the parameter `reduce_dims`. Take a look in `LpLoss.reduce_all` (line 104 of `neuralop.losses.data_losses.py`).

Luca-DTU:
Thank you for the explanation!

PR #297, opened 2024-02-27
remove torch_scatter and torch_cluster from installs in PyPI publish workflow to avoid cpu issue

dhpitt:


JeanKossaifi:
Good point, thanks @dhpitt !

Issue #280, opened 2024-01-24
[Request] Dataset creation and acquisiton documentation


JoeMWatson:
Hi, 

The current examples look nice, but rely on having a preloaded dataset file. It would be good if there was an example that showed how to translate a given PDE into the required dataset format and some best practice on preprocessing for the model. 

The scripts (at least navier stokes) also seem to assume the dataset ready but there is no documentation of how to get this. 
The script `train_navier_stokes.py` returns
`FileNotFoundError: [Errno 2] No such file or directory: '/data/navier_stokes/nsforcing_128_train.pt'`

Hrrsmjd:
You can download some datasets from https://github.com/pdebench/PDEBench or https://huggingface.co/pdearena

RanieryMendes:
Hi, 

I'm facing a similar issue to @JoeMWatson. I would like to kindly request access or directions to the data used for the experiments reported in Li et. al FNO [paper ](https://arxiv.org/abs/2010.08895).

dhpitt:
The  data is provided in `neuralop.data.datasets` as of #299.

PR #299, opened 2024-03-01
Update datasets to include download, dataset metaclass, dataset objects

dhpitt:
Connected to #280: we previously discussed adding `torchvision`-style dataset files that include the ability to download data for sample problems, e.g. Navier-Stokes or Darcy flow. This PR adds this functionality as well as:

- creates a base `PTDataset` class
- turns existing sample datasets into objects that inherit from PTDataset
- point to data storage hosted publicly on Zenodo and allow users to download data 

The Zenodo data is currently stored on archives owned by the NeuralOperator Zenodo organization, linked here: https://zenodo.org/communities/neuraloperator/records?q=&l=list&p=1&s=10&sort=newest


neuralop/datasets/web_utils.py
```diff
@@ -0,0 +1,151 @@
...
+    with open(fpath, "rb") as f:
+        while chunk := f.read(chunk_size):
+            md5.update(chunk)
+    return md5.hexdigest()
+
+def check_md5(fpath: Union[str, Path], md5: str, **kwargs: Any) -> bool:
+    return md5 == calculate_md5(fpath, **kwargs)
+
+def check_integrity(fpath: Union[str, Path], md5: Optional[str] = None) -> bool:
+    if not os.path.isfile(fpath):
```
JeanKossaifi, line 38: We already import pathlib, let's use it instead of os.path, it is more robust.

neuralop/datasets/web_utils.py
```diff
@@ -0,0 +1,151 @@
...
+        md5 checksum to verify file correctness, by default None
+    size: Optional[int], optional
+        size of the file expected in bytes, to check
+    chunk_size : int, optional
+        size of chunks to use when streaming files
+    extract_tars: bool, optional
+        whether to extract .tgz archives, by default True
+    """
+
+    root = os.path.expanduser(root)
```
JeanKossaifi, line 76: Same, you can do all these with Path (exists, mkdir, etc)

JeanKossaifi (commented):
I'll make a more thorough review later but it looks great at first glance! Let's use pathlib instead of os.path.

neuralop/datasets/navier_stokes.py
```diff
@@ -1,188 +1,77 @@
...
-#                                                         batch_size=test_batch_size,
-#                                                         shuffle=False,
-#                                                         num_workers=num_workers, 
-#                                                         pin_memory=pin_memory, 
-#                                                         persistent_workers=persistent_workers)
-
-#     return train_loader, test_loaders, data_processor
-
-
-def load_navier_stokes_pt(data_path, train_resolution,
```
m4e7, line 70: Before deleting this function we need to replace its usage in `scripts/`, namely:
https://github.com/neuraloperator/neuraloperator/blob/ae6bdb948b1733a8c1bb862de8127c55c97e3074/scripts/train_navier_stokes.py#L74

neuralop/datasets/web_utils.py
```diff
@@ -0,0 +1,151 @@
...
+                    curr_size += chunk_size
+                    prog = curr_size / size
+                    print(f"Download in progress: {prog:.2%}", end='\r')
+
+            assert os.path.getsize(fpath) == size
+
+    # check integrity of downloaded file
+    if not check_integrity(fpath, md5):
+        raise RuntimeError("File not found or corrupted.")
+    else:
```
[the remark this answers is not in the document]
dhpitt: Not sure what you mean by this - the if/else is the difference between raising an error and doing the rest of the file processing

neuralop/datasets/tests/test_web_downloads.py
```diff
@@ -0,0 +1,27 @@
...
+                           batch_size=1,
+                           test_batch_sizes=[1],
+                           train_resolution=resolution,
+                           test_resolutions=[resolution, resolution*2],
+                           download=True)
+    
+    downloaded_files = os.listdir(test_data_dir)
+    for split in ['train','test']:
+        for res in [resolution, resolution*2]:
+            assert f"darcy_{split}_{res}.pt" in downloaded_files
```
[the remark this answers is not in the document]
dhpitt: The MD5 check happens in the course of downloading

neuralop/datasets/pt_dataset.py
```diff
@@ -1,177 +1,139 @@
...
+        test_resolutions, n_tests
+        ):
+            print(
+                f"Loading test db at resolution {res} with {n_test} samples "
+            )
+            data = torch.load(Path(root_dir).joinpath(f"{dataset_name}_test_{res}.pt").as_posix())
+            x_test = (
+                data["x"][:n_test, ...].unsqueeze(channel_dim).type(torch.float32).clone()
+            )
+            y_test = data["y"][:n_test, ...].unsqueeze(channel_dim).clone()
```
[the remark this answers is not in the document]
dhpitt: Comes from the legacy code - my impression is that the typecasting is more necessary for model inputs but @JeanKossaifi may have a better idea

dhpitt:
Also aims to fix #328

neuralop/data/datasets/web_utils.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # check if file is already present locally
+    if check_integrity(fpath, md5):
+        logger.info("Using downloaded and verified file: " + str(fpath))
+        return
+
+    # download and stream file in chunks
+    with requests.get(url, stream=True) as r:
+        r.raise_for_status()
+        confirm = input("Confirm download [Y/n]: ")
```
JeanKossaifi, line 96: One minor comment: I'm not sure we want/should ask for input. This is helpful locally but at scale or on clusters may quickly become a hinderance. 

I'd just make sure we abide by sensitive flags in downstream functions (e.g. `download=True`)

JeanKossaifi (commented):
Looks good, just one comment on prompting users for confirmation. It is helpful locally but at scale or on clusters may quickly become a hinderance.

I'd just make sure we abide by sensitive flags in downstream functions (e.g. `download=True`)

neuralop/data/datasets/web_utils.py
dhpitt: Thanks! Removed them.

JeanKossaifi:
Looks great, thanks David!

PR #301, opened 2024-03-05
Update Car-CFD PR with new trainer code

dhpitt:
This is an update of #249 that factors in the new code that has been integrated since it was opened. Broadly, this PR adds weighted L2 drag loss and an example script that showcases training an `FNOGNO` on the public `car_cfd` dataset, as well as the processing script to coerce the unzipped `car_cfd` dataset into the form expected by the dataset.

Changes to the car CFD dataset may come later as a component of #299 

**Issues**

1. What to do about Open3d dependency? It seems required for this specific datamodule, but not for the functionality of the codebase overall


neuralop/losses/data_losses.py
```diff
@@ -313,101 +313,3 @@ def rel(self, x, y, vol_elm):
     
     def forward(self, y_pred, y, vol_elm, **kwargs):
         return self.rel(y_pred, y, vol_elm)
```
JeanKossaifi, line 315: Why was the rest deleted? 
`self.rel` doesn't seem defined. If we don't need this let's just delete the class alltogether

JeanKossaifi (commented):
Looks good @dhpitt, thank you!

JeanKossaifi:
Looks good to me, merging!

PR #305, opened 2024-03-13
Deepcopy base model

dhpitt:
The models in `neuralop.models` subclass `neuralop.models.BaseModel`, which adds lots of useful functionality but does not implement the `__deepcopy__` method necessary to copy a torch module's architecture and state. This PR adds a method to do that.

### DIFF
--- a/.github/workflows/deploy_pypi.yml
+++ b/.github/workflows/deploy_pypi.yml
@@ -30,8 +30,6 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           python -m pip install -r requirements.txt
-          python -m pip install torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/cpu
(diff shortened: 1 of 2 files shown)


JeanKossaifi:
Wouldn't the default __deepcopy__ work out of the box?

dhpitt:
> Wouldn't the default **deepcopy** work out of the box?
No, it throws an error - there's no way to access the internal module state from base_model by default.

JeanKossaifi (commented):
I'm not sure I get why we need this? 

Also shouldn't we explicitly copy the arguments? From the [doc](https://docs.python.org/3/library/copy.html):

> If the __deepcopy__() implementation needs to make a deep copy of a component, it should call the [deepcopy()](https://docs.python.org/3/library/copy.html#copy.deepcopy) function with the component as first argument and the memo dictionary as second argument. The memo dictionary should be treated as an opaque object.

JeanKossaifi:
Was this resolved @dhpitt ?

PR #313, opened 2024-04-08
Make FNO compatible with 4D inputs

dhpitt:
This PR adds functionality to make the FNO compatible with 4D data. To do so, all `MLP`, `skip_connection`, and `Norm` modules are rewritten to roll up the last dimensions of 4D and higher data into a longer 3D tensor and apply 3D processing. Tests for 4D are also included.


neuralop/layers/mlp.py
```diff
@@ -45,6 +45,8 @@ def __init__(
             else None
         )
 
+        if n_dim > 3:
+            n_dim = 3
```
JeanKossaifi, line 49: Can we do a quick timing comparison on GPU, using 3D conv vs always 1D Conv + reshaping? 
Since the inputs are C-ordered and continguous, it might not make a (big) difference, in which case it will be cleaner to always use e.g Conv1D or Conv2D

neuralop/layers/fno_block.py
```diff
@@ -137,14 +137,22 @@ def __init__(
...
-                    getattr(nn, f"InstanceNorm{self.n_dim}d")(
-                        num_features=self.out_channels
-                    )
-                    for _ in range(n_layers * self.n_norms)
-                ]
-            )
+            if self.n_dim <= 3:
+                self.norm = nn.ModuleList(
+                    [
+                        getattr(nn, f"InstanceNorm{self.n_dim}d")(
```
JeanKossaifi, line 143: I tink the getattr could be a single line here.
This takes a lot of space and makes it a little hard to read.

neuralop/layers/fno_block.py
```diff
@@ -239,7 +249,9 @@ def forward_with_preactivation(self, x, index=0, output_shape=None):
 
         if self.mlp is not None:
             x_skip_mlp = self.mlp_skips[index](x)
-            x_skip_mlp = self.convs[index].transform(x_skip_mlp, output_shape=output_shape)
+            x_skip_mlp = self.convs[index].transform(
```
JeanKossaifi, line 252: Same here I'd keep these on a single line

neuralop/layers/mlp.py
```diff
@@ -44,27 +44,47 @@ def __init__(
...
+        # if x's data is >1d, reshape into 1d b, c, x
+        reshaped = False
+        size = list(x.shape)
+        if x.ndim > 3:  # batch, channels, x1, x2... extra dims
+            # flatten last data dims
+            if x.is_contiguous():
+                # view only works on contiguous tensors but is better
+                x = x.view(*size[:2], -1) 
+            else:
+                x = x.reshape((*size[:2], -1))
```
JeanKossaifi, line 70: That makes a lot of sense, view is ideal. Here we shouldn’t need it though:
I already rely on the data being in C-order to justify using a reshape + Conv1D. This means that a view should be created under the hood.

neuralop/layers/mlp.py
```diff
@@ -44,27 +44,47 @@ def __init__(
...
             if self.dropout is not None:
                 x = self.dropout[i](x)
 
+        # if x was an Nd tensor reshaped into 3d, undo the reshaping
+        if reshaped:
+            # same logic as above: only contiguous tensors can call .view()
+            if x.is_contiguous:
+                x = x.view(size[0], self.out_channels, *size[2:])
+            else:
+                x = x.reshape((size[0], self.out_channels, *size[2:]))
```
JeanKossaifi, line 86: I think here it may be clearer to just use reshape all the time: if it wasn't reshaped, nothing happens. If a view is possible it will be used (most likely, see above, unless the input isn't contiguous), otherwise it will be reshaped. 

We should comment on the logic behind these decisions though!

JeanKossaifi (commented):
This looks great now @dhpitt! Added one more comment re. view vs reshape.

neuralop/layers/normalization_layers.py
```diff
@@ -28,3 +28,26 @@ def forward(self, x):
         weight, bias = torch.split(self.mlp(self.embedding), self.in_channels, dim=0)
 
         return nn.functional.group_norm(x, self.in_channels, weight, bias, eps=self.eps)
+
+class FlattenedInstanceNorm1d(nn.Module):
+    def __init__(self, num_features: int):
```
juliusberner, line 33: Should we allow for the additional `kwargs` of InstanceNorm1d, in particular `affine` and `eps`?

juliusberner: But I think we can also just directly use `torch.nn.functional.instance_norm` (since we anyways take care that the shape is correct and that seems to be the only difference)

juliusberner (commented):
lgtm, but we might not need the flattened instancenorm. Also, we should link https://github.com/neuraloperator/neuraloperator/pull/213

neuralop/layers/mlp.py
dhpitt: Gotcha! Moved to just `reshape` and commented about it

neuralop/layers/normalization_layers.py
dhpitt: I updated the structure to pass in the additional `kwargs`.

PR #319, opened 2024-05-03
Fix docstring for `SpectralConv`

plutonium-239:
Fixed 3 entries for the docstring of `SpectralConv` (`in_channels`, `out_channels`, `n_modes`) 

- There were 2 entries both called `max_n_modes` where the first one should've been `n_modes`.
- Removed the 'optional' (and default args) since they are not defined with any default arguments in the `__init__` function.

### DIFF
--- a/neuralop/layers/spectral_convolution.py
+++ b/neuralop/layers/spectral_convolution.py
@@ -187,11 +187,11 @@ class SpectralConv(BaseSpectralConv):
 
     Parameters
     ----------
-    in_channels : int, optional
+    in_channels : int
         Number of input channels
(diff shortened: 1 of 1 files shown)


dhpitt:
Old PR closed with #356

plutonium-239:
Hi @dhpitt, there are 2 other changes as well not present in the merged PR, which remove the `optional` from the first two parameters because they aren't optional and don't have a default value

I can open a fresh PR with those changes if you want.

dhpitt:
That would be great. Thank you!

plutonium-239:
No problem :)
I've opened #392

PR #320, opened 2024-05-03
Adding the GINO model

dhpitt:
This PR adds a GINO model that can map between functions defined on arbitrary coordinate meshes. 

It also contains a bug fix (oops) for the naive CPU-runner-only implementation of segment_csr that uses `einsum` for an added speed bonus.


JeanKossaifi:
Excellent, thanks @dhpitt! Can we add a small test case for the bug you fixed? I'll merge this so others can start using GINO!

PR #323, opened 2024-05-15
Add script for training UQNO model 

dhpitt:


samaonline:
there are a few legacy comments that could be deleted and I've pointed out to David. otherwise the code structures look good to me.

neuralop/losses/data_losses.py
```diff
@@ -402,6 +401,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+            in the output domain expected to fall within predicted quantiles
+        reduce_dims : int, optional
+            dimensions to reduce when summing, by default 0
+            This loss was formulated for functions with a co-domain in R, 
+            so for now only 0 is supported
+        reductions : str, optional
+            how to apply reduction (sum or mean), by default 'sum'
+
+        References
+        1. Calibrated Uncertainty Quantification for Operator Learning via Conformal Prediction
```
JeanKossaifi, line 422: The references should be a proper section and the reference and actual link. Let's please try to use correct rst syntax whenever possible so it gets rendered correctly.

neuralop/losses/data_losses.py
```diff
@@ -402,6 +401,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+        self.alpha = alpha
+
+        if isinstance(reduce_dims, int):
+            self.reduce_dims = [reduce_dims]
+        else:
+            self.reduce_dims = reduce_dims
+
+        if self.reduce_dims is not None:
+            if isinstance(reductions, str):
+                assert reductions == 'sum' or reductions == 'mean'
```
JeanKossaifi, line 437: Add an error message. I think assert can be used as a function now, which may make it slightly clearer.

neuralop/losses/data_losses.py
```diff
@@ -402,6 +401,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+                for j in range(len(reductions)):
+                    assert reductions[j] == 'sum' or reductions[j] == 'mean'
+                self.reductions = reductions
+
+    def reduce_all(self, x):
+        for j in range(len(self.reduce_dims)):
+            if self.reductions[j] == 'sum':
+                x = torch.sum(x, dim=self.reduce_dims[j], keepdim=True)
+            else:
+                x = torch.mean(x, dim=self.reduce_dims[j], keepdim=True)
```
JeanKossaifi, line 449: Why not store the function directly as the reduction rather than the string, above, since we already do all these checks, and use that function directly?

neuralop/losses/data_losses.py
```diff
@@ -402,6 +401,77 @@ def __call__(self, y_pred, y, h=None, **kwargs):
...
+        y : torch.tensor
+            true pointwise diffs (model pred - ytrue)
+        """
+
+        quantile = 1 - self.alpha
+        y_abs = torch.abs(y)
+        diff = y_abs - y_pred
+        yscale,_ = torch.max(y_abs, dim=0)
+        yscale = yscale + eps
+        ptwise_loss = torch.max(quantile * diff, -(1-quantile) * diff)
```
JeanKossaifi, line 466: pep8

neuralop/models/uqno.py
```diff
@@ -0,0 +1,59 @@
...
+        quantile = self.residual_model(*args, **kwargs)
+        return (solution, quantile)
+    
+    def save_checkpoint(self, save_folder, save_name):
+        # weights of solution model are frozen
+        # save state of residual model for later
+        self.residual_model.save_checkpoint(save_folder, save_name)
+    
+    def from_checkpoint(self, save_folder, save_name):
+        self.residual_model = self.residual_model.from_checkpoint(save_folder, save_name)
```
JeanKossaifi, line 54: Not sure if that's my diff - spacing looks odd

neuralop/models/uqno.py
```diff
@@ -0,0 +1,59 @@
...
+        return self
+    
+    def eval(self):
+        self.base_model.eval()
+        self.residual_model.eval()
```
JeanKossaifi, line 59: If we override eval, shouldn't train also be overridden? 
Can you double check whether this is needed, since BaseModel should be an nn.Module, wouldn't that be done automatically?

scripts/train_uqno_darcy.py
```diff
@@ -0,0 +1,585 @@
...
+        self.device = device
+        self.base_data_processor = self.base_data_processor.to(device)
+        self.residual_normalizer = self.residual_normalizer.to(device)
+        return self
+    
+    def train(self):
+        self.base_data_processor.train()
+    
+    def eval(self):
+        self.base_data_processor.eval()
```
JeanKossaifi, line 331: Same comment as above - let's check if this is needed.

JeanKossaifi (commented):
Left a few comments, looks good to me otherwise.

neuralop/losses/data_losses.py
dhpitt: We should do this for all losses, as its own PR

scripts/train_uqno_darcy.py
dhpitt: added a test

PR #324, opened 2024-05-16
Make `save_checkpoint` and `load_checkpoint` compatible with single-machine DDP

dhpitt:


dhpitt:
Super small fix to allow proper saving and loading checkpoints in single-machine DDP settings. Ensures that only `rank=0` process saves state dicts during `save_training_state`, and all machines load state dicts using `map_location` during `load_training_state`

neuralop/training/callbacks.py
```diff
@@ -401,42 +401,23 @@ def on_train_start(self, *args, **kwargs):
         if self.resume_from_dir:
             saved_modules = [x.stem for x in self.resume_from_dir.glob("*.pt")]
 
-            assert (
-                "best_model_state_dict" in saved_modules
-                or "model_state_dict" in saved_modules
-            ), "Error: CheckpointCallback expects a model state dict named model.pt or best_model.pt."
-
             # no need to handle exceptions if assertion that either model file exists passes
             if "best_model_state_dict" in saved_modules:
```
JeanKossaifi, line 405: I would just build the path and then use if `path.exists()` (using pathlib)

neuralop/training/training_state.py
```diff
@@ -24,12 +25,16 @@ def load_training_state(save_dir: Union[str, Path], save_name: str,
     save_name : str
         name of model to load
     """
+    map_location = None
+    if dist.is_initialized():
+        map_location = {"cuda:0" : f"cuda:{dist.get_rank}"}
```
JeanKossaifi, line 30: I would let the user specify an optional map_location. This may not always work.

neuralop/training/training_state.py
```diff
@@ -71,6 +76,10 @@ def save_training_state(save_dir: Union[str, Path], save_name: str,
     save_name : str
         name of model to load
     """
+    # if running in DDP, only save state on rank 0
+    if dist.is_initialized():
```
JeanKossaifi, line 80: We should not be handling this, at least not here -- the user should be responsible for calling this only on the rank(s) they want.

dhpitt: got it

neuralop/training/callbacks.py
dhpitt: done

PR #327, opened 2024-05-17
Test native `segment_csr` behavior, add documentation for integralTransform to website

dhpitt:
Adds tests for `segment_csr`, cleans up docstrings, updates the trees for autogenerated Sphinx docs website, changes the forward call signature for a `GINO` to take `x` instead of `f` for maximum compatibility.


doc/source/modules/api.rst
```diff
@@ -126,6 +126,34 @@ Neural operator Layers
 
     SphericalConv
 
+**Kernel integration**
```
JeanKossaifi, line 129: Perhaps we add some text around to example the high level goal?

doc/source/modules/api.rst
```diff
@@ -126,6 +126,34 @@ Neural operator Layers
...
+    :template: function.rst
+
+    simple_neighbor_search
+
+.. autosummary::
+    :toctree: generated
+    :template: function.rst
+
+    segment_csr
+
```
JeanKossaifi, line 157: Should we also add GINO to the API?

JeanKossaifi (commented):
Looks good! I think we can just add some text in the API and add GINO too.

doc/source/modules/api.rst
dhpitt: It should be autogenerated in `models`

JeanKossaifi:
Looks like there's a typo that's breaking the tests.

JeanKossaifi:
Looks good, thanks David!

PR #331, opened 2024-05-22
Better UnitGaussianNormalizer testing

dhpitt:
Plus, fixes the unbiased estimator of standard deviation for incremental updates and uses plain `torch.std` when possible for a single data batch.


JeanKossaifi:
This looks great, thanks @dhpitt!

PR #332, opened 2024-05-22
Refactor and create `data` module

dhpitt:
Old structure:

```
neuralop
| --- datasets
        | --- transforms
        | --- data_transforms
        | --- darcy
        | --- etc
```

New structure:

```
neuralop
| --- data
        | --- datasets
                | --- darcy
                | --- etc
        | --- transforms
                | --- normalizers
                | --- etc
```


neuralop/data/transforms/base_transforms.py
```diff
@@ -0,0 +1,136 @@
...
+
+    @abstractmethod
+    def cpu(self):
+        pass
+
+    @abstractmethod
+    def to(self, device):
+        pass
+
+class Composite(Transform):
```
JeanKossaifi, line 34: Should this just be named Compose?

neuralop/data/transforms/normalizers.py
```diff
@@ -242,3 +193,57 @@ def from_dataset(cls, dataset, dim=None, keys=None, mask=None):
                 if key in keys:
                     instances[key].partial_fit(sample.unsqueeze(0))
         return instances
+
+class MultipleFieldUnitGaussianNormalizer(DictTransform):
```
JeanKossaifi, line 197: Should we name this just DictGaussianNormalizer?

neuralop/data/transforms/positional_embeddings.py
```diff
@@ -0,0 +1,224 @@
+from typing import List
+
+import torch
+from torch import nn
+
+from .grid_transforms import regular_grid_2d, regular_grid_nd
+
+class Euclidean2D(nn.Module):
```
JeanKossaifi, line 8: I'd name this just Regular2DGrid, that would be clearer to users.

JeanKossaifi: Or maybe Embedding rather than Grid so it's consistent for sinusoidal case

neuralop/data/transforms/transforms.py
```diff
@@ -73,6 +64,79 @@ def to(self, device):
         self.transforms = [t.to(device) for t in self.transforms if hasattr(t, 'to')]
         return self
 
+class DictTransform(Transform):
```
JeanKossaifi, line 67: Why is this duplicated? We already have it in base_transforms no?

JeanKossaifi (commented):
Thanks for the changes @dhpitt! Left some comments.

neuralop/data/transforms/positional_embeddings.py
dhpitt: What about Regular2DGridEmbedding? Too wordy?

neuralop/data/transforms/transforms.py
dhpitt: Yeah it was, just removed this file

neuralop/data/transforms/positional_embeddings.py
JeanKossaifi: If its Regular it's probably a grid? How about just GridEmbedding? (or GridEmbedding2D)

dhpitt: That sounds good. Updated.

JeanKossaifi:
Looks great, thank you @dhpitt! 
Merging.

matteoguarrera:
docs still need to be fixed :smile:  I opened an issue #341 i can do it if you want guys

PR #338, opened 2024-06-10
Data processor train() and eval() more explicit

dhpitt:
`DefaultDataProcessor` currently has behavior involving `self.train` (which is currently a bound method of `DataProcessor` inherited from `nn.Module`). This fixes that and also sets data procesor train and eval by default within the `Trainer`.


neuralop/datasets/data_transforms.py
```diff
@@ -108,6 +114,12 @@ def forward(self, **data_dict):
         output = self.model(data_dict["x"])
         output = self.postprocess(output)
         return output, data_dict
+    
+    def train(self):
+        self._train = True
+    
+    def train(self):
+        self._train = False
```
JeanKossaifi, line 122: I don't get these two functions - you're redefining? 
We should define `train` as a property. But if it's an nn.Module why not reuse the existing one?

neuralop/datasets/data_transforms.py
```diff
@@ -45,6 +45,11 @@ def wrap(self, model):
     def forward(self, x):
         pass
 
+    def train(self):
```
JeanKossaifi, line 48: I don't see the need for this class. Not only does it not bring feature it actually breaks the default behaviour of the functions (it should at the very least call super). I'd be in favor of removing it alltogether.

dhpitt: update

neuralop/datasets/data_transforms.py
dhpitt: Good idea

neuralop/data/transforms/data_processors.py
```diff
@@ -108,6 +112,17 @@ def forward(self, **data_dict):
         output = self.model(data_dict["x"])
         output = self.postprocess(output)
         return output, data_dict
+        
+    def train(self, val: bool=True):
+        super().train(val)
+        if self.model:
```
JeanKossaifi, line 118: should this be "is not None"?

JeanKossaifi (commented):
Looks good, just one minor point.

neuralop/data/transforms/data_processors.py
dhpitt: Changed.

JeanKossaifi:
Looks good, merging! Thanks @dhpitt !

PR #344, opened 2024-06-21
Update losses to include MSE, Loss base class, track individual components of weighted sum

dhpitt:


neuralop/losses/meta_losses.py
```diff
@@ -62,28 +67,39 @@ def __call__(self, pred: torch.Tensor, truth: torch.Tensor, **kwargs):
...
-    def __init__(self, losses, weights=None):
+    def __init__(self, losses, weights=None, return_individual=True, compute_grads=False):
         super().__init__()
         if weights is None:
             weights = [1.0 / len(losses)] * len(losses)
         if not len(weights) == len(losses):
             raise ValueError("Each loss must have a weight.")
-        self.losses = list(zip(losses, weights))
+        #self.losses = list(zip(losses, weights))
+        self.losses = {x.__name__: [x,y] for x,y in zip(losses,weights)}
```
juliusberner, line 82: why `__name__` instead of `name`?

neuralop/losses/loss.py
```diff
@@ -0,0 +1,21 @@
+"""
+Base Loss class for neuralop
+"""
+from  abc import ABC, abstractmethod
+from torch import nn
+
+class Loss(nn.Module, ABC):
```
juliusberner, line 7: Why does `Loss` need to inherit from `nn.Module`?

neuralop/training/trainer.py
```diff
@@ -197,16 +216,32 @@ def train(
...
-                if self.overrides_loss:
+                if self.overrides_training_loss:
                     loss += self.callbacks.compute_training_loss(
                         out=out, **sample, amp_autocast=self.amp_autocast
                     )
                 else:
-                    if self.amp_autocast:
-                        with amp.autocast(enabled=True):
-                            loss += training_loss(out, **sample)
+                    if self.monitor_individual_train_losses:
```
juliusberner, line 224: An idea would be to always output a loss dictionary (i.e., automatically use the weighted loss, even for a single loss) to simplify the trainer?

neuralop/training/trainer.py
```diff
@@ -197,16 +216,32 @@ def train(
...
+                            with amp.autocast(enabled=True):
+                                loss_outputs = training_loss(out, **sample)
+                        else:
+                            loss_outputs = training_loss(out, **sample)
+
+                        for loss_name, loss_value in loss_outputs.items():
+                            loss += loss_value
+                            train_losses[loss_name] += loss_value
+                            # accumulate training gradients for each subloss over the epoch
+                            if self.compute_train_grads:
```
juliusberner, line 235: This is a nice feature, but we typically do not want to do that in every train step but just periodically.

juliusberner: Theoretically, we could also save one grad computation (i.e., the `.backward`) by aggregating the grads appropriately, but I think it is not necessary if we only compute it periodically.

neuralop/training/trainer.py
```diff
@@ -142,6 +143,15 @@ def train(
 
         if training_loss is None:
             training_loss = LpLoss(d=2)
```
juliusberner, line 145: Should we set the default to `MSE`?

JeanKossaifi: Default Lp loss should be MSE (l2 loss)

dhpitt:
Closing to split this into 2 separate PRs. MSE is simpler and more urgent

Issue #284, opened 2024-02-07
Training logs no longer appear in update 0.3.0


AlwaysRedNeverGreen:
After updating the library to the latest version, the training logs are no longer displayed in the terminal. This change in behavior Is there a new configuration step or parameter that needs to be set to enable logging, or is this an unintended regression?

srinathdama:
Tried setting ```verbose``` to ```True```. Still, the training logs are not being printed! Is it something to do with the callbacks? 

<img width="797" alt="image" src="https://github.com/neuraloperator/neuraloperator/assets/8924450/fd72b717-b1d5-435a-a1ed-4d69ee0fe82e">

AlwaysRedNeverGreen:
I'm not sure what the issue was so I downgraded back to 0.2.0

dhpitt:
Fixed this in #365. Pull the dev version of the library to check it out! New pip release soon too

PR #347, opened 2024-06-21
Update BasicLoggerCallback to function in more situations (eval-only, when `Trainer.verbose=True` but no callbacks passed)

dhpitt:
#284 mentions that logging no longer happens by default because we moved logging to Callbacks. I think a default logging callback would be useful


dhpitt:
Closing and superseding soon with `stateless-callbacks` PR

PR #351, opened 2024-06-25
Better docs for Callbacks

dhpitt:
Improves docstring for `Callback` in `neuralop.training.callbacks.py` and adds a small example to showcase the simplest case of a Callback in training. Also fixes formatting for the IFNO example to be in line with the others (maybe we should update the website to credit all authors of examples/code)


dhpitt:
Closing and superseding soon with `stateless-callbacks` PR

PR #357, opened 2024-07-02
Loss module improvements

dhpitt:
Adds `MSELoss`, refactors finite difference methods out of `data_losses`, adds docstrings to losses, adds tests for all losses


JeanKossaifi:
Thank you @dhpitt !

PR #361, opened 2024-07-05
`WeightedSumLoss` compatible with standard `Trainer`, migrate callbacks to stateless

dhpitt:
This is a WIP. I have not fully decided how to design Callbacks in an abstract, clean and user-friendly way. But this PR contains working code to use loss-balancing from [1] while logging and backpropagating the outputs of a custom `SumLossOutput` object. 

This PR also removes the `BasicLoggerCallback` to simplify the interface of the trainer re: #284 

[1]. Wang et al., 2023. "An Expert's Guide to Training Physics-Informed Neural Networks," https://arxiv.org/pdf/2308.08468

### DIFF
--- a/neuralop/losses/meta_losses.py
+++ b/neuralop/losses/meta_losses.py
@@ -81,17 +81,19 @@ class WeightedSumLoss(object):
 
     def __call__(self, *args, **kwargs):
 
-        output_dict = {}
+        loss_dict = {}
+        weight_dict = {}
         for name, (loss,weight) in self.losses.items():
             loss_value = loss(*args, **kwargs)
-            if self.return_individual:
-                output_dict[name] = (weight, loss_value)
-            else:
+            if self.return_sum:
                 weighted_loss += weight * loss_value
+            else:
+                loss_dict[name] = loss_value
+                weight_dict[name] = weight
         if self.return_sum:
             return weighted_loss
         else:
-            return SumLossOutput(output_dict)
+            return SumLossOutput(loss_dict, weight_dict)
 
     def __str__(self):
         description = "Combined loss: "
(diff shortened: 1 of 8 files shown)


neuralop/training/callbacks.py
```diff
@@ -633,7 +493,30 @@ def grad_explained(self):
...
+    # Wang et al., "An Expert's Guide to Training Physics-Informed Neural Networks"
+    def __init__(self, alpha: float=0.9):
+        super().__init__()
+        self.alpha = alpha
+    
+    def compute_training_loss(self, out, sample, loss, model, **kwargs):
+        loss_vals = loss(out, **sample) # sumlossoutput
+        grad_norms = {}
+        total_grad_norm = 0.
+        for name, val in loss_vals.losses.items():
```
juliusberner, line 512: We do *not* want to auto-balance every step since this is too expensive. We should be able to periodically update the weights of the `WeightedSumLoss` --- is this possible with this callback?

neuralop/losses/meta_losses.py
```diff
@@ -62,28 +65,123 @@ def __call__(self, pred: torch.Tensor, truth: torch.Tensor, **kwargs):
...
 
     def __str__(self):
         description = "Combined loss: "
-        for loss, weight in self.losses:
-            description += f"{loss} (weight: {weight}) "
+        for name, (loss, weight) in self.losses.items():
+            description += f"{name} (weight: {weight}) "
         return description
+    
+class SumLossOutput(dict):
```
juliusberner, line 104: What is the benefit of introducing `SumLossOutput`? 
Could we also just let every loss output the loss as well as additional `metrics`, which could be an empty dict but also, e.g., include individual losses, additional training metrics, etc.?

dhpitt:
Closing for a different API.

PR #293, opened 2024-02-23
Add a transformer neural operator model and an accompanying example for training it on Darcy flow

zijieli-Jlee:
Main changes:

* Add `transformer_no.py` under `models`, which is a Transformer encoder-decoder architecture for learning operator.  Small changes to the `__init__.py` and `model_dispatcher.py` accordingly.
* Add Gaussian Fourier feature and Siren to the `embeddings.py` which is useful for building the neural field (query point embedding) in the Transformer decoder.
* Add a training example `train_darcy_transformer.py` under `scripts`, the corresponding config yaml file is `darcy_transformer_config.yaml`.
* Fixed a bug in previous attention layer test [#290](https://github.com/neuraloperator/neuraloperator/issues/290)

PR #363, opened 2024-07-08
Merge `Transformer-NO` with newest library code, convert wrapper and custom embedding to `DataProcessor` and `GridEmbedding`

dhpitt:
This is an update of #293 that addresses feedback from @JeanKossaifi and @juliusberner . The main difference is an update in the script `train_darcy_transformer.py` that gets the code more in line with the new library's API.

Specifically, instead of creating a custom `ModelWrapper` class and creating custom grid embedding tensors (as well as using an unneeded multigrid-patching callback), I wrote a custom `TransformerNODataProcessor` that does the preprocessing necessary to convert darcy-flow inputs to the form expected by the TransformerNO model.

This model works on irregular meshes, so I also verified that the model's outputs are invariant to the ordering of the input coordinates, and changed the variable names that refer to `grid` to refer to `mesh` instead to disambiguate.


dhpitt:
Closed #363 to transfer branch to my fork to save space in `git clone`. Will re-open.

PR #365, opened 2024-07-09
Add simple logging back into the `Trainer`

dhpitt:
Issue #284 raises an important point - Callbacks are sometimes unnecessarily complex. The `BasicLogger` is extraneous and I think it makes sense to replace this with `Trainer.log_epoch()`.


JeanKossaifi:
Thank you David, merging!

PR #371, opened 2024-07-12
Remove `Callbacks` and replace with a cleaner, cooler and more modular `Trainer`

dhpitt:
This PR modularizes the existing `Trainer` class by splitting functionality up into class methods:

- `train_one_batch`
- `train_one_epoch`
- `evaluate`
- `eval_one_epoch`

These methods are simple and flexible and can easily be overwritten. See `neuralop.training.incremental.IncrementalFNOTrainer` for an example of extending the `Trainer` for @Robertboy18 's Incremental-FNO without any extra stateful classes.

This PR is at an early stage. Feedback would be great.


doc/source/sg_execution_times.rst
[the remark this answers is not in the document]
dhpitt: Good catch, my mistake

neuralop/training/trainer.py
```diff
@@ -130,173 +93,163 @@ def train(
...
+                    loader_metrics = self.evaluate(eval_losses, loader,
+                                            log_prefix=loader_name)                        
+                    all_metrics.update(**loader_metrics)
 
-                # print msg to console and optionally log to wandb
+                # log metrics at eval_interval
                 if self.verbose:
                     lr = None
-                    for pg in optimizer.param_groups:
+                    for pg in self.optimizer.param_groups:
```
JeanKossaifi, line 166: I'm wondering if we'd want to move this to the actual evaluation function (maybe with an additional is_last ?) so the user has the option to i) do it themselves and ii) log images too

neuralop/training/trainer.py
```diff
@@ -130,173 +93,163 @@ def train(
...
-        return all_errors
+        if isinstance(self.scheduler, torch.optim.lr_scheduler.ReduceLROnPlateau):
+            self.scheduler.step(train_err)
+        else:
+            self.scheduler.step()
+
+        epoch_train_time = default_timer() - t1
+
+        train_err /= len(train_loader)
+        avg_loss /= self.n_samples
```
JeanKossaifi, line 244: Separately, (not in the scope of this PR), -perhaps we should remove this averaging and leave it to the loss to normalize -or not. 

Either way, len(train_loader) would be the batch size here, *not* the length of the dataset, we should use the accumulated n_samples

JeanKossaifi (commented):
Looks good overall to me!

neuralop/training/trainer.py
dhpitt: We should still divide by number of batches though, right?

neuralop/training/incremental.py
```diff
@@ -0,0 +1,163 @@
...
+from torch import nn
+
+from .trainer import Trainer
+from ..models import FNO, TFNO
+from ..utils import compute_explained_variance
+
+class IncrementalFNOTrainer(Trainer):
+    """IncrementalFNOTrainer subclasses the Trainer 
+    to implement specific logic for the Incremental-FNO
+    as described in [1]
```
JeanKossaifi, line 14: The reference should be in the appropriate References section of the docstring.

doc/source/index.rst
```diff
@@ -100,7 +100,7 @@ We provide a ``Trainer`` object that automates the logic of a basic neural opera
    trainer = Trainer(model=model, n_epochs=20,
                      data_processor=data_processor,
                      wandb_log=False,
-                     log_test_interval=3,
+                     eval_interval=3,
```
JeanKossaifi, line 103: Should we make this default to 1?

examples/checkpoint_FNO_darcy.py
```diff
@@ -95,20 +88,18 @@
               optimizer=optimizer,
               scheduler=scheduler, 
               regularizer=False, 
-              training_loss=train_loss)
+              training_loss=train_loss, 
+              save_every=1,
```
JeanKossaifi, line 92: Perhaps we don't save by default here? This may create a lot of disk for unsuspecting users. 

I'd be for a clear documentation and no default saving.

neuralop/training/incremental.py
```diff
@@ -0,0 +1,161 @@
...
+
+    References
+    -----------
+    [1] George, R., Zhao, J., Kossaifi, J., Li, Z., and Anandkumar, A. (2024)
```
JeanKossaifi: It still isn't a correct Rest reference, the text above in the description should be `[1]_` to link to this reference. Please check the doc https://docutils.sourceforge.io/docs/user/rst/quickref.html

neuralop/training/tests/test_trainer.py
```diff
@@ -200,12 +176,18 @@ def test_incremental():
...
+        incremental_loss_gap=False,
+        incremental_grad=True,
+        incremental_grad_eps=0.9999,
+        incremental_buffer=5,
+        incremental_max_iter=1,
+        incremental_grad_max_iter=2,)
 
     optimizer = torch.optim.Adam(model.parameters(), 
-                                lr=8e-3, 
+                                lr=3e-4,
```
JeanKossaifi, line 190: I see we changed the default LR - new one seems more reasonable, but can we add some documentation on this and the parameters of incremental FNO generally (either in the doctoring or user guide) @Robertboy18 ?

examples/checkpoint_FNO_darcy.py
```diff
@@ -117,4 +108,5 @@
               optimizer=optimizer,
               scheduler=scheduler, 
               regularizer=False, 
-              training_loss=train_loss)
\ No newline at end of file
+              training_loss=train_loss,
+              resume_from_dir="./checkpoints")
```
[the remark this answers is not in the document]
dhpitt: Just the diff

examples/checkpoint_FNO_darcy.py
dhpitt: That sounds good.

dhpitt: We don't save by default. You have to pass a value to `save_every` for the checkpointing (default is `None`)

config/default_config.yaml
```diff
@@ -66,7 +66,7 @@ default: &DEFAULT
     n_train: 10000
     train_resolution: 128
     n_tests: [2000, 1000, 1000] #, 1000]
-    test_resolutions: [128, 256, 512] #, 1024] 
+    test_resolutions: [128] #, 1024]
```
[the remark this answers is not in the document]
dhpitt: Done

neuralop/training/tests/test_trainer.py
Robertboy18: Sure I can add that!

PR #383, opened 2024-07-31
Rename MLP to ChannelMixing

dhpitt:


config/default_config.yaml
```diff
@@ -108,8 +108,8 @@ original_fno:
     norm: None
     skip: 'linear'
     
-    use_mlp: 0
-    mlp:
+    use_channel_mixing: 0
+    channel_mixing:
```
JeanKossaifi, line 112: Let's be consistent and either nest everywhere or nowhere.

JeanKossaifi (commented):
Let's decide on naming - `ChannelMLP` vs `ChannelMixing`

PR #385, opened 2024-08-01
Make `wandb` an optional dependency

dhpitt:
Changes wandb-related tests and utils to only source if wandb is installed. This should help with pip builds.

### DIFF
--- a/neuralop/tests/test_utils.py
+++ b/neuralop/tests/test_utils.py
@@ -5,10 +5,11 @@ import pytest
 
 # Only import wandb and use if installed
 wandb_available = False
-from importlib.util import find_spec
-if find_spec('wandb') is not None:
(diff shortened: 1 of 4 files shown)


neuralop/tests/test_utils.py
```diff
@@ -2,7 +2,14 @@
 from ..utils import count_model_params, count_tensor_params
 from pathlib import Path
 import pytest
-import wandb
+
+# Only import wandb and use if installed
+wandb_available = False
+from importlib.util import find_spec
+if find_spec('wandb') is not None:
```
JeanKossaifi, line 9: I think it may be better to 
```
try: 
     import wandb 
 except ImportError
    ...
```

I think find_spec will only give tell us the package is found in python path not that it can actually be imported

PR #386, opened 2024-08-01
Add spatial positional embedding to the beginning of FNO by default

dhpitt:
Also updates scripts/`DefaultDataProcessor`'s default options


neuralop/models/fno.py
```diff
@@ -152,6 +164,17 @@ def __init__(
         self.preactivation = preactivation
         self.fno_block_precision = fno_block_precision
 
+        if spatial_pos_embed == "grid":
```
JeanKossaifi, line 167: spatial_pos_embed seems redundant, why not just `positional_embedding`?

config/navier_stokes_config.yaml
```diff
@@ -66,7 +66,7 @@ default: &DEFAULT
     n_tests: [2000, 1000] #, 1000]
     test_resolutions: [128, 1024] #, 1024] 
     test_batch_sizes: [16, 4] #, 1]
-    positional_encoding: True
+    positional_encoding: False
```
JeanKossaifi, line 69: Why was this changed to False?

JeanKossaifi (commented):
We should be consistent between encoding and embedding, and have only one param: if we move it to the model then let's remove the dataset one to avoid confusion.

For the API I would just have one argument that is either a string (e.g. grid which is then used with default params) or an actual module.

neuralop/models/fno.py
dhpitt: In case we add embedding in Fourier space, to avoid ambiguity

config/navier_stokes_config.yaml
dhpitt: Because it happens in the model by default

neuralop/models/fno.py
JeanKossaifi: Let's call it positional_embedding (or encoding) it will be clearer. Fourier space embedding is still being tested and can be named differently.

config/navier_stokes_config.yaml
JeanKossaifi: As I mentioned in the main comments, then let's remove that option alltogether -- there should be only one way to do it

JeanKossaifi (commented):
Responded  in the code

neuralop/models/fno.py
dhpitt: Sounds good

JeanKossaifi:
We should make the regular grid the default to match previous version

JeanKossaifi:
It is much cleaner with this PR, thanks @dhpitt!

PR #388, opened 2024-08-01
Make torch-harmonic an optional dependency

dhpitt:


neuralop/models/fno.py
```diff
@@ -4,7 +4,13 @@
 import torch.nn.functional as F
 
 from ..layers.spectral_convolution import SpectralConv
-from ..layers.spherical_convolution import SphericalConv
+# Only reference SphericalConv if torch_harmonics is built locally
```
JeanKossaifi, line 7: What about just moving the whole sfno logic to an sfno file?

dhpitt: Sounds good. done

neuralop/models/sfno.py
```diff
@@ -0,0 +1,13 @@
+'''
```
JeanKossaifi, line 1: Nitpick for consistency: We use """ generally in the library

neuralop/models/__init__.py
```diff
@@ -1,6 +1,9 @@
 from .fno import TFNO, TFNO1d, TFNO2d, TFNO3d
 from .fno import FNO, FNO1d, FNO2d, FNO3d
-from .fno import SFNO
+# only import SFNO if torch_harmonics is built locally
+from importlib.util import find_spec
```
JeanKossaifi, line 4: Following up on the other PR: let's use try/except everywhere if choose that

Issue #382, opened 2024-07-31
Einsum error using separable Fourier convolution 


HenrikAsmuth:
Hi,

There seems to be an issue in the tensor contraction when using the separable Fourier convolution option (`separable=True`).
The problem occurred when I tested the `plot_FNO_darcy.py` example with Tucker contraction as well as in my own toy project. 
Seems like the Einstein summation is not correctly set.

Below, I only pasted the part of the stacktrace which I deemed most relevant. The full error should be reproducible when adding the above option to the model init in the example.

`Traceback (most recent call last):
  File "C:\Users\Henrik\virtual_environments\neuraloperator\venv\lib\site-packages\tensorly\plugins.py", line 72, in cached_einsum
    expression = OPT_EINSUM_PATH_CACHE[key]
KeyError: 'abcd,fghi,bf,eg,ch,di->aecd,(32, 32, 9, 9),(20, 10, 6),(32, 20),(16, 10),(9, 6)'`
...

` File "C:\Users\Henrik\virtual_environments\venv_pytorch\lib\site-packages\opt_einsum\parser.py", line 354, in parse_einsum_input
    raise ValueError("Number of einsum subscripts must be equal to the " "number of operands.")
ValueError: Number of einsum subscripts must be equal to the number of operands.`

dhpitt:
Hi @HenrikAsmuth , thanks for this issue. When I've seen this error in the past, it's been because the shape of the input is not correct. What input shape are you using in your forward pass?

dhpitt:
Hi @HenrikAsmuth , i looked into this this morning and found a bug in `SpectralConv` that is fixed by #389 . Thank you so much for bringing this to our attention!

HenrikAsmuth:
Thank Dave for the quick fix. Will test it asap once it's merged.

dhpitt:
Merged #389

PR #389, opened 2024-08-01
Fix inscrutable einsum error in SpectralConv.forward when `separable=True`

dhpitt:
Changes to how the `SpectralConv` layer's tensor contraction function is chosen based on `implementation`, `factorization`, and `separable` to avoid einsum symbols and dense contract being incorrect.

Fixes issue #382 

### bug

Normally a `SpectralConv` is created with the option `separable=False` and its weights have shape `(in_channels, out_channels, modes_1, ...)`. When `separable=True`, the weights have shape `(channels, modes_1, ...)`. The indexing of weight into `weight[slices_w]` around lines 440-460 did not take this into account, leading to bugs in either the factorized tensor contraction einsums or in the slicing required before calling `self._contract`. 

### fix
In order to fix this I changed the indexing and slices mentioned above to ensure that slices are the proper length for the correct cases. I also made sure the proper tensor contraction function was being returned by `self.get_contract_fun` and added unit tests for all separable tensor cases.

### DIFF
--- a/neuralop/layers/spectral_convolution.py
+++ b/neuralop/layers/spectral_convolution.py
@@ -151,7 +151,6 @@ def get_contract_fun(weight, implementation="reconstructed", separable=False):
     """
     if implementation == "reconstructed":
         if separable:
-            print("SEPARABLE")
             return _contract_dense_separable
         else:
             return _contract_dense
@@ -215,6 +214,8 @@ class SpectralConv(BaseSpectralConv):
         * If None, all the n_modes are used.
 
     separable : bool, default is True
+        whether to use separable contraction
+        only checked if `implementation` == 'reconstructed'
     init_std : float or 'auto', default is 'auto'
         std to use for the init
     n_layers : int, optional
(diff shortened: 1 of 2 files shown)


neuralop/layers/spectral_convolution.py
```diff
@@ -46,12 +47,6 @@ def _contract_dense(x, weight, separable=False):
         return tl.einsum(eq, x, weight)
 
 
-def _contract_dense_separable(x, weight, separable=True):
```
JeanKossaifi, line 49: Why was this removed?

neuralop/layers/spectral_convolution.py
```diff
@@ -132,7 +127,7 @@ def _contract_tt(x, tt_weight, separable=False):
         return tl.einsum(eq, x, *tt_weight.factors)
 
 
-def get_contract_fun(weight, implementation="reconstructed", separable=False):
```
JeanKossaifi, line 135: Why did you remove the separable arg? How will we choose the right function?

JeanKossaifi (commented):
Looks good overall though the separable function you removed should be a lot more efficient.

neuralop/layers/spectral_convolution.py
dhpitt: All the factorized contraction functions have their own separable arg, which is passed at the call time for `self._contract`. In the reconstructed case, `_contract_dense_separable` is just elem-wise multiplication `x * w` as opposed to using einsum. I can add it back in

PR #392, opened 2024-08-02
Fix `SpectralConv` docs

plutonium-239:
As discussed in #319 
- remove `optional` from in/out channels and n_modes (no default values)
- correct n_modes to `int or int tuple` instead of `None or int tuple`, which seems to have been copied over from `max_n_modes`


dhpitt (approved):
Thanks @plutonium-239 ! Approving and merging.

PR #395, opened 2024-08-06
Add FDM Darcy-flow PINO loss

dhpitt:


JeanKossaifi:
I think we need to spend some time defining the API for these - currently names and docs are not very consistent or helpful.

dhpitt:
Closing for now pending new PINO work from SURFs.

PR #397, opened 2024-08-06
fix normalization in training/eval modes of `DefaultDataProcessor`

dhpitt:


JeanKossaifi:
Looks good, thanks for the fix @dhpitt !

PR #403, opened 2024-08-12
Fix to channel-wise norm by default in `PTDataset`-provided `Normalizer` instances

dhpitt:


JeanKossaifi:
Looks good - let's try not to mix config changes and more fundamental changes when possible for other PRs.

PR #405, opened 2024-08-12
Minimal working `DistributedDataParallel`

dhpitt:
Requires `openmpi` (from `conda-forge`) to be compatible with Slurm. This update changes DDP to use the flexible `torchrun` process launcher instead of older, more complicated wireup strategies.

### DIFF
--- a/neuralop/mpu/comm.py
+++ b/neuralop/mpu/comm.py
@@ -98,56 +98,14 @@ def get_model_parallel_group():
 
 # get 
 def init(config, verbose = False):
-    
-    # set up global and local communicator
-    if config.distributed == "env":
-
-        world_size = int(os.getenv('WORLD_SIZE', 1))
-        world_rank = int(os.getenv('WORLD_RANK', 0))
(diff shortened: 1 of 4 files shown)


neuralop/training/torch_setup.py
```diff
@@ -19,6 +21,10 @@ def setup(config):
         device : torch.device
         is_logger : bool
     """
+    # We use Torchrun for spawning MP jobs, and torchrun auto-initializes env vars. 
+    # Check for override in case config option conflicts with the fact that we are inside a torch distributed elastic run.
+    if os.getenv("LOCAL_RANK") is not None:
+        config.distributed.use_distributed = True
```
JeanKossaifi, line 27: Is that robust? Don't we want users to specify this? Seems risky to override users config without warning.

JeanKossaifi (commented):
Much better with torchrun.

neuralop/mpu/comm.py
wdlctc, line 1: It is very risky to use dist.init_process_group() function without rank and worldsize.

PR #406, opened 2024-08-14
Automate configuring optimizer and scheduler from config in scripts

dhpitt:
Refactors some old repeated logic out of the training recipe scripts and puts it in `neuralop/training/torch_setup.py`

### DIFF
--- a/neuralop/training/torch_setup.py
+++ b/neuralop/training/torch_setup.py
@@ -1,4 +1,7 @@
+from typing import Iterable
+
 import torch
+from torch import nn
 import neuralop.mpu.comm as comm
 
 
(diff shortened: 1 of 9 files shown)


neuralop/training/torch_setup.py
```diff
@@ -91,6 +91,47 @@ def setup(config):
 
     return device, is_logger
 
+def config_opt_and_scheduler(config, model):
```
JeanKossaifi, line 94: This looks too opaque to me : can we just have get_optimizer and get_scheduler methods that explicitly list their options?

neuralop/training/torch_setup.py
```diff
@@ -91,6 +91,47 @@ def setup(config):
...
+            optimizer, T_max=config.opt.scheduler_T_max
+        )
+    elif config.opt.scheduler == "StepLR":
+        scheduler = torch.optim.lr_scheduler.StepLR(
+            optimizer, step_size=config.opt.step_size, gamma=config.opt.gamma
+        )
+    else:
+        raise ValueError(f"Got scheduler={config.opt.scheduler}")
+
+    return optimizer, scheduler
```
JeanKossaifi, line 133: All this can be reused though

JeanKossaifi (commented):
Minor suggestion to make code clearer, looks good otherwise.

neuralop/training/torch_setup.py
dhpitt: Sure, that sounds simpler

dhpitt:
One question @JeanKossaifi - should these functions be in `torch_setup` or should they be in `neuralop.utils`?

dhpitt:
Closing as the code is simple enough as-is.

PR #409, opened 2024-08-15
Makes factorized tensor optional

JeanKossaifi:
Use factorized tensor only if factorization is not None.


dhpitt (commented):
This is great! Could we also change the imports to mirror the structure used for `open3d`, so that people have the option to install or not install tensorly and tltorch?

JeanKossaifi:
Right now I'm also using TensorLy to cache the optimal contraction path and use opt-einsum under the hood - removing this will make it significantly slower.

PR #411, opened 2024-08-15
(fix) Complex-FNO compatible with SFNO and docs

dhpitt:
Makes sure not to pass complex data to an SFNO


neuralop/layers/fno_block.py
```diff
@@ -67,16 +67,22 @@ class FNOBlocks(nn.Module):
             see layers.skip_connections for more details
         complex_data : bool, optional
             whether the FNO's data takes on complex values in space, by default False
+<<<<<<< HEAD
```
JeanKossaifi: Conflict unresolved

JeanKossaifi (commented):
LGTM

PR #414, opened 2024-08-16
Update the quickstart guide on the docs website

dhpitt:


doc/source/user_guide/quickstart.rst
JeanKossaifi, line 1: Can we git mv instead of deleting and recreating to not break history? I don't think we need to rename this file? I also don't see the new links being used - is that coming later?

dhpitt:
Fixed via `git mv`. I combined `index.rst` and the old `quickstart.rst` into one file. Merging once tests pass

PR #415, opened 2024-08-16
Add a GNOBlock layer, use GNOBlock in GINO

dhpitt:
This PR mainly adds a new layer, the `GNOBlock`, which implements a single-layer Graph Neural Operator to compute a kernel integral of a function from an input point cloud to a cloud of output query points. Also updates `GINO` to use `GNOBlock` layers for clarity.


zongyi-li:
Thanks David, It looks good to me.

I guess we can assume d1 = d2 for the two point clouds?

```
        Parameters
        ----------
        y : torch.Tensor of shape [n, d1]
            n points of dimension d1 specifying
            the space to integrate over.
            If batched, these must remain constant
            over the whole batch so no batch dim is needed.
        x : torch.Tensor of shape [m, d2], default None
            m points of dimension d2 over which the
            output function is defined.
```

neuralop/layers/gno_block.py
```diff
@@ -0,0 +1,176 @@
...
+        ----------
+        y : torch.Tensor of shape [n, d1]
+            n points of dimension d1 specifying
+            the space to integrate over.
+            If batched, these must remain constant
+            over the whole batch so no batch dim is needed.
+        x : torch.Tensor of shape [m, d1], default None
+            m points of dimension d1 over which the
+            output function is defined. Must share domain
+            with y if provided
```
JeanKossaifi, line 145: why "if provided" since both are mendatory?

JeanKossaifi (commented):
Some initial comments.

neuralop/layers/gno_block.py
```diff
@@ -0,0 +1,176 @@
...
+        by default independently of this parameter.
+    use_open3d_neighbor_search: bool, optional
+    use_torch_scatter_reduce : bool, optional
+        whether to reduce in integral computation using a function
+        provided by the extra dependency torch_scatter or the slower
+        native PyTorch implementation, by default True
+
+    References
+    -----------
+    _[1]. Neural Operator: Graph Kernel Network for Partial Differential Equations.
```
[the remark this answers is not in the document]
dhpitt: You're right, I had the underscores in the wrong place, thanks!

neuralop/models/gino.py
```diff
@@ -263,25 +261,33 @@ def __init__(
...
-        out_kernel_in_dim += fno_hidden_channels if out_gno_transform_type != 'linear' else 0
-        out_gno_channel_mlp_hidden_layers.insert(0, out_kernel_in_dim)
-        out_gno_channel_mlp_hidden_layers.append(fno_hidden_channels)
-        self.gno_out = IntegralTransform(
-                    channel_mlp_layers=out_gno_channel_mlp_hidden_layers,
-                    channel_mlp_non_linearity=gno_channel_mlp_non_linearity,
-                    transform_type=out_gno_transform_type,
-                    use_torch_scatter=gno_use_torch_scatter
+
+        self.gno_out = GNOBlock(
```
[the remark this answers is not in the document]
dhpitt: Not in my editor, might be how it's being viewed

neuralop/layers/gno_block.py
```diff
@@ -0,0 +1,176 @@
...
+    The kernel integral computed in IntegralTransform 
+    computes one of the following:
+        (a) \int_{A(x)} k(x, y) dy
+        (b) \int_{A(x)} k(x, y) * f(y) dy
+        (c) \int_{A(x)} k(x, y, f(y)) dy
+        (d) \int_{A(x)} k(x, y, f(y)) * f(y) dy
+    
+    Parameters
+    ----------
+    ## TODO @REVIEWERS: how should we define input channels when the GNO can proceed
```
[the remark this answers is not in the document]
dhpitt: Good point, thank you

zongyi-li:
For pos_embedding, maybe we need a way to set embedding dimension. For example, if we use `SinusoidalEmbedding2D` with `num_channels=10`, then the `kernel_in_dim` should be `num_channels*2=20` instead of `coord_dim*2=6`.

dhpitt:
Thank you for the feedback @zongyi-li and @JeanKossaifi . I responded to both your comments with updates. @zongyi-li , the GNOBlock now sets up the integral transform's input channels according to the pos embed dim if one exists, and I added tests to verify. Please see `neuralop.layers.GNOBlock::120-130`

zongyi-li:
Thank you David. It looks good to me!

neuralop/layers/neighbor_search.py
```diff
@@ -1,6 +1,14 @@
 import torch
 from torch import nn
 
+# only install open3d if imported
```
juliusberner, line 4: Should it be "only import if installed"?

neuralop/layers/gno_block.py
```diff
@@ -0,0 +1,202 @@
...
+        kernel_in_dim = self.coord_dim * 2
+        kernel_in_dim_str = "dim(y) + dim(x)"
+
+         # x and y dim will be embedding dim if pos embedding is applied
+        if self.pos_embedding is not None:
+            # Sinusoidal pos_embedding channels are multiplicative
+            if isinstance(self.pos_embedding, SinusoidalEmbedding2D):
+                kernel_in_dim *= self.pos_embedding.out_channels
+            # Grid pos_embedding channels are added to the end of inputs
+            elif isinstance(self.pos_embedding, GridEmbedding2D) or isinstance(self.pos, GridEmbeddingND):
```
juliusberner, line 121: I guess we should raise a `ValueError` if `self.pos_embedding` is an instance of another class?

neuralop/layers/gno_block.py
```diff
@@ -0,0 +1,202 @@
...
+    in_channels : int
+        number of channels in input function. Only used if transform_type
+        is (c) "nonlinear" or (d) "nonlinear_kernelonly"
+    out_channels : int
+        number of channels in output function
+    coord_dim : int
+        dimension of domain on which x and y are defined
+    radius : float
+        radius in which to search for neighbors
+    sinusoidal_coord_embed_dim : int, optional
```
juliusberner, line 43: This seems not to correspond to the signature of the function?

neuralop/layers/gno_block.py
juliusberner, line 121: `self.pos` seems to be undefined?

juliusberner, line 121: I am not sure if `GridEmbedding` makes sense for GNOs, since `x` and `y` are anyways already specifying the coordinates?

dhpitt: Good idea.

neuralop/layers/gno_block.py
dhpitt: Fixed, thanks

neuralop/layers/gno_block.py
dhpitt: This is a good point

neuralop/models/gino.py
```diff
@@ -13,7 +13,9 @@
 from ..layers.gno_block import GNOBlock
 
 class GINO(nn.Module):
-    """GINO: Geometry-informed Neural Operator
+    """GINO: Geometry-informed Neural Operator. Learns a mapping between
+        functions defined over arbitrary coordinate meshes with spectral convolution
```
kazizzad, line 17: they are note defined over arbitrary coordinate, they are presented over arbitrary coordinate meshes. 

Also, with may refer to the functions. Put a period, and say the model carry global integration through spectral convolution.

neuralop/layers/neighbor_search.py
dhpitt: Right, thank you.

neuralop/models/gino.py
```diff
@@ -191,16 +223,39 @@ def __init__(
         if self.fno_norm == "ada_in":
             if fno_ada_in_features is not None:
                 self.adain_pos_embed = SinusoidalEmbedding(in_channels=fno_ada_in_dim,
-                                                           num_frequencies=fno_ada_in_features,
-                                                           embedding_type='transformer',
-                                                           max_positions=gno_embed_max_positions)
+                                                           num_frequencies=fno_ada_in_features, 
+                                                           max_positions=10000,
+                                                           embedding_type='transformer')
```
Xinyi-Li-4869, line 228: I am not sure should we set the embedding_type to be gno_pos_embed_type

neuralop/models/gino.py
```diff
@@ -319,51 +345,8 @@ def latent_embedding(self, in_p, ada_in=None):
...
-        out = out.permute(0, 2, 1)
-        # Project pointwise to out channels
-        #(b, n_in, out_channels)
-        out = self.projection(out).permute(0, 2, 1)  
-
-        return out
     
-    def forward(self,  x, input_geom, latent_queries, output_queries, ada_in=None, **kwargs):
+    def forward(self,  x, input_geom, latent_queries, output_queries, latent_features=None, ada_in=None, **kwargs):
         """forward pass of GNO --> latent embedding w/FNO --> GNO out
```
Xinyi-Li-4869, line 350: I wonder what is the meaning of "latent embedding w/FNO" here?

neuralop/models/gino.py
dhpitt: Good point, changed

neuralop/models/gino.py
dhpitt: Cleaned up a bit, thanks

PR #419, opened 2024-08-20
Small update for updating max_n_modes.

Robertboy18:
Just creating a setter function that can update the max_n_modes.

### DIFF
--- a/neuralop/layers/fno_block.py
+++ b/neuralop/layers/fno_block.py
@@ -170,7 +170,7 @@ class FNOBlocks(nn.Module):
         self.convs = conv_module(
             self.in_channels,
             self.out_channels,
-            self.n_modes,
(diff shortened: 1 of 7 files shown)


JeanKossaifi:
@Robertboy18 as we discussed this needs to be updated to actually expand the parameter size, e.g. using [resize_](https://pytorch.org/docs/stable/generated/torch.Tensor.resize_.html)  (we could also use [set](https://pytorch.org/docs/stable/generated/torch.Tensor.set_.html)_ but it might be a lot more obscure).

JeanKossaifi:
@Robertboy18 are you still working on this?

Robertboy18:
Hey, I'm not working on this anymore, so I'm going to close it out. Thanks for following up!

PR #422, opened 2024-08-21
Add FC more explicitly for dimensions other than 2

vduruiss:
The Fourier Continuation layer is only explicitly written for 2D inputs.

The continuation is done along each dimension separately, so the code could actually be used for 1D inputs or higher dimensional inputs with minor changes.

Could we make the use of the FC layer more explicit for other dimensions?

I suggested a simple code update to have 1D FC, but you might want to make it more general and have the forward take as an input a tuple of dimensions along which the FC is performed.

### DIFF
--- a/neuralop/layers/fourier_continuation.py
+++ b/neuralop/layers/fourier_continuation.py
@@ -63,11 +63,17 @@ class FCLegendre(nn.Module):
 
         return torch.cat((x, ext), dim=-2)
     
+    def extend1d(self, x):
+        return self.extend_left_right(x)
+    
     def extend2d(self, x):
         x = self.extend_left_right(x)
         x = self.extend_top_bottom(x)
 
         return x
     
-    def forward(self, x):
-        return self.extend2d(x)
+    def forward(self, x, dim):
+        if dim == 1:
+            return self.extend1d(x)
+        if dim == 2:
+            return self.extend2d(x)


JeanKossaifi:
This looks good to me @vduruiss - could we add a small unit-test to verify that e.g. a non-periodic input is correctly padded into a periodic one?

vduruiss:
I added a unit-test for the Fourier Continuation layer in 1d and 2d. 

The Fourier Continuation does not repeat the boundary value(s) in the extension, so instead of asserting the periodicity, I asserted that the shapes are correct and that the original values are not modified in the extension. 

I included some code to plot the signals and their extensions to visually verify periodicity, and left that code commented. Here are the plots for reference

Fourier_Continuation

Let me know if that's what you had in mind!

vduruiss:
@JeanKossaifi @dhpitt  
Just bumping this small PR. I don't think there is too much to discuss here, if the unit tests are satisfactory

dhpitt (approved):
This looks good to me, thanks for the reminder @vduruiss

neuralop/layers/tests/test_fourier_continuation.py
```diff
@@ -0,0 +1,84 @@
...
+    f_extend_one_side = Extension(f, dim=1, one_sided=True)
+    f_extend_both_sides = Extension(f, dim=1, one_sided=False)
+    
+    # Check shape
+    assert f_extend_both_sides.shape[-1] == f_extend_one_side.shape[-1] == f.shape[-1] + add_pts
+    # Check values of original signal
+    torch.testing.assert_close(f, f_extend_one_side[...,:length_signal])
+    torch.testing.assert_close(f, f_extend_both_sides[...,add_pts//2:-add_pts//2])
+    
+    # # Plot to verify
```
JeanKossaifi, line 23: Should we turn these plots below into an example?

JeanKossaifi (commented):
This fell off my radar - looks good to me @vduruiss. 
Is it worth turning the commented plots into an example?

dhpitt:
The example builds and looks great. Merging.

PR #434, opened 2024-09-14
Add NeRF positional embedding to `neuralop.layers.embeddings`

dhpitt:
Changes:

- Adds `Embedding` base class with `out_channels` abstractmethod
- Fixes `in_channels` param in all embedding classes for clarity, exposes `out_channels`
- Adds NeRF embedding to sinusoidal embedding
- Refactors both transformer and nerf sine embeddings into `SinusoidalEmbedding` master class with `embedding_type={'nerf', 'transformer'}` option
- Tests sinusoidal embeddings for correctness, adds all embeddings to API doc


neuralop/layers/embeddings.py
```diff
@@ -137,10 +153,64 @@ def forward(self, data, batched=True):
...
+
+    def forward(self, x):
+        """
+        x: torch.Tensor, shape (n_in, out_channels)
+        
+        """
+        assert x.ndim in [2,3], f"Error: expected inputs of shape (batch, n_in, channels)\
+            or (n_in, channels), got inputs with ndim=={x.ndim}, shape={x.shape}"
+        if x.ndim == 2:
+            batched = False
```
Rlin232, line 188: I wonder if we can add some sort of warning here? Just to notify the user in case there was some mistake in their input where there was mistakenly one less dim. Not sure if this is necessary, just a thought.

Rlin232 (commented):
New changes look good to me. The abstraction and annotations definitely make things feel a bit clearer. Just added a quick comment about setting things to batch mode and whether we should notify the user of this.

neuralop/layers/embeddings.py
```diff
@@ -137,35 +162,117 @@ def forward(self, data, batched=True):
...
-        freqs = freqs / (self.num_channels // 2 - (1 if self.endpoint else 0))
-        freqs = (1 / self.max_positions) ** freqs
-        x = x.ger(freqs.to(x.dtype))
-        x = torch.cat([x.cos(), x.sin()], dim=1)
-        return x
-
+        """
+        Parameters 
+        -----------
+        x: torch.Tensor, shape (n_in, out_channels) or (batch, n_in, out_channels)
```
Xinyi-Li-4869, line 241: Just small issues in annotation, it should be in_channels here instead of out_channels. It might be clearer to also clarify the return here.

neuralop/layers/embeddings.py
```diff
@@ -137,35 +162,117 @@ def forward(self, data, batched=True):
...
+            g(p)_k = sin(2^(k) * Pi * p) if i is even
+
+                   = cos(2^(k) * Pi * p) if i is odd
+
+        * 'transformer' for transformer-style encoding.
+            Let z = `p / max_positions`. Then:
+
+            g(p)_k = sin((p / max_positions) ^ {2k / N}) if i is even
+
+                   = cos((p / max_positions) ^ {2k / N}) if i is odd
```
Xinyi-Li-4869, line 194: "if i is even", I wonder what is 'i' here? In my understanding it sould be:
For 0 <= k < 2L:    
        * 'nerf' : NERF-style encoding.  
            g(p)_k = sin(2^(k/2) * Pi * p) if k is even
                   = cos(2^((k-1)/2) * Pi * p) if k is odd

neuralop/layers/embeddings.py
```diff
@@ -137,35 +162,117 @@ def forward(self, data, batched=True):
...
+        freqs = torch.einsum('bij, k -> bijk', x, freqs)
+
+        # shape len(x), 2, len(freqs)
+        freqs = torch.stack((freqs.sin(),freqs.cos()), dim=-1)
+
+        # transpose the inner per-entry matrix and ravel to interleave sin and cos
+        # for nerf, end result is (sin(2^0 * pi * p_0), 
+        #   cos(2^0 * pi * p_0), ...cos(2^k * pi * p_N))
+        # for transformer, end result is (sin(p_0/max_positions^{2k/N}), 
+        #   cos(p_0/max_positions^{2k/N}), ...cos(p_N/max_positions^{2k/N}))
```
Xinyi-Li-4869, line 270: These annotations seem unclear. Maybe it is clearer to write something like:
nerf: (sin(2^0 * pi * p_0), ..., sin(2^num_frequencies * pi * p_{n_in-1})
        cos(2^0 * pi * p_0), ..., cos(2^num_frequencies * pi * p_{n_in-1}))
transformer: (sin(p_0/max_positions^{2*0/n_in}),  ..., sin(p_N/max_positions^{2*num_frequencies/n_in})
                      cos(p_0/max_positions^{2*0/n_in}), ..., cos(p_N/max_positions^{2*num_frequencies/n_in}))

neuralop/layers/embeddings.py
dhpitt: Right, thank you!

neuralop/layers/embeddings.py
dhpitt: I think I'm just going to remove this line and clean up the refs at the top.

neuralop/layers/embeddings.py
dhpitt: Not sure about this. Torch automatically adds a batch dim via `collate_fn` in most dataloaders, so anything not in this form was probably done intentionally by a user. @JeanKossaifi what do you think?

JeanKossaifi:
Thanks @dhpitt for the changes and @Rlin232 @Xinyi-Li-4869 for the reviews!
Looks good to me, let's just keep transformers as the default before merging.

dhpitt:
Fixed `embedding_type='transformer'` to be default. Merging once tests pass.

PR #436, opened 2024-09-16
Added Codomain attention blocks (CODABlocks)

ashiq24:
1. Added one file in layers/codano.py
2. added test script
3. one additional dependency - added ''einops" to requirement.txt


neuralop/layers/codano.py
```diff
@@ -0,0 +1,460 @@
...
+from .fno_block import FNOBlocks
+from .spectral_convolution import SpectralConv1d, SpectralConv2d, SpectralConv3d
+
+
+AffineNormalizer2D = partial(nn.InstanceNorm2d, affine=True)
+AffineNormalizer3D = partial(nn.InstanceNorm3d, affine=True)
+
+# Codomain Attention Blocks
+
+class CODABlocks(nn.Module):
```
Robertboy18, line 18: The overall structure looks great, thanks Ashiq! But can you add some more comments as well? and remove the self.logger statements? we dont usually add that in our layers.

ashiq24: ok.

neuralop/layers/codano.py
```diff
@@ -0,0 +1,462 @@
...
+from einops import rearrange
+import torch
+from torch import nn
+import torch.nn.functional as F
+from .fno_block import FNOBlocks
+from .spectral_convolution import SpectralConv1d, SpectralConv2d, SpectralConv3d
+
+
+AffineNormalizer2D = partial(nn.InstanceNorm2d, affine=True)
+AffineNormalizer3D = partial(nn.InstanceNorm3d, affine=True)
```
JeanKossaifi, line 14: Is it worth defining this? I'd just explicitly pass the affine keyword

neuralop/layers/codano.py
```diff
@@ -0,0 +1,462 @@
...
+        if kqv_non_linear:
+            kqv_activation = non_linearity
+        else:
+            kqv_activation = torch.nn.Identity()
+
+        self.permutation_eq = permutation_eq
+
+        # if self.n_head is not None:
+        #     # recalculating the value of `head_codim`
+        #     self.head_codimension = max(token_codimension // self.n_head, 1)
```
JeanKossaifi, line 109: Can this be removed?

neuralop/layers/codano.py
```diff
@@ -0,0 +1,462 @@
...
+
+        attention = torch.matmul(dprod, v)
+        attention = rearrange(
+            attention,
+            'b a t (d h w) -> b t a d h w',
+            d=self.head_codimension,
+            h=value_x,
+            w=value_y,
+        )
+        attention = rearrange(attention, 'b t a d h w -> (b t) (a d) h w')
```
JeanKossaifi, line 284: Why are we doing two rearrange one after the other on the same variable?

neuralop/layers/codano.py
```diff
@@ -0,0 +1,462 @@
...
+        q = self.Q(xa)
+        v = self.V(xa)
+
+        value_x, value_y = v.shape[-2], v.shape[-1]
+
+        rearrangement = dict(
+            pattern='(b t) (a d) h w -> b a t (d h w)',
+            b=batch_size,
+            a=self.n_head,
+        )
```
JeanKossaifi, line 267: Could we perhaps use permute/reshape directly to not add new dependency to einops?

JeanKossaifi (commented):
It looks good, thanks for the PR @ashiq24!
Left a few comments in the code.

neuralop/layers/codano.py
ashiq24: I have removed these; however, "affine" cannot be passed as a keyword because we are using "Normalizer," which is callable and may not have the affine keyword if it is some custom normalizer function.

neuralop/layers/codano.py
ashiq24: Removed.

neuralop/layers/codano.py
ashiq24: Thanks for noticing. It can be by one rearrange operation.

neuralop/layers/codano.py
ashiq24: Done.

JeanKossaifi:
@dhpitt can you also check after ICLR deadline?

neuralop/layers/codano.py
```diff
@@ -0,0 +1,607 @@
...
+        use_channel_mlp=False,
+        non_linearity=F.gelu,
+        preactivation=False,
+        fno_skip='linear',
+        channel_mlp_skip='soft-gating',
+        channel_mlp_expansion=1.0,
+        separable=False,
+        factorization='tucker',
+        rank=1.0,
+        SpectralConvolution=None,
```
dhpitt, line 64: ideally change this arg to something other than the class name

neuralop/layers/codano.py
```diff
@@ -0,0 +1,607 @@
...
+import torch
+import torch
+from torch import nn
+import torch.nn.functional as F
+from .fno_block import FNOBlocks
+from .spectral_convolution import SpectralConv1d, SpectralConv2d, SpectralConv3d
+
+
+# Codomain Attention Blocks
+class CODABlocks(nn.Module):
```
dhpitt, line 13: I'm confused by this structure. Is the CODABlock a model, or a layer? If not a model, what else is required to turn this block into a full CODA-NO? I would also suggest naming the layer file `coda_blocks` or similar to disambiguate between the block and the model, if a difference exists.

neuralop/layers/codano.py
```diff
@@ -0,0 +1,607 @@
...
+
+        # Co-dimension of each variable/token. The token embedding space is
+        # identical to the variable space, so their dimensionalities are equal.
+        self.token_codimension = token_codimension
+
+        # (maybe) codim of attention from each head
+        self.head_codimension = (head_codimension
+                                 if head_codimension is not None
+                                 else token_codimension)
+        self.n_head = n_head  # number of heads
```
dhpitt, line 88: nit: `n_heads` (plural)

neuralop/layers/codano.py
```diff
@@ -0,0 +1,607 @@
...
+        self.output_scaling_factor = output_scaling_factor
+        self.temperature = temperature
+
+        # K,Q,V operator with or without non_lin
+
+        if kqv_non_linear:
```
dhpitt, line 94: is there a simpler thing we could call this? Is `kqv` the same as attention?

neuralop/layers/codano.py
```diff
@@ -0,0 +1,607 @@
...
+        kwargs["Normalizer"] = Normalizer
+
+        Convolution = kwargs.get("SpectralConvolution")
+        if Convolution is None:
+            Convolution = SpectralConv2d
+        kwargs["SpectralConvolution"] = Convolution
+
+        super().__init__(*args, **kwargs)
+
+    # XXX rewrite comments on TNO*3D
```
dhpitt, line 241: stale comment?

neuralop/layers/codano.py
```diff
@@ -0,0 +1,607 @@
...
+        t = k.size(0) // batch_size  # Compute the time dimension `t`
+        d = k.size(1) // self.n_head  # Compute the dimension `d`
+
+        k = k.view(batch_size, t, self.n_head, d, k.size(-2),
+                   k.size(-1))
+        q = q.view(batch_size, t, self.n_head, d, q.size(-2),
+                   q.size(-1))
+        v = v.view(batch_size, t, self.n_head, d, v.size(-2),
+                   v.size(-1))
+        k = k.permute(0, 2, 1, 3, 4, 5).contiguous()
```
dhpitt, line 262: These permutations make me think this code isn't general for all spatial dimensions. Could you either be more clear about which shapes/dims this is valid for, or change the reshaping/permuting logic to handle all dimensions?

neuralop/layers/codano.py
```diff
@@ -0,0 +1,607 @@
...
+            output.size(3))
+        output = output.view(
+            batch_size,
+            t * output.size(2),
+            output.size(3),
+            output.size(4))
+
+        return output
+
+    def _forward_non_equivariant(self, x):
```
dhpitt, line 354: Would be good to add docstrings explaining the difference between different forward call types

neuralop/models/uno.py
```diff
@@ -237,7 +237,7 @@ def __init__(
             self.domain_padding = DomainPadding(
                 domain_padding=domain_padding,
                 padding_mode=domain_padding_mode,
-                output_scaling_factor=self.end_to_end_scaling_factor,
```
dhpitt, line 240: Is this change necessary/relevant?

dhpitt (commented):
These changes look good overall. I have one overall concern, which is that it isn't clear to me whether the `CODABlocks` is a block, or a full model. If it is just a block, the file should be renamed, and if it is a full model, this file should be split into layer and model files separately. I also think we should move away from separate implementations for 2 and 3d blocks, and implement the forward call using general logic that works for 2,3, and maybe higher-dim cases. Otherwise the code will become messy and harder to understand/maintain. 

few nitpicks about documentation for clarity

neuralop/layers/codano.py
ashiq24: ok.

dhpitt:
Are there plans to introduce a full model using the CodaBlocks, or to add a
training recipe for CODANO?

neuralop/layers/codano.py
ashiq24: ok

neuralop/layers/codano.py
ashiq24: Yes, the notation aligns with the regular attention mechanism.

ashiq24:
We will release the whole model. But that will be done as the next step.

neuralop/layers/codano.py
ashiq24: Assertion added.
The rearrange operation is much clearer using eiops rearrange operation.
I had that in the first place. But it adds a new dependency. But it produces much readable code. As @JeanKossaifi  suggested, I removed that.

neuralop/models/uno.py
ashiq24: Reverted.

neuralop/layers/codano.py
ashiq24: Added.

neuralop/layers/codano.py
ashiq24: removed.

neuralop/layers/codano.py
dhpitt: Sounds good

neuralop/layers/codano.py
dhpitt: So could we call this `nonlinear_attention` or similar?

neuralop/layers/codano.py
dhpitt: Two ideas:

1. Combine CODA 2d and 3D into one layer as I suggested, and make this logic more general (check the reshaping/permuting in `neuralop/models/gino.py`)
2. Could you use `torch.einsum`

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,398 @@
...
+        # codim of attention from each head
+        self.head_codimension = (head_codimension
+                                 if head_codimension is not None
+                                 else token_codimension)
+        self.n_heads = n_heads  # number of heads
+        self.output_scaling_factor = output_scaling_factor
+        self.temperature = temperature
+        self.num_dims = len(n_modes)
+
+        if Normalizer is None:
```
JeanKossaifi, line 98: We should make this consistent with FNO. Let's have one interface for both.

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,398 @@
...
+        )
+        self.V = FNOBlocks(
+            output_scaling_factor=1,
+            conv_module=spectral_convolution,
+            **kqv_args,
+            **common_args,
+        )
+
+        if self.n_heads * self.head_codimension != self.token_codimension:
+            self.proj = FNOBlocks(
```
JeanKossaifi, line 178: This name collides with the projection in FNO (that is an MLP on channels). 

Perhaps we can make it more explicitly that it is an FNOBLock not a simple projection layer.

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,398 @@
...
+                out_channels=codimension_size,
+                n_layers=2,
+                **mixer_args,
+                **common_args,
+            )
+            self.norm1 = Normalizer(codimension_size)
+            self.norm2 = Normalizer(codimension_size)
+            self.mixer_out_normalizer = Normalizer(codimension_size)
+
+    def compute_attention(self, xa, batch_size):
```
JeanKossaifi, line 235: Let's try to give helpful names to the variables - `xa` isn't very clear

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,398 @@
...
+        xa: torch.tensor. Has shape (b * t, d, h, w, ...) where,
+                        b is the batch_size,
+                        t is the number of tokens,
+                        d is the token codimension,
+                        and h, w, .. are the domain dimensions.
+        """
+
+        k = self.K(xa)
+        q = self.Q(xa)
+        v = self.V(xa)
```
JeanKossaifi, line 249: Might make sense to call them explicitly key, value and query but no strong feeling about this.

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,398 @@
...
+        k = k.view(batch_size, t, self.n_heads, d, *k.shape[-self.num_dims:])
+        q = q.view(batch_size, t, self.n_heads, d, *q.shape[-self.num_dims:])
+        v = v.view(batch_size, t, self.n_heads, d, *v.shape[-self.num_dims:])
+
+        # permuating 'b n t d h w -> b t n d h w ...'
+        rearrangement = dict(
+            pattern=f'b t n d {" ".join(einsum_symbols[-i] for i in range(self.num_dims))} -> b n t d {" ".join(einsum_symbols[-i] for i in range(self.num_dims))}')
+        k = torch.einsum(rearrangement["pattern"], k)
+        q = torch.einsum(rearrangement["pattern"], q)
+        v = torch.einsum(rearrangement["pattern"], v)
```
JeanKossaifi, line 269: We don't need an einsum here, it's just a permutation of the 2nd and 3rd dimensions. You can do it with just `torch.transpose(k, 1, 2)`.

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,398 @@
...
+        # resahpe
+        k = k.view(batch_size, self.n_heads, t, d *
+                   math.prod(k.shape[-self.num_dims:]))
```
JeanKossaifi, line 272: This could be simply
```k.view(batch_size, self.n_heads, t, -1)```

File: neuralop/layers/coda_blocks.py [hunk 20]
```diff
@@ -0,0 +1,398 @@
...
+        t = k.size(0) // batch_size  # Compute the number of tokens `t`
+        # Computer per head token codimension `d`
+        d = k.size(1) // self.n_heads
+
+        # reshaping '(b t) (n d) h w -> b n t d h w ...'
+        k = k.view(batch_size, t, self.n_heads, d, *k.shape[-self.num_dims:])
+        q = q.view(batch_size, t, self.n_heads, d, *q.shape[-self.num_dims:])
+        v = v.view(batch_size, t, self.n_heads, d, *v.shape[-self.num_dims:])
+
+        # permuating 'b n t d h w -> b t n d h w ...'
```
JeanKossaifi, line 264: It would be good to have more explanations in the code about what we are doing. What are n t d, etc. It is hard to follow for someone who isn't already familiar with CoDaNO

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,398 @@
...
+        # resahpe
+        k = k.view(batch_size, self.n_heads, t, d *
+                   math.prod(k.shape[-self.num_dims:]))
+        q = q.view(batch_size, self.n_heads, t, d *
+                   math.prod(q.shape[-self.num_dims:]))
+        v = v.view(batch_size, self.n_heads, t, d *
+                   math.prod(v.shape[-self.num_dims:]))
+
+        # attention mechanism
+        dprod = (torch.matmul(q, k.transpose(-1, -2)) /
```
JeanKossaifi, line 279: Why didn't we do the transpose above for k?

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,398 @@
...
+
+        # Reshape from (b, n, t, d * h * w) to (b, n, t, d, h, w, ...)
+        attention = attention.view(
+            attention.size(0),
+            attention.size(1),
+            attention.size(2),
+            d,
+            *xa.shape[-self.num_dims:])
+        attention = torch.einsum(
+            f'b n t d {" ".join(einsum_symbols[-i] for i in range(self.num_dims))} -> b t n d {" ".join(einsum_symbols[-i] for i in range(self.num_dims))}', attention)
```
JeanKossaifi, line 293: Same, this is just a transpose right?

JeanKossaifi (commented):
Left some more comments - let's try to simplify the flow and make it easier for readers to understand what CoDaNO actually does from the code.

neuralop/layers/coda_blocks.py
ashiq24: I didn't understand the comment. Could you please elaborate?

neuralop/layers/coda_blocks.py
ashiq24: done

neuralop/layers/coda_blocks.py
ashiq24: Done

neuralop/layers/coda_blocks.py
ashiq24: yes, done

neuralop/layers/coda_blocks.py
dhpitt: Seconding this. In-line explanations for operations, including einsum notation, can go a long way.

neuralop/layers/coda_blocks.py
dhpitt: why not `tokens`?

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,380 @@
...
+                        t is the number of tokens,
+                        d is the token codimension,
+                        and h, w, .. are the domain dimensions.
+        """
+
+        k = self.Key(tokens)
+        q = self.Query(tokens)
+        v = self.Value(tokens)
+        assert k.size(
+            1) % self.n_heads == 0, "Number of channels in k, q, and v should be divisible by number of heads"
```
ashiq24, line 255: This block can be replaced by
```
batch_size = x.shape[0]
output_shape = x.shape[-self.num_dims:]

assert x.shape[1] % self.token_codimension == 0

xa = rearrange(x, f'b (k d) {" ".join(f"d{i}" for i in range(self.num_dims))} -> (b k) d {" ".join(f"d{i}" for i in range(self.num_dims))}', d=self.token_codimension)
```

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,380 @@
...
+                        b is the batch_size,
+                        t is the number of tokens,
+                        d is the token codimension,
+                        and h, w, .. are the domain dimensions.
+        """
+
+        k = self.Key(tokens)
+        q = self.Query(tokens)
+        v = self.Value(tokens)
+        assert k.size(
```
ashiq24, line 254: This block can be replaced by
```
batch_size = x.shape[0]
output_shape = x.shape[-self.num_dims:]

assert x.shape[1] % self.token_codimension == 0

xa = rearrange(x, f'b (k d) {" ".join(f"d{i}" for i in range(self.num_dims))} -> (b k) d {" ".join(f"d{i}" for i in range(self.num_dims))}', d=self.token_codimension)
```

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,380 @@
...
+        q = q.view(batch_size, t, self.n_heads, d, *q.shape[-self.num_dims:])
+        v = v.view(batch_size, t, self.n_heads, d, *v.shape[-self.num_dims:])
+
+        k = torch.transpose(k,1,2)
+        q = torch.transpose(q,1,2)
+        v = torch.transpose(v,1,2) 
+        # resahpe
+        k = k.view(batch_size, self.n_heads, t, -1)
+        q = q.view(batch_size, self.n_heads, t, -1)
+        v = v.view(batch_size, self.n_heads, t, -1)
```
ashiq24, line 273: This block can be replaced by (lines 254-273) with rearrange
```python
batch_size = x.shape[0]
output_shape = x.shape[-self.num_dims:]

assert x.shape[1] % self.token_codimension == 0

xa = rearrange(x, f'b (k d) {" ".join(f"d{i}" for i in range(self.num_dims))} -> (b k) d {" ".join(f"d{i}" for i in range(self.num_dims))}', d=self.token_codimension)
```

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,380 @@
...
+            attention.size(0),
+            attention.size(1),
+            attention.size(2),
+            d,
+            *tokens.shape[-self.num_dims:])
+        attention = torch.transpose(attention, 1,2) 
+        attention = attention.view(attention.size(0) * attention.size(1),
+                                   attention.size(2) * attention.size(3),
+                                   *attention.shape[-self.num_dims:])  # shape (b * t, a * d, h, w, ...)
+
```
ashiq24, line 293: Lines 283-293 can be replaced by the following with rearrange.

```python 
attention = torch.matmul(dprod, v)
rearrange_args = dict(
            pattern=f'b a k (d {" ".join(f"d{i}" for i in range(self.num_dims))}) -> (b k) (a d) {" ".join(f"d{i}" for i in range(self.num_dims))}',
            d=self.head_codimension,
        )
rearrange_args.update({f'd{i}': v_shape[i] for i in range(self.num_dims)})
attention = rearrange(attention, **rearrange_args)```

File: neuralop/layers/coda_blocks.py [hunk 27]
```diff
@@ -0,0 +1,380 @@
...
+            self.norm1 = Normalizer(codimension_size)
+            self.norm2 = Normalizer(codimension_size)
+            self.mixer_out_normalizer = Normalizer(codimension_size)
+
+    def compute_attention(self, tokens, batch_size):
+        """
+        Compute the key-query-value variant of the attention matrix for input token functions tokens.
+
+        Assumes input ``tokens`` has been normalized.
+        tokens: torch.tensor. Has shape (b * t, d, h, w, ...) where,
```
dhpitt, line 244: formatting here is a bit odd, can you indent less, use line breaks, and/or use asterisk bullet points? Otherwise these will all render as one line in rst

neuralop/layers/coda_blocks.py
dhpitt: This isn't much more clear or readable, it's all format strings.

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,380 @@
...
+            d,
+            *tokens.shape[-self.num_dims:])
+        attention = torch.transpose(attention, 1,2) 
+        attention = attention.view(attention.size(0) * attention.size(1),
+                                   attention.size(2) * attention.size(3),
+                                   *attention.shape[-self.num_dims:])  # shape (b * t, a * d, h, w, ...)
+
+        return attention
+
+    def forward(self, x, output_shape=None):
```
dhpitt, line 296: Unused arg

neuralop/layers/tests/test_codano.py
```diff
@@ -0,0 +1,92 @@
+import pytest
```
dhpitt, line 1: Could we rename to `test_coda_blocks`?

neuralop/layers/tests/test_codano.py
```diff
@@ -0,0 +1,92 @@
...
+import torch
+from ..coda_blocks import *
+from ..spectral_convolution import *
+
+
+@pytest.mark.parametrize('token_codimension',
```
dhpitt, line 7: you could parametrize `n_dim = [2,3]` quite simply. see the FNO or GINO tests

neuralop/layers/tests/test_codano.py
```diff
@@ -0,0 +1,92 @@
...
+                         [1, 2, 5])
```
dhpitt, line 8: Are there other parameter values you think are important to test?

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,380 @@
...
+
+        if norm is None:
+            Normalizer = torch.nn.Identity
+        elif norm == "instance_norm":
+            Normalizer = partial(nn.InstanceNorm2d, affine=True) if self.num_dims == 2 else partial(
+                nn.InstanceNorm3d, affine=True)
+        else:
+            raise ValueError(f"Unknown normalization type {norm}")
+
+        if spectral_convolution is None:
```
dhpitt, line 106: Have you tested this layer with `SphericalConv` blocks? Good to keep this option if so, but if not, better to fix the conv module.

dhpitt (commented):
Looks nearly good to go, thanks @ashiq24 ! I left a few comments.

neuralop/layers/coda_blocks.py
ashiq24: `spectral_convolution` is passed as argument. So if required, one can pass `SphericalConv`.
What additional 'option' are you suggesting?

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,380 @@
...
+import torch.nn.functional as F
+from .fno_block import FNOBlocks
+from .spectral_convolution import SpectralConv
+
+einsum_symbols = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
+# Codomain Attention Blocks
+
+
+class CODABlocks(nn.Module):
+    """ Class for the Co-domain Attention Blocks (CODABlocks)
```
dhpitt, line 17: Our docs use NumpyDoc, can you reformat this into a [Numpy docstring](https://numpydoc.readthedocs.io/en/latest/format.html)?

ashiq24: Reformated.

dhpitt (commented):
@ashiq24 This mostly looks good. I updated the argument order and docstring ([PR here](https://github.com/neuraloperator/neuraloperator/pull/462)), and also fixed the tests (a lot of old/stale arguments were getting passed in and folded into kwargs/never used). 

One last question before approval. Should the FNO Block hidden channels be accessible as a parameter? The default CODABlock creates an extreme memory demand and cannot be sized down.

dhpitt:
Superseded by #462 (merged)

PR #447, opened 2024-10-04
(fix2) file naming convention for GitHub Pages

dhpitt:
Tests didn't uncover that GH pages requires a specially-named `index` file, which plain Sphinx does not. This fixes the broken page on the site


JeanKossaifi:
I don't think we should have renamed the file, just added an index with the links we want, including quickstart. I believe sphinx needs it anyway to not have orphan pages. I thought I'd added a one previously.

PR #452, opened 2024-10-15
Add `CarCFDDataset` with downloadable Zenodo record

dhpitt:
Also updates the car cfd script to be named explicitly for FNOGNO (to prep for adding a GINO example script)


neuralop/data/datasets/mesh_datamodule.py
```diff
@@ -3,7 +3,15 @@
...
 import numpy as np
-import open3d as o3d
+
+# import open3d for io if built. Otherwise,
+# the class will build, but no files will be loaded.
+try:
+    import open3d as o3d
+    open3d_available = True
+except ModuleNotFoundError:
+    open3d_available = False
```
JeanKossaifi, line 13: I don't think we need the additional flag: we're only wanting to import to extract the doc. If a user runs the code without the dependency, let's let it error rather than fail later.

JeanKossaifi (commented):
Looks great, just one comment on import.

neuralop/data/datasets/mesh_datamodule.py
dhpitt: I think importing the module without the dependency fails

JeanKossaifi:
Thanks, merging!

PR #454, opened 2024-10-21
Fix inline refs in `References` docstring sections

dhpitt:


dhpitt:
No code changes. Just doc updates to make everything render properly :)

PR #455, opened 2024-10-22
Add more references to all Fourier layers/models, discussion and figure of FFT shift in fno

dhpitt:


JeanKossaifi:
Thanks @dhpitt - looks good to me after the small changes we discussed + high res version of the image.

PR #456, opened 2024-10-24
Bbonev/local neuraloperators

bonevbs:
Implements local integral neural operator layers via the discrete-continuous (DISCO) convolution.

### DIFF
--- a/examples/plot_DISCO_convolutions.py
+++ b/examples/plot_DISCO_convolutions.py
@@ -1,9 +1,10 @@
 """
-Training a SFNO on the spherical Shallow Water equations
+Visualization of discrete-continuous convolutions
 ==========================================================
 
-In this example, we demonstrate how to use the small Spherical Shallow Water Equations example we ship with the package
-to train a Spherical Fourier-Neural Operator
+In this example, we demonstrate the usage of the discrete-continuous (DISCO) convolutions
+used in the localized neural operator framework. These modules can be used on both equidistant
+and unstructured grids.
 """
 
 # %%
--- a/examples/plot_DISCO_convolutions.py
+++ b/examples/plot_DISCO_convolutions.py
@@ -30,8 +30,8 @@ device = torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')
 # Let's start by loading an example image
 os.system("curl https://upload.wikimedia.org/wikipedia/commons/thumb/d/d3/Albert_Einstein_Head.jpg/360px-Albert_Einstein_Head.jpg -o ./einstein.jpg")
 
-nx = 90
-ny = 120
+nx = 120
+ny = 180
 
 img = iio.imread('./einstein.jpg')
 data = nn.functional.interpolate(torch.from_numpy(img).unsqueeze(0).unsqueeze(0), size=(ny,nx)).squeeze()
@@ -77,8 +77,8 @@ plt.show()
 
 # %%
 # For the convolution output we require an output mesh
-nxo = 45
-nyo = 60
+nxo = 120
+nyo = 180
 
 x_out = torch.linspace(0, 2, nxo)
 y_out = torch.linspace(0, 3, nyo)
@@ -105,18 +105,25 @@ psi = conv.get_psi()
 # in order to compute the convolved image, we need to first bring it into the right shape with `batch_size x n_channels x n_grid_points`
 out = conv(data.reshape(1, 1, -1))
 
+# plt.figure(figsize=(4,6), )
+# plt.tripcolor(grid_out[0], grid_out[1], out.squeeze().detach(), cmap=cmap, shading="flat")
+# plt.colorbar()
+# plt.xlim(0,2)
+# plt.ylim(0,3)
+# plt.show()
+
 plt.figure(figsize=(4,6), )
-plt.tripcolor(grid_out[0], grid_out[1], out.squeeze().detach(), cmap=cmap, shading="flat")
+plt.imshow(torch.flip(out.squeeze().detach().reshape(nxo, nyo).transpose(0,1), dims=(-2, )), cmap=cmap)
 plt.colorbar()
-plt.xlim(0,2)
-plt.ylim(0,3)
 plt.show()
+
+out1= torch.flip(out.squeeze().detach().reshape(nxo, nyo).transpose(0,1), dims=(-2, ))
 
 # %% do the same but on an equidistant grid:
 
 from neuralop.layers.discrete_continuous_convolution import EquidistantDiscreteContinuousConv2d
 
-conv_equi = EquidistantDiscreteContinuousConv2d(1, 1, (nx, ny), (nxo, nyo), kernel_shape=[2,4], radius_cutoff=3/nyo)
+conv_equi = EquidistantDiscreteContinuousConv2d(1, 1, (nx, ny), (nxo, nyo), kernel_shape=[2,4], radius_cutoff=3/nyo, domain_length=[2,3])
 
 # initialize a kernel resembling an edge filter
 w = torch.zeros_like(conv.weight)
@@ -124,7 +131,7 @@ w[0,0,1] = 1.0
 w[0,0,3] = -1.0
 conv_equi.weight = nn.Parameter(w)
 
-data = nn.functional.interpolate(torch.from_numpy(img).unsqueeze(0).unsqueeze(0), size=(120,90)).float()
+data = nn.functional.interpolate(torch.from_numpy(img).unsqueeze(0).unsqueeze(0), size=(nyo,nxo)).float()
 
 out_equi = conv_equi(data)
 
@@ -133,7 +140,21 @@ plt.imshow(out_equi.squeeze().detach(), cmap=cmap)
 plt.colorbar()
 plt.show()
 
+out2 = out_equi.squeeze().detach()
+
 # %%
 
 plt.figure(figsize=(4,6), )
 plt.imshow(conv_equi.get_psi()[0].detach(), cmap=cmap)
+plt.colorbar()
+
+# %%
+
+print(out1.shape)
+
+print(out2.shape)
+
+plt.figure(figsize=(4,6), )
+plt.imshow(out1 - out2, cmap=cmap)
+plt.colorbar()
+plt.show()
--- a/examples/plot_DISCO_convolutions.py
+++ b/examples/plot_DISCO_convolutions.py
@@ -92,7 +92,7 @@ grid_out = torch.stack([x_out.reshape(-1), y_out.reshape(-1)])
 from neuralop.layers.discrete_continuous_convolution import DiscreteContinuousConv2d, DiscreteContinuousConvTranspose2d
 
 # initialize convolution module
-conv = DiscreteContinuousConv2d(1, 1, grid_in=grid_in, grid_out=grid_out, quad_weights=quad_weights, kernel_shape=[2,4], radius_cutoff=3/90, periodic=False).float()
+conv = DiscreteContinuousConv2d(1, 1, grid_in=grid_in, grid_out=grid_out, quad_weights=quad_weights, kernel_shape=[2,4], radius_cutoff=3/nyo, periodic=False).float()
 
 # initialize a kernel resembling an edge filter
 w = torch.zeros_like(conv.weight)
--- a/examples/plot_DISCO_convolutions.py
+++ b/examples/plot_DISCO_convolutions.py
@@ -18,7 +18,7 @@ import numpy as np
 import math
 from functools import partial
 
-import imageio.v3 as iio
+from matplotlib import image
 
 from torch_harmonics.quadrature import legendre_gauss_weights, lobatto_weights, clenshaw_curtiss_weights
 
@@ -36,7 +36,7 @@ os.system("curl https://upload.wikimedia.org/wikipedia/commons/thumb/d/d3/Albert
 nx = 90
 ny = 120
 
-img = iio.imread('./einstein.jpg')
+img = image.imread('./einstein.jpg')
 data = nn.functional.interpolate(torch.from_numpy(img).unsqueeze(0).unsqueeze(0), size=(ny,nx)).squeeze()
 plt.imshow(data, cmap=cmap)
 plt.show()
--- a/examples/plot_DISCO_convolutions.py
+++ b/examples/plot_DISCO_convolutions.py
@@ -111,3 +111,20 @@ plt.tripcolor(grid_out[0], grid_out[1], out.squeeze().detach(), cmap=cmap, shadi
 plt.xlim(0,2)
 plt.ylim(0,3)
 plt.show()
+
+# %% do the same but on an equidistant grid:
+
+from neuralop.layers.discrete_continuous_convolution import EquidistantDiscreteContinuousConv2d
+
+conv = EquidistantDiscreteContinuousConv2d(1, 1, (90, 120), (60, 120), kernel_shape=[2,4], radius_cutoff=0.1)
+
+# initialize a kernel resembling an edge filter
+w = torch.zeros_like(conv.weight)
+w[0,0,1] = 1.0
+w[0,0,3] = -1.0
+conv.weight = nn.Parameter(w)
+
+data = nn.functional.interpolate(torch.from_numpy(img).unsqueeze(0).unsqueeze(0), size=(120,90)).float()
+
+plt.figure(figsize=(4,6), )
+plt.imshow(conv(data)[0,0].detach().cpu(), cmap=cmap)
--- a/examples/plot_DISCO_convolutions.py
+++ b/examples/plot_DISCO_convolutions.py
@@ -1,0 +1,113 @@
+"""
+Training a SFNO on the spherical Shallow Water equations
+==========================================================
+
+In this example, we demonstrate how to use the small Spherical Shallow Water Equations example we ship with the package
+to train a Spherical Fourier-Neural Operator
+"""
+
+# %%
+#
+
+
+import os
+import torch
+import torch.nn as nn
+import numpy as np
+import math
+from functools import partial
+
+import imageio.v3 as iio
+
+from torch_harmonics.quadrature import legendre_gauss_weights, lobatto_weights, clenshaw_curtiss_weights
+
+import matplotlib.pyplot as plt
+
+cmap="inferno"
+device = torch.device('cuda:0' if torch.cuda.is_available() else 'cpu')
+
+# %%
+# Let's start by loading an example image
+os.system("curl https://upload.wikimedia.org/wikipedia/commons/thumb/d/d3/Albert_Einstein_Head.jpg/360px-Albert_Einstein_Head.jpg -o ./einstein.jpg")
+
+nx = 90
+ny = 120
+
+img = iio.imread('./einstein.jpg')
+data = nn.functional.interpolate(torch.from_numpy(img).unsqueeze(0).unsqueeze(0), size=(ny,nx)).squeeze()
+plt.imshow(data, cmap=cmap)
+plt.show()
+
+# %%
+# Let's create a grid on which the data lives
+
+x_in = torch.linspace(0, 2, nx)
+y_in = torch.linspace(0, 3, ny)
+
+x_in, y_in = torch.meshgrid(x_in, y_in)
+grid_in = torch.stack([x_in.reshape(-1), y_in.reshape(-1)])
+
+# compute the correct quadrature weights
+# IMPORTANT: this needs to be done right in order for the DISCO convolution to be normalized proeperly
+w_x = 2*torch.ones_like(x_in) / nx
+w_y = 3*torch.ones_like(y_in) / ny
+quad_weights = (w_x * w_y).reshape(-1)
+
+# %%
+# Visualize the grid
+
+plt.figure(figsize=(4,6), )
+plt.scatter(grid_in[0], grid_in[1], s=0.2)
+plt.xlim(0,2)
+plt.ylim(0,3)
+plt.show()
+
+
+# %%
+# Format data into the same format and plot it on the grid
+
+data = data.permute(1,0).flip(1).reshape(-1)
+
+plt.figure(figsize=(4,6), )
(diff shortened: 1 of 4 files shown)


examples/plot_DISCO_convolutions.py
```diff
@@ -0,0 +1,200 @@
+"""
+Training a SFNO on the spherical Shallow Water equations
```
dhpitt, line 2: I think this title and caption are incorrect

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,611 @@
...
+        out = torch.einsum("bgckx,gock->bgox", x, self.weight.reshape(self.groups, -1, self.weight.shape[1], self.weight.shape[2]))
+        out = out.reshape(out.shape[0], -1, out.shape[-1])
+
+        if self.bias is not None:
+            out = out + self.bias.reshape(1, -1, 1)
+
+        return out
+
+
+class DiscreteContinuousConvTranspose2d(DiscreteContinuousConv):
```
dhpitt, line 244: could we replace all instances of `DiscreteContinuous` with DISCO? The names are quite wordy.

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,611 @@
...
+
+    idx = idx.permute(1, 0)
+
+    if normalize:
+        vals = _normalize_convolution_tensor_2d(idx, vals, grid_in, grid_out, kernel_shape, quad_weights, transpose_normalization=transpose_normalization)
+
+    return idx, vals
+
+
+class DiscreteContinuousConv2d(DiscreteContinuousConv):
```
dhpitt, line 104: Are DISCO convs only defined for 2d currently?

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,611 @@
...
+    """
+    Discrete-continuous convolutions (DISCO) on arbitrary 2d grids as implemented for [1].
+
+    [1] Liu-Schiaffini M., Berner J., Bonev B., Kurth T., Azizzadenesheli K., Anandkumar A.; Neural Operators with Localized Integral and Differential Kernels;  arxiv:2402.16845
```
dhpitt, line 108: You can add a references section to the docstring, e.g.

```
References
------------

.. [1] (paper 1...)

etc
```

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,611 @@
...
+    [2] Ocampo J., Price M.A. , McEwen J.D.; Scalable and equivariant spherical CNNs by discrete-continuous (DISCO) convolutions, ICLR (2023), arXiv:2209.13603
+
+    Parameters
+        ----------
```
dhpitt, line 112: Remove indent here

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,611 @@
...
+        n_out: Optional[Tuple[int]] = None,
+        quad_weights: Optional[torch.Tensor] = None,
+        periodic: Optional[bool] = False,
+        groups: Optional[int] = 1,
+        bias: Optional[bool] = True,
+        radius_cutoff: Optional[float] = None,
+    ):
+        super().__init__(in_channels, out_channels, kernel_shape, groups, bias)
+
+        # the instantiator supports convenience constructors for the input and output grids
```
dhpitt, line 157: This wording could be a little clearer - what does this mean?

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,611 @@
...
+    idx = idx.permute(1, 0)
+
+    if normalize:
+        vals = _normalize_convolution_tensor_2d(idx, vals, grid_in, grid_out, kernel_shape, quad_weights, transpose_normalization=transpose_normalization)
+
+    return idx, vals
+
+
+class DiscreteContinuousConv2d(DiscreteContinuousConv):
+    """
```
dhpitt, line 105: Also would it be possible to add a short blurb describing the operation of a DISCO conv?

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,611 @@
...
+
+class DiscreteContinuousConvTranspose2d(DiscreteContinuousConv):
+    """
+    Transpose Discrete-continuous convolutions (DISCO) on arbitrary 2d grids as implemented for [1]
+
+    [1] Liu-Schiaffini M., Berner J., Bonev B., Kurth T., Azizzadenesheli K., Anandkumar A.; Neural Operators with Localized Integral and Differential Kernels;  arxiv:2402.16845
+    [2] Ocampo J., Price M.A. , McEwen J.D.; Scalable and equivariant spherical CNNs by discrete-continuous (DISCO) convolutions, ICLR (2023), arXiv:2209.13603
+
+    Parameters
+        ----------
```
dhpitt, line 252: same nit here, no indent

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,611 @@
...
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        # pre-multiply x with the quadrature weights
+        x = self.quad_weights * x
+
+        psi = self.get_psi()
+
+        # extract shape
+        B, C, _ = x.shape
+
+        # bring into the right shape for the bmm and perform it
```
dhpitt, line 370: shape annotations, or a description of "the right shape for the bmm" might be helpful

dhpitt (commented):
Thanks for these changes @bonevbs ! I left a few style/documentation comments.

One broader comment that I think is worth discussing. How much of the function of the different types of DISCO conv layer can be abstracted into a single class and a unified forward pass? I wonder if the logic for different grid types could be refactored to make the choice of which DISCO conv to use parameterizable within one class, like

`DiscreteContinuousConvolution(mode={'equidistant', 'transpose', ...}, ...)`

neuralop/layers/discrete_continuous_convolution.py
bonevbs: I think this is a matter of taste. How about creating aliases for those? The reasoning is that writing it out helps when printing model.

neuralop/layers/discrete_continuous_convolution.py
bonevbs: yes

neuralop/layers/discrete_continuous_convolution.py
bonevbs: which wording exactly? Do you mean the variables in the `super().__init__` call?

bonevbs:
> Thanks for these changes @bonevbs ! I left a few style/documentation comments.
> 
> One broader comment that I think is worth discussing. How much of the function of the different types of DISCO conv layer can be abstracted into a single class and a unified forward pass? I wonder if the logic for different grid types could be refactored to make the choice of which DISCO conv to use parameterizable within one class, like
> 
> `DiscreteContinuousConvolution(mode={'equidistant', 'transpose', ...}, ...)`
> 
> Another thing is I suggest adding a unit-test file to accompany this, with some small sanity checks to make sure the forward pass works across a range of parameters (see `neuralop/models/tests/test_fno.py` or similar)

This is why we have the abstract base class `DiscreteContinuousConv` from which we derive all of these ops. This is similar in style to how PyTorch handles things, as the unifiable parts are either in the functional subroutines `_normalize_convolution_tensor_2d, _precompute_convolution_tensor_2d` or in the `__init__` call of the abstract base class.

I would advise agains unifying further as this will make code less readable. Moreover despite equidistant being a special case of the more general unstructured version, there is a lot of extra work that goes  into making it fit the convolution computation and we cannot guarantee that the two things will give matching results. So for this reason alone, I would like to keep the 4 variants separate.

examples/plot_DISCO_convolutions.py
```diff
@@ -0,0 +1,200 @@
...
+from neuralop.layers.discrete_continuous_convolution import DiscreteContinuousConv2d, DiscreteContinuousConvTranspose2d, EquidistantDiscreteContinuousConv2d, EquidistantDiscreteContinuousConvTranspose2d
+
+# %%
+# Let's start by loading an example image
+os.system("curl https://upload.wikimedia.org/wikipedia/commons/thumb/d/d3/Albert_Einstein_Head.jpg/360px-Albert_Einstein_Head.jpg -o ./einstein.jpg")
+
+nx = 90
+ny = 120
+
+img = iio.imread('./einstein.jpg')
```
dhpitt, line 38: I don't think you need to add imageio to read in an image

neuralop/layers/discrete_continuous_convolution.py
bonevbs: addressed in the main docstring

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,619 @@
...
+    Discrete-continuous convolutions (DISCO) on arbitrary 2d grids as implemented for [1]. To evaluate continuous convolutions on a computer, they can be evaluated semi-discretely, where the translation operation is performet continuously, and the qudrature/projection is performed discretely on a grid [2]. They are the main building blocks for local Neural Operators [1]. Forward call expects an input of shape batch_size x num_channels x num_grid_points.
+
+    References
+    ----------
+    .. [1] Liu-Schiaffini M., Berner J., Bonev B., Kurth T., Azizzadenesheli K., Anandkumar A.; Neural Operators with Localized Integral and Differential Kernels;  arxiv:2402.16845
+    .. [2] Ocampo J., Price M.A. , McEwen J.D.; Scalable and equivariant spherical CNNs by discrete-continuous (DISCO) convolutions, ICLR (2023), arXiv:2209.13603
+
+    Parameters
+    ----------
+        in_channels: int
```
dhpitt, line 115: nit: remove one indent

examples/plot_DISCO_convolutions.py
bonevbs: adressed this in new commit

examples/plot_DISCO_convolutions.py
bonevbs: you are right - replaced it with matplotlib's image read function

neuralop/layers/discrete_continuous_convolution.py
bonevbs: like this ?
```
    Parameters
    ----------
    in_channels: int
```

dhpitt: Exactly

neuralop/layers/discrete_continuous_convolution.py
dhpitt: I mean, what is the context for 
```
# the instantiator....
```

This could be clearer

neuralop/layers/discrete_continuous_convolution.py
dhpitt: As per earlier conversation, I agree we should keep the long names

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,649 @@
...
+    Discrete-continuous convolutions (DISCO) on equidistant 2d grids as implemented for [1]. This implementation maps to 2d convolution kernels which makes it more efficient than the unstructured implementation above. Due to the mapping to an equidistant grid, the domain lengths need to be specified in order to compute the effective resolution and the corresponding cutoff radius. Forward call expects an input of shape batch_size x in_channels x in_shape[0] x in_shape[1].
+
+    References
+    ----------
+    .. [1] Liu-Schiaffini M., Berner J., Bonev B., Kurth T., Azizzadenesheli K., Anandkumar A.; Neural Operators with Localized Integral and Differential Kernels;  arxiv:2402.16845
+    .. [2] Ocampo J., Price M.A. , McEwen J.D.; Scalable and equivariant spherical CNNs by discrete-continuous (DISCO) convolutions, ICLR (2023), arXiv:2209.13603
+
+    Parameters
+    ----------
+        in_channels: int
```
dhpitt, line 417: same nit: indent

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,649 @@
...
+
+        return out
+
+class EquidistantDiscreteContinuousConvTranspose2d(DiscreteContinuousConv):
+    """
+    Transpose Discrete-continuous convolutions (DISCO) on equidistant 2d grids as implemented for [1]. This implementation maps to 2d convolution kernels which makes it more efficient than the unstructured implementation above. Due to the mapping to an equidistant grid, the domain lengths need to be specified in order to compute the effective resolution and the corresponding cutoff radius. Forward call expects an input of shape batch_size x in_channels x in_shape[0] x in_shape[1].
+
+    References
+    ----------
+    .. [1] Liu-Schiaffini M., Berner J., Bonev B., Kurth T., Azizzadenesheli K., Anandkumar A.; Neural Operators with Localized Integral and Differential Kernels;  arxiv:2402.16845
```
dhpitt, line 532: Nit: move references to bottom

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,649 @@
...
+    .. [2] Ocampo J., Price M.A. , McEwen J.D.; Scalable and equivariant spherical CNNs by discrete-continuous (DISCO) convolutions, ICLR (2023), arXiv:2209.13603
+
+    Parameters
+    ----------
+        in_channels: int
```
dhpitt, line 537: nit: indent

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,649 @@
...
+            psi_loc[f, j] = v
+
+        # compute local version of the filter matrix
+        psi_loc = psi_loc.reshape(self.kernel_size, self.psi_local_h, self.psi_local_w)
+
+        self.register_buffer("psi_loc", psi_loc, persistent=False)
+
+    def get_psi(self):
+        """
+        Returns Psi
```
dhpitt, line 506: What is psi? A line or two would be useful

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,649 @@
...
```
dhpitt, line 627: Same comment, adding a line to explain what psi is

dhpitt (approved):
Thanks for responding to the earlier review. I left a few last small nitpicks about docstring formatting, other than that I think it's good to go.

examples/plot_DISCO_convolutions.py
dhpitt: No change is showing

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,648 @@
...
+
+    def forward(self, x: torch.Tensor) -> torch.Tensor:
+        """
+        Forward call. Expects an input of shape batch_size x in_channels x in_shape[0] x in_shape[1].
+        """
+
+        kernel = torch.einsum("kxy,ogk->ogxy", self.get_psi(), self.weight)
+
+        # padding is rounded down to give the right result when even kernels are applied
+        # Check https://pytorch.org/docs/stable/generated/torch.nn.Conv2d.html for output shape math
```
mliuschi, line 639: Should this link be https://pytorch.org/docs/stable/generated/torch.nn.ConvTranspose2d.html instead?

mliuschi (approved):
Just a couple minor comments from me. Looks great!

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,648 @@
...
+
+    if normalize:
+        vals = _normalize_convolution_tensor_2d(idx, vals, grid_in, grid_out, kernel_shape, quad_weights, transpose_normalization=transpose_normalization)
+
+    return idx, vals
+
+
+class DiscreteContinuousConv2d(DiscreteContinuousConv):
+    """
+    Discrete-continuous convolutions (DISCO) on arbitrary 2d grids as implemented for [1]. To evaluate continuous convolutions on a computer, they can be evaluated semi-discretely, where the translation operation is performet continuously, and the qudrature/projection is performed discretely on a grid [2]. They are the main building blocks for local Neural Operators [1]. Forward call expects an input of shape batch_size x in_channels x n_in.
```
white-alistair, line 106:
```suggestion
    Discrete-continuous convolutions (DISCO) on arbitrary 2d grids as implemented for [1]. To evaluate continuous convolutions on a computer, they can be evaluated semi-discretely, where the translation operation is performed continuously, and the quadrature/projection is performed discretely on a grid [2]. They are the main building blocks for local Neural Operators [1]. Forward call expects an input of shape batch_size x in_channels x n_in.
```
Just taking a look around, sorry to nitpick :sweat_smile:

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,648 @@
...
+
+    return idx, vals
+
+
+class DiscreteContinuousConv2d(DiscreteContinuousConv):
+    """
+    Discrete-continuous convolutions (DISCO) on arbitrary 2d grids as implemented for [1]. To evaluate continuous convolutions on a computer, they can be evaluated semi-discretely, where the translation operation is performed continuously, and the quadrature/projection is performed discretely on a grid [2]. They are the main building blocks for local Neural Operators [1]. Forward call expects an input of shape batch_size x in_channels x n_in.
+
+    Parameters
+    ----------
```
JeanKossaifi, line 109: Pretty sure we need a newline here

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,648 @@
...
+    groups: int, optional
+        number of groups in the convolution
+    bias: bool, optional
+        whether to use a bias
+    radius_cutoff: float, optional
+        cutoff radius for the kernel
+
+    References
+    ----------
+    .. [1] Liu-Schiaffini M., Berner J., Bonev B., Kurth T., Azizzadenesheli K., Anandkumar A.; Neural Operators with Localized Integral and Differential Kernels;  arxiv:2402.16845
```
JeanKossaifi, line 137: I _think_ we may need new lines between the references for it to be rendered properly though not 100% sure of the top of my head.

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,648 @@
...
+    quad_weights: torch.Tensor, optional
+        quadrature weights on the input grid
+    periodic: bool, optional
+        whether the domain is periodic
+    groups: int, optional
+        number of groups in the convolution
+    bias: bool, optional
+        whether to use a bias
+    radius_cutoff: float, optional
+        cutoff radius for the kernel
```
JeanKossaifi, line 133: Could we have a little more info for this parameter, how to choose it, what it means?

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,648 @@
...
+    grid_out: torch.Tensor or str
+        output grid in the form of a point cloud. Can also pass a string to generate a regular (tensor) grid.
+    kernel_shape: Union[int, List[int]]
+        kernel shape. Expects either a signle integer for isotropic kernels or two integers for anisotropic kernels
+    n_in: Tuple[int], optional
+        number of input points
+    n_out: Tuple[int], optional
+        number of output points
+    quad_weights: torch.Tensor, optional
+        quadrature weights on the input grid
```
JeanKossaifi, line 125: We probably want to specify the expected size

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,648 @@
...
+    Discrete-continuous convolutions (DISCO) on arbitrary 2d grids as implemented for [1]. To evaluate continuous convolutions on a computer, they can be evaluated semi-discretely, where the translation operation is performed continuously, and the quadrature/projection is performed discretely on a grid [2]. They are the main building blocks for local Neural Operators [1]. Forward call expects an input of shape batch_size x in_channels x n_in.
+
+    Parameters
+    ----------
+    in_channels: int
+        input channels to DISCO convolution
+    out_channels: int
+        output channels of DISCO convolution
+    grid_in: torch.Tensor or str
+        input grid in the form of a point cloud. Can also pass a string to generate a regular (tensor) grid.
```
JeanKossaifi, line 115: Can we list the available strings and expected format of the point cloud?

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,648 @@
...
+        # to improve performance, we make psi a matrix by merging the first two dimensions
+        # This has to be accounted for in the forward pass
+        idx = torch.stack([idx[0] * self.n_out + idx[1], idx[2]], dim=0)
+
+        self.register_buffer("psi_idx", idx.contiguous(), persistent=False)
+        self.register_buffer("psi_vals", vals.contiguous(), persistent=False)
+
+    def get_psi(self):
+        """
+        Returns Psi
```
JeanKossaifi, line 218: I could have guessed from the name :) 
Can we say instead what Psi represents?

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -0,0 +1,648 @@
...
+        super().__init__(in_channels, out_channels, kernel_shape, groups, bias)
+
+        # the instantiator supports convenience constructors for the input and output grids
+        if isinstance(grid_in, torch.Tensor):
+            assert isinstance(quad_weights, torch.Tensor)
+            assert not periodic
+        elif isinstance(grid_in, str):
+            assert n_in is not None
+            assert len(n_in) == 2
+            x, wx = _precompute_grid(n_in[0], grid=grid_in, periodic=periodic)
```
JeanKossaifi, line 165: given that _precompute_grid is in another lib, can we name x, wx with self explanatory names and/or add a short comment explaining they're the quadrature weights?

dhpitt:
Superseded by #467, thanks @bonevbs !

PR #458, opened 2024-10-24
Simplify the GINO docstring (round 1)

dhpitt:
Breaks `GINO` and `FNOGNO` docstrings into the Params and Other Params schemes. Does the same for GNOBlock. More updates to come after this release, when we implement a custom numpy-like docstring module for more param sections


JeanKossaifi (approved):
Looks good to me.

PR #462, opened 2024-10-29
CODABlock with argument reorders

dhpitt:
an update of #436 with more argument organization, sensible defaults for FNOBlocks, a few var renames to put in line with the rest of the package, and updates to the tests to remove stale config options


neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,417 @@
...
+        attention = attention.view(
+            attention.size(0),
+            attention.size(1),
+            attention.size(2),
+            d,
+            *tokens.shape[-self.n_dim:])
+        attention = torch.transpose(attention, 1, 2)
+        attention = attention.reshape(attention.size(0) * attention.size(1),
+                                   attention.size(2) * attention.size(3),
+                                   *attention.shape[-self.n_dim:])
```
JeanKossaifi, line 321: I think I'd commented on this before: I find it strange to do a view followed by a transpose and a reshape: why not just one transpose followed by one reshape here?

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,417 @@
...
+            batch_size,
+            t * output.size(1),
+            *output.shape[-self.n_dim:])
+
+        return output
+
+    def _forward_non_equivariant(self, x):
+        """uses non-permuatation equivariant mixer
+        layer and normalizations.
+        Parameters
```
JeanKossaifi, line 377: There should be new line before and after the section title I believe for it to render properly. Would be good to check the doc when we add these. 

We should add this to the API doc as well.

neuralop/layers/coda_blocks.py
```diff
@@ -0,0 +1,417 @@
...
+
+    def forward(self, x):
+        if self.permutation_eq:
+            return self._forward_equivariant(x)
+        else:
+            return self._forward_non_equivariant(x)
+
+    def _forward_equivariant(self, x):
+        """Use permutation equivariant mixer layer after the attention mechanism. 
+        Shares the same mixer layer for all the variables.
```
JeanKossaifi, line 333: Can we have more details here on what exactly it means and implementation of the equivariant and non equivariant version? That will be quite unintuitive for users.

JeanKossaifi (commented):
Looks good, just some small comments and I think it's ready to merge.

neuralop/layers/coda_blocks.py
dhpitt: These won't be documented as they are private members. The main forward pass will be documented and we should add info on both forms of the forward pass (similar to preactivation vs. postactivation forward calls in FNOBlock)

PR #464, opened 2024-10-31
Diffs to fix transpose conv

dhpitt:


neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -570,6 +736,11 @@ def __init__(
         **kwargs
     ):
         super().__init__(in_channels, out_channels, kernel_shape, groups, bias)
+        # torch ConvTranspose2d expects grouped weights stacked along the out_channels
+        # shape (in_channels, out_channels/groups, h, w)
+        self.weight = nn.Parameter(self.weight.permute(1,0,2).reshape(self.groupsize * self.groups,
```
bonevbs, line 741: Is this the only change necessary to get it to work?

bonevbs (commented):
Looks good to me - if this simple one-liner made it to work, I am quite happy to merge it.

neuralop/layers/discrete_continuous_convolution.py
dhpitt: Yes, and the einsum in the transpose conv

PR #465, opened 2024-11-01
Adding DISCO into Local FNO blocks

mliuschi:
Updates to Local FNO blocks to incorporate DISCO and updates to the local FNO blocks tests.


mliuschi:
See `neuralop/layers/local_fno_block.py` and `neuralop/layers/tests/test_local_fno_block.py`.

neuralop/layers/local_fno_block.py
```diff
@@ -19,107 +21,136 @@
...
+        See differential_conv.py.
+    conv_padding_mode : str in ['periodic', 'circular', 'replicate', 'reflect', 'zeros'], optional
+        Padding mode for spatial convolution kernels.
+    default_in_shape : Tuple[int]
+        Default input shape on spatiotemporal dimensions. Repeated above for differential kernel.
+    fin_diff_kernel_size : odd int, optional
+        Conv kernel size for finite difference convolution.
+    mix_derivatives : bool, optional
+        Whether to mix derivatives across channels
+    EquidistantDiscreteContinuousConv2d Params
```
dhpitt, line 112: This section won't render properly. I suggest moving to `Parameters` and `Other Parameters`

dhpitt (commented):
These changes look good, thanks @mliuschi! Let's merge once we merge DISCO.

neuralop/layers/local_fno_block.py
mliuschi: Thanks! I updated it to follow the convention in `FNOBlocks`

neuralop/layers/tests/test_local_fno_block.py
```diff
@@ -14,7 +14,7 @@ def test_LocalFNOBlock_resolution_scaling_factor():
     mlp_skip='linear'
     for dim in [1, 2, 3]:
```
dhpitt, line 15: Instead of looping through here, let's add an 

`@pytest.mark.parametrize` call

dhpitt (commented):
Thanks for the changes @mliuschi ! Almost ready to go, just a small style suggestion for the tests

neuralop/layers/tests/test_local_fno_block.py
dhpitt: and you don't need call the test functions at the bottom, Pytest collects and runs them automatically

mliuschi: Thanks @dhpitt! Yeah, sorry I added the test functions at the bottom just to be able to quickly test myself but forgot to remove them in my initial commit.

mliuschi:
Thanks for the feedback @dhpitt and for resolving some of the minor issues! Should we move forward with merging or are we waiting on anyone else to review?

dhpitt:
I think we can go ahead and merge

JeanKossaifi:
@dhpitt let's please make sure the doc builds and is properly rendered before merging.
@mliuschi could we potentially add a correctness test - e.g. there a simplistic case where we could init the conv with a known kernel and verify it matches a regular convolution?

neuralop/layers/tests/test_local_fno_block.py
```diff
@@ -12,60 +13,59 @@ def test_LocalFNOBlock_resolution_scaling_factor():
...
     if norm == 'ada_in':
         embedding = torch.randn(ada_in_features)
         block.set_ada_in_embeddings(embedding)
 
-    x = torch.randn(2, 3, *size[:dim])
+    x = torch.randn(2, 3, *size[:n_dim])
     res = block(x)
-    assert(list(res.shape[2:]) == size[:dim])
\ No newline at end of file
+    assert(list(res.shape[2:]) == size[:n_dim])
```
JeanKossaifi, line 71: @mliuschi could we potentially add a correctness test - e.g. there a simplistic case where we could init the conv with a known kernel and verify it matches a regular convolution?

neuralop/layers/local_fno_block.py
```diff
@@ -17,108 +19,137 @@
...
+        fixed_rank_modes parameter for SpectralConv, by default False
+    implementation : str, optional
+        implementation parameter for SpectralConv, by default "factorized"
+    decomposition_kwargs : _type_, optional
+        kwargs for tensor decomposition in SpectralConv, by default dict()
+    
+    References
+    -----------
+    .. [1] :
+
```
JeanKossaifi, line 120: The references here seem broken - I don't think this will get rendered properly

JeanKossaifi (commented):
Looks good to me, but let's make sure that the documentation builds and is rendered properly before merging any PR @dhpitt

neuralop/layers/tests/test_local_fno_block.py
mliuschi: Hmm, that seems like something that might be better suited for testing the DISCO layer itself, right? This is just for merging a working DISCO layer with FNO and the differential kernels. What do you think?

neuralop/layers/local_fno_block.py
mliuschi: This format for the reference I copied from `FNOBlocks`. @dhpitt Do you know if these would render correctly?

JeanKossaifi: There shouldn't be all the newlines: see e.g. https://github.com/tensorly/tensorly/blob/55a022e6558b281d63b2d342daf2a1cc2f2d8281/tensorly/decomposition/_cp.py#L316 for an example

JeanKossaifi:
The doc is also currently broken : there should be no `.py` in import (in `import neuralop.layers.local_fno_block.py`).

neuralop/layers/tests/test_local_fno_block.py
JeanKossaifi: Yes, I agree, I didn't get time to review before it got merged. Would you be able to make another small PR to add this @mliuschi?

neuralop/layers/local_fno_block.py
JeanKossaifi: The references in FNOBlocks have the same issue.

neuralop/layers/tests/test_local_fno_block.py
mliuschi: This might pose a slight issue because of the radial basis that we have defined: it is not straightforward to directly parameterize a known kernel under our basis. 

I think Boris has done something similar in his plotting example though: he defined something similar to a traditional edge detector filter and applied it over a sample image. Would something like that address this?

neuralop/layers/local_fno_block.py
mliuschi: Ah, I see. I'll make a commit and fix both

mliuschi: Just pushed!

neuralop/layers/tests/test_local_fno_block.py
JeanKossaifi: Yes exactly, just some correctness check - doesn't have to be the full general case but better than just checking the shape.

JeanKossaifi:
Closing as superseded by #468

PR #467, opened 2024-11-01
Add DISCO conv

dhpitt:
An update of @bonevbs 's PR that merges with main and adds more documentation


bonevbs:
LGTM. feel free to merge

PR #473, opened 2024-11-14
Update package to PEP621 standards

dhpitt:
* replaces `setup.py` entirely with a `pyproject.toml`
* ensures all required dependencies are installed when `pip install` is run
* adds `[doc], [dev], [all]` optional dependency extensions
* changes build backend from setuptools to PyPA's `build`


JeanKossaifi:
Thanks @dhpitt - looks great!

I have one concern: having duplicated information for description and version. Previously I was reading those from the package directly in setup.py but this isn't possible in the toml.

JeanKossaifi:
Great, thank you!

PR #484, opened 2024-12-11
Only warn about open3d when *creating* `MeshDataModule`

dhpitt:
Currently, if `MeshDataModule` is imported without `open3d` built, the package prints a warning message. Since all data modules are imported in `neuralop.data.__init__`, this means a warning message is printed every time `neuralop` is imported. 

This PR changes to only print the warning when `MeshDataModule` is initialized without `open3d`.


JeanKossaifi:
Merging - I think we should make it into a lazy import in a future PR.

PR #486, opened 2024-12-13
Simpler loss reduce interface

dhpitt:


JeanKossaifi (commented):
Great, thanks @dhpitt, this is much better - I wonder if perhaps we shouldn't also just rename `L` to just `quadrature`: while d and p are somewhat clear in this context, L isn't

JeanKossaifi:
Actually - we can rename L to measure and h to quadrature.

JeanKossaifi:
Thank you, merging!

Issue #480, opened 2024-11-26
Checkpointing and loading training states


Dzz99:
I encountered an issue while running the "Checkpointing and loading training states" example from the user manual. After using resume_from_dir, the model is not training properly.

Here is the code I used:

device = "cuda" if torch.cuda.is_available() else "cpu"
# device ="cpu"
train_loader, test_loaders, data_processor = load_darcy_flow_small(
        n_train=1000, batch_size=32,
        test_resolutions=[16, 32], n_tests=[100, 50],
        test_batch_sizes=[32, 32],
)

model = FNO(n_modes=(4,4),
             in_channels=1,
             out_channels=1,
             hidden_channels=4,
             projection_channel_ratio=1,
             factorization='tucker',
             rank=0.42)

model = model.to(device)
optimizer = AdamW(model.parameters(),
                  lr=8e-3,
                  weight_decay=1e-4)
scheduler = torch.optim.lr_scheduler.CosineAnnealingLR(optimizer, eta_min=8e-3, T_max=30)
l2loss = LpLoss(d=2, p=2)
h1loss = H1Loss(d=2)

train_loss = h1loss
eval_losses = {'h1': h1loss, 'l2': l2loss}

trainer = Trainer(model=model, n_epochs=10,
                  device=device,
                  data_processor=data_processor,
                  wandb_log=False,
                  eval_interval=1,
                  use_distributed=False,
                  verbose=True)
trainer.train(train_loader=train_loader,
              test_loaders={},
              optimizer=optimizer,
              scheduler=scheduler,
              regularizer=False,
              training_loss=train_loss,
              save_every=7,
              save_dir="./checkpoints")

# .. resume_from_dir:
trainer = Trainer(model=model, n_epochs=10,
                  device=device,
                  data_processor=data_processor,
                  wandb_log=False,
                  eval_interval=1,
                  use_distributed=False,
                  verbose=True)
trainer.train(train_loader=train_loader,
              test_loaders={},
              optimizer=optimizer,
              scheduler=scheduler,
              regularizer=False,
              training_loss=train_loss,
              resume_from_dir="./checkpoints")
The code for saving the model state works fine, but after calling resume_from_dir, the model parameters do not update during training. Based on my testing, it seems that the issue lies in the train_one_epoch function in trainer.py, where the command self.optimizer.step() does not seem to have any effect.

dhpitt:
[...]

PR #491, opened 2024-12-17
(fix) Remove `from_checkpoint` in `load_training_state` to avoid overwriting model

dhpitt:
Thanks to authors of #480 and @JeanKossaifi for catching this.


JeanKossaifi (commented):
Thanks @dhpitt  - let's merge this as a temporary fix and figure out better loading down the line.

PR #493, opened 2024-12-18
Save metadata in `BaseModel.state_dict()`, check version tag in `BaseModel.load_state_dict()`

dhpitt:


neuralop/models/base_model.py
```diff
@@ -76,7 +76,33 @@ def __new__(cls, *args, **kwargs):
...
-    
+
+    def state_dict(self, **kwargs):
+        """
+        state_dict subclasses nn.Module.state_dict() and adds a metadata field
+        to track the model version and ensure only compatible saves are loaded.
+        """
+        print("In state dict")
+        state_dict = super().state_dict(**kwargs)
+        state_dict.update({'neuralop_metadata': self._init_kwargs})
```
JeanKossaifi, line 87: Shall we just call it `_metadata`? Perhaps we add a test to check that there isn't already a key under that name in the unlikely event that this happened.
I would also just do `state_dict['_metadata'] = self._init_kwargs` since we are only updating the values for one key.

neuralop/models/base_model.py
```diff
@@ -76,7 +76,33 @@ def __new__(cls, *args, **kwargs):
...
+        return state_dict
+
+    def load_state_dict(self, state_dict, **kwargs):
+        """
+        state_dict subclasses nn.Module.load_state_dict() and adds a metadata field
+        to track the model version and ensure only compatible saves are loaded.
+        """
+        metadata = state_dict.get('neuralop_metadata')
```
JeanKossaifi, line 95: Is this tested? I believe get returns None by default so the `if metadata` below would fail.
Let's make the default `None` explicit and test `if metadata is not None` below to remove the ambiguity.

Let's also use pop rather than get so we don't need to remove the metadata before passing to super().

neuralop/models/base_model.py
```diff
@@ -76,7 +76,33 @@ def __new__(cls, *args, **kwargs):
...
+        if metadata:
+            saved_version = metadata['_version']
+            self_version = self._init_kwargs['_version']
```
JeanKossaifi, line 98: Let's use get('_version', None) and add another warning if the model doesn't have a version attached.

neuralop/models/base_model.py
```diff
@@ -76,7 +76,33 @@ def __new__(cls, *args, **kwargs):
...
+        """
+        state_dict subclasses nn.Module.state_dict() and adds a metadata field
+        to track the model version and ensure only compatible saves are loaded.
+        """
+        print("In state dict")
+        state_dict = super().state_dict(**kwargs)
+        state_dict.update({'neuralop_metadata': self._init_kwargs})
+        return state_dict
+
+    def load_state_dict(self, state_dict, **kwargs):
```
JeanKossaifi, line 90: Should we match the default signature? 
https://pytorch.org/docs/stable/generated/torch.nn.Module.html

No strong preference here though since this obviates the need to update to match pytorch.

JeanKossaifi (commented):
Thanks @dhpitt - left a few comments in the code.

neuralop/models/base_model.py
dhpitt: We can just call it `'_metadata'`, but we need to update because the PyTorch state dict doesn't support item assignment (throws an error)

neuralop/models/base_model.py
dhpitt: Sounds good.

dhpitt: Note that `pop` will throw a keyerror though.

neuralop/models/base_model.py
```diff
@@ -76,7 +76,38 @@ def __new__(cls, *args, **kwargs):
...
+        state_dict subclasses nn.Module.load_state_dict() and adds a metadata field
+        to track the model version and ensure only compatible saves are loaded.
+        """
+        metadata = state_dict.pop('_metadata', None)
+        if metadata is not None:
+
+            saved_version = metadata.get('_version', None)
+            if saved_version is None:
+                warnings.warn(f"Saved instance of {self.__class__} has no stored version attribute.")
+            self_version = self._init_kwargs['_version']
```
JeanKossaifi, line 104: This should just be the class attribute, so just self._version, it shouldn't be passed at init.

neuralop/models/base_model.py
```diff
@@ -76,7 +76,38 @@ def __new__(cls, *args, **kwargs):
...
+            warnings.warn("Attempting to update metadata for a module with metadata already in self.state_dict()")
+        return state_dict
+
+    def load_state_dict(self, state_dict, **kwargs):
+        """
+        state_dict subclasses nn.Module.load_state_dict() and adds a metadata field
+        to track the model version and ensure only compatible saves are loaded.
+        """
+        metadata = state_dict.pop('_metadata', None)
+        if metadata is not None:
```
JeanKossaifi, line 99: Nitpick: the newlines are odd - i'd have a new line before the condition, not just after the if. Also not consistent with the next one below.

neuralop/models/base_model.py
```diff
@@ -76,7 +76,38 @@ def __new__(cls, *args, **kwargs):
...
+
+            saved_version = metadata.get('_version', None)
+            if saved_version is None:
+                warnings.warn(f"Saved instance of {self.__class__} has no stored version attribute.")
+            self_version = self._init_kwargs['_version']
+            if saved_version != self_version:
```
JeanKossaifi, line 105:
```suggestion
            if saved_version != self._version:
```
and let's omit the line defining self_version above

JeanKossaifi (commented):
Just a few fixes - the version should be a class attribute of each model

neuralop/models/base_model.py
```diff
@@ -76,7 +76,37 @@ def __new__(cls, *args, **kwargs):
...
 
         return instance
-    
+
+    def state_dict(self, **kwargs):
+        """
+        state_dict subclasses nn.Module.state_dict() and adds a metadata field
+        to track the model version and ensure only compatible saves are loaded.
+        """
+        print("In state dict")
```
JeanKossaifi, line 85: Let's remove the print.

JeanKossaifi (commented):
I realized we can actually use PyTorch's load_state_dict option `assign=False`, instead of loading model and optimizer separately: 
https://pytorch.org/docs/stable/generated/torch.nn.Module.html

JeanKossaifi:
Looks good, thank you @dhpitt

PR #497, opened 2024-12-20
Codano model

ashiq24:
1. Fixed a minor issue in codano_block. Also renamed one parameter to match the other layers.
2. Introduced the CODANO model.


neuralop/models/codano.py
```diff
@@ -0,0 +1,371 @@
...
+from ..layers.padding import DomainPadding
+from ..layers.coda_blocks import CODABlocks
+from ..layers.resample import resample
+from ..layers.embeddings import GridEmbedding2D, GridEmbeddingND
+
+class CODANO(nn.Module):
+    """
+    Parameters
+    ---
+    output_variable_codimension : int
```
dhpitt, line 17: If you provide default values for arguments, they should be noted in the docstring

neuralop/models/codano.py
```diff
@@ -0,0 +1,371 @@
...
+        number of hidden channels in lifting/projection module for each variable
+    use_positional_encoding : bool
+        whether to use varibale specific positional encoding
+    positional_encoding_dim : int
+        dimension (number of channels) of positional encoding
+    static_channel_dim : int
+        number of channels in static channel. These channels contains additional static information of the PDE such as boundary conditions.
+    positional_encoding_modes : list
+        number of Fourier modes used in positional encoding
+    n_variables : int
```
dhpitt, line 31: more clarity on what this means would be useful

neuralop/models/codano.py
```diff
@@ -0,0 +1,371 @@
...
+        additional arguments for the CODABlocks
+    projection : bool
+        whether to use projection module
+    lifting : bool
+        whether to use lifting module
+    domain_padding : float
+        padding factor for domain padding
+    domain_padding_mode : str
+        domain padding mode (one-sided or two-sided)
+    enable_cls_token : bool
```
dhpitt, line 72: Could you reorder these args by importance?

neuralop/models/codano.py
```diff
@@ -0,0 +1,371 @@
...
+from ..layers.spectral_convolution import SpectralConv
+from ..layers.skip_connections import skip_connection
+from ..layers.padding import DomainPadding
+from ..layers.coda_blocks import CODABlocks
+from ..layers.resample import resample
+from ..layers.embeddings import GridEmbedding2D, GridEmbeddingND
+
+class CODANO(nn.Module):
+    """
+    Parameters
```
dhpitt, line 15: A small blurb here explaining the basics of the CODANO would be useful.

neuralop/models/codano.py
```diff
@@ -0,0 +1,371 @@
...
+    def forward(self, x: torch.Tensor, static_channel=None):
+        '''
+        Parameters
+        ----------
+        x : torch.Tensor
+            input tensor of shape (batch_size, channel, H, W, ...)
+        static_channel : torch.Tensor
+            static channel tensor of shape (batch_size, static_channel_dim, H, W, ...)
+        '''
+        batch, n_variables, *spatial_shape = x.shape
```
dhpitt, line 299: is `n_variables` allowed to be `None`? Some explanation of what is expected to happen in that case would be good

neuralop/models/codano.py
```diff
@@ -0,0 +1,371 @@
...
+                    cls_token,
+                    x,
+                ],
+                dim=1,
+            )
+
+        if self.domain_padding is not None:
+            x = self.domain_padding.pad(x)
+
+        output_shape = [
```
dhpitt, line 328: Comments here explaining the different operations of the forward pass might be helpful too.

neuralop/models/codano.py
dhpitt, line 72: And could you describe what `CLS` is?

neuralop/models/tests/test_codano.py
```diff
@@ -0,0 +1,67 @@
...
+    if static_channel_dim > 0:
+        static_channels = torch.randn(2, static_channel_dim, 32, 32)
+    else:
+        static_channels = None
+
+    # Test forward pass
+    out = model(in_data, static_channels)
+
+    # Check output size
+    assert list(out.shape) == [2, n_variables, 32, 32]
```
dhpitt, line 59: Are there any correctness tests you could imagine that test something about the model outputs more complex than its shape?

dhpitt (commented):
Thanks for the PR @ashiq24 ! I left a few comments, mostly to do with documentation. This model is quite complex and comments are very important to help understand the model's function.

neuralop/models/codano.py
ashiq24: added

neuralop/models/codano.py
ashiq24: Added further clarifications.

neuralop/models/codano.py
ashiq24: Clarified that CLS is the CLASS token.

neuralop/models/codano.py
ashiq24: Added.

neuralop/models/codano.py
ashiq24: I renamed the variable and made it different from the 'self.n_variables'.

neuralop/models/codano.py
ashiq24: Added

neuralop/models/codano.py
ashiq24: Clarified.

ashiq24:
@dhpitt 
I've added the comments. Please check if any further modifications are needed.

neuralop/models/codano.py
```diff
@@ -0,0 +1,455 @@
...
+from ..layers.skip_connections import skip_connection
+from ..layers.padding import DomainPadding
+from ..layers.coda_blocks import CODABlocks
+from ..layers.resample import resample
+from ..layers.embeddings import GridEmbedding2D, GridEmbeddingND
+
+class CODANO(nn.Module):
+
+    """ Codomain Attention Neural Operators (CoDA-NO) uses a specialized attention mechanism in the codomain space for data in 
+    infinite dimensional spaces as described in [1]_. The model treates each input channel as a variable of the pyscial system
```
dhpitt, line 16: nit: `physical system` (typo)

neuralop/models/codano.py
```diff
@@ -0,0 +1,455 @@
...
+    encoding and static channel information for additional context of the physical system such as external force or inlet condition.
+
+
+    Parameters
+    ----------
+    output_variable_codimension : int
+        The codomain dimension or the number of output channels corresponding to each input variable (or channel). Default is 1.
+        Example: For a input with 3 variables (channels) and output_variable_codimension=2, the output will have 6 channels (3 variables × 2 codimension).
+
+    lifting : bool
```
dhpitt, line 28: `lifting_variable_codimension` and `hidden_variable_codimension` could be clarified, or the behavior should be explained in the case where `lifting=False`. I suggest changing the argument name to `lifting_codimension` or `lifting_channels` and allowing `None` instead of a separate bool option

neuralop/models/codano.py
dhpitt: I think it could even be clearer - what are the variables? How is positional encoding performed?

neuralop/models/codano.py
```diff
@@ -0,0 +1,455 @@
...
+        Example: For a 5-layer architecture, horizontal_skips_map={4: 0, 3: 1} creates skip connections from layer 0 to layer 4 and layer 1 to layer 3.
+
+    n_layers : int
+        The number of codomain attention layers. Default is 4.
+
+    n_modes : list
+        The number of Fourier modes to use in integral operations in the CoDA-NO block along each dimension. Default is None.
+        Example: For a 5-layer 2D CoDA-NO, n_modes=[[16, 16], [16, 16], [16, 16], [16, 16], [16, 16]].
+
+    scalings : list
```
dhpitt, line 73: Could this be `scaling_factors`? How is this different than the `output_shape` parameters of the FNO? Behavior should be clearer

neuralop/models/codano.py
```diff
@@ -0,0 +1,455 @@
...
+        The output scaling factor for each layer along each dimension. Default is None.
+        Example: For a 2D input, scalings=[[1, 1], [0.5, 0.5], [1, 1], [2, 2], [1, 1]].
+
+    n_heads : list
+        The number of attention heads for each layer. Default is None.
+        Example: For a 4-layer CoDA-NO, n_heads=[2, 2, 2, 2].
+
+    attention_scalings : list
```
dhpitt, line 81: same comment as above

neuralop/models/codano.py
dhpitt: Thanks

dhpitt: I don't know if this is clear enough. What exactly is a learnable class token?

neuralop/models/codano.py
```diff
@@ -0,0 +1,455 @@
...
+        return x
+
+    def forward(self, x: torch.Tensor, static_channel=None):
+        '''
+        Parameters
+        ----------
+        x : torch.Tensor
+            input tensor of shape (batch_size, num_inp_var, H, W, ...)
+        static_channel : torch.Tensor
+            static channel tensor of shape (batch_size, static_channel_dim, H, W, ...)
```
dhpitt, line 360: What is the static channel tensor/its use case?

neuralop/models/codano.py
dhpitt: And is `n_variables` analogous to `in_channels`? If so this should be a higher-ordered arg, and its meaning to the model should be explained.

e.g., `number of distinct physical variables in data, encoded as channels`

neuralop/models/codano.py
ashiq24: This term [class] token or [class] embedding has already been introduced in ViT and another model (Figure 1 [1]). Should I provide a link to such work? 


[1]. https://arxiv.org/pdf/2010.11929

neuralop/models/codano.py
ashiq24: Renamed those to `lifting_channel_ratio` following the fno model.

neuralop/models/codano.py
ashiq24: Added. This is explained in detail in the doc while describing `static_channel_dim`.

neuralop/models/codano.py
ashiq24: I think explicit parameterization with `lifting=False/True` is much more precise than indirect parametrization through setting `lifting_channel_ratio=None.`

neuralop/models/codano.py
ashiq24: Rename as "per_layer_scaling_factor." Its functionality is similar to uno_scalings. I also added an example.

ashiq24: For `positional encoding,` I added a reference to a standard transformer model, such as BERT, as we are using a similar technique.

neuralop/layers/coda_blocks.py
```diff
@@ -253,6 +255,7 @@ def __init__(
             self.norm2 = norm_module(self.mixer_token_codimension)
```
Robertboy18, line 255: maybe we could name this mixer norm?

neuralop/models/codano.py
```diff
@@ -0,0 +1,587 @@
...
+        each unique physical varibles in the dataset.
+
+        For example:
+        If the dataset consists of only Navier Stokes equations, the variable_ids=['u_x', 'u_y', 'p'], representing the velocity
+        components in x and y directions and pressure, respectively. Please note that we consider each input channel as a physical
+        variable of the PDE.
+
+        If the dataset consists of multiple PDEs, such as Navier Stokes and Heat equation, the variable_ids=['u_x', 'u_y', 'p', 'T'],
+        where 'T' represents the temperature variable for thee Heat equation and 'u_x', 'u_y', 'p' are the velocity components and pressure
+        for the Navier Stokes equations. This is required when we aim to learn a single solver for multiple different PDEs.
```
JeanKossaifi, line 86: This is a fairly complex one - we should guide the use for each use-case, e.g. if one variables has a codim > 1.

neuralop/models/codano.py
```diff
@@ -0,0 +1,587 @@
...
+
+        This parameter is not required when `use_positional_encoding=False`.
+
+    use_horizontal_skip_connection : bool
+        Indicates whether to use horizontal skip connections, similar to U-shaped architectures. Default is False.
+
+    horizontal_skips_map : dict
+        A mapping that specifies horizontal skip connections between layers. Only required when `use_horizontal_skip_connection=True`. Default is None.
+        Example: For a 5-layer architecture, horizontal_skips_map={4: 0, 3: 1} creates skip connections from layer 0 to layer 4 and layer 1 to layer 3.
```
JeanKossaifi, line 95: Do you think we need it? Perhaps we put it in optional arg list @dhpitt

neuralop/models/codano.py
JeanKossaifi: I think more an explanation of what it is used for and how. It isn't super clear for anyone not intimately familiar with vision transformers.

neuralop/models/codano.py
```diff
@@ -0,0 +1,587 @@
...
+    def _get_positional_encoding(self, x, input_variable_ids):
+        encoding_list = []
+        for i in input_variable_ids:
+            encoding_list.append(
+                torch.fft.irfftn(self.positional_encoding[i], s=x.shape[-self.n_dim :])
+            )
+
+        return torch.stack(encoding_list, dim=1)
+
+    def _get_cls_token(self, x):
```
JeanKossaifi, line 406: This one could use some docstring perhaps

JeanKossaifi (commented):
Thanks @ashiq24 for adding the CODANO, and @dhpitt for the reviews.

Looks good overall, I think we just need to make the code as straightforward as possible: for instance, the various arguments: we have lifting (bool), hidden_variable_codimension that is ignored if lifting is False, and lifting_channels? 
Let's only have one argument `lifting_channels` that is None (no lifting) and int>0 otherwise.

@ashiq24 I would also try to keep it as simple as possible and not give too many options - we can focus on the exact architecture we used in the paper.

neuralop/layers/coda_blocks.py
```diff
@@ -190,19 +192,19 @@ def __init__(
             n_layers=1,
         )
         self.Key = FNOBlocks(
```
dhpitt, line 194: I think the `common_args` might be refined a little bit, maybe renamed to disambiguate. They're architecture args shared between all the blocks?

Also just note that `apply_skip` is not a valid argument and should be fixed or removed.

neuralop/models/codano.py
```diff
@@ -0,0 +1,619 @@
...
+
+        This parameter is **only** required when `use_positional_encoding=True` to initialize learnable positional embeddings for
+        each unique physical varibles in the dataset.
+
+        For example:
+        If the dataset consists of only Navier Stokes equations, the variable_ids=['u_x', 'u_y', 'p'], representing the velocity
+        components in x and y directions and pressure, respectively. Please note that we consider each input channel as a physical
+        variable of the PDE.
+
+        Please note that the 'velocity' varibale is composed of two channels (codimension=2) and we have splitted the velocity field
```
dhpitt, line 75: nit: typo (varibale --> variable)

neuralop/models/codano.py
```diff
@@ -0,0 +1,619 @@
...
+        learnable [class] token used in vision transformers such as ViT, which is learned directly in Fourier space. Default is False.
+
+        The [class] function is realized on the input grid by performing an inverse Fourier transform of the learned Fourier coefficients.
+        Then, the [class] token function is added to the set of input token functions before passing to the codomain attention layer. It aggregates
+        information from all the other tokens through the attention mechanism. The output token corresponding to the [class] token is discarded in the
+        output of the last CoDA block.
+
+    Other parameters
+    ------------------
+
```
dhpitt, line 139: nit: remove line

dhpitt (commented):
Overall, this looks really good. Thanks for your changes @ashiq24 . 
The forward passes in your model and layer and clear and well-documented, which is fantastic. 

As a general note, I'm wondering about the clarity of calling one CODA multi-head attention layer a `CODABlock[s]` layer, given that the `FNOBlock` actually provides the option to do multiple layers of convolutions + skips. I also left a few minor comments on spelling and formatting. But overall this PR is in good shape.

dhpitt (approved):
These are great changes, thank you @ashiq24 ! Approving.

JeanKossaifi (commented):
Looks good, thank you @ashiq24 for the changes and @dhpitt for the reviews!

PR #502, opened 2024-12-23
[UDNO] U-shaped DISCO Neural Operator

armeetj:
### `models/udno.py`
- **Implemented the “U-shaped DISCO Neural Operator” (UDNO):**  
  An in-place neural operator replacement for the U-Net. The UDNO is tailored for tasks requiring local biases and benefits from the bottleneck encoder/decoder architecture typical of U-shaped designs. Applications include imaging and medical reconstruction: https://arxiv.org/abs/2410.16290.  

  - The U-Net is central to many fields and models. Providing an in-place implementation aids in the adoption of the `neuraloperator` library in existing projects.  
  - Followed Meta’s U-Net implementation [here](https://github.com/facebookresearch/fastMRI/blob/main/fastmri/models/unet.py) using **InstanceNorm**, **LeakyReLU**, and **Dropout2d**.  
  - Set reasonable DISCO convolution defaults (similar to a conventional 3x3-based U-Net) for ease of use.  

### `models/tests/test_udno.py`
- 75 new test cases to test UDNO
- tests forward pass, backward pass, and output shapes against reference UNet implementation

### Documentation
- Added `UDNO` class to documentation `docs/sources/modules/api.rst` using automated class template. 
- Compiled and verified that html generates.

## Future Steps
- add tutorial to gallery (MNIST-esque or medical imaging)


armeetj:
Moved to https://github.com/neuraloperator/mri

Issue #321, opened 2024-05-09
Using `pip install neuraloperator` will not install all the requirements automatically. 


Theodore-wgh:
the `wandb` and `torch_harmonic` packages need to be  extra installed.

dhpitt:
combining this issue with #317. Dependencies need an update

Tsir1225:
I did encounter this, too. Here is how I solved it:
1. check out the installation guide for neural operators by the following link: [Installing NeuralOperator](https://neuraloperator.github.io/dev/install.html#pre-requisites). You should be able to notice the prerequisites include Python 3, NumPy, Scipy, PyTorch, TensorLy, and TensorLy-Torch.
2. I hope you know how to use Anaconda, a great Python project manager, so I don't need to tell you how to install Python 3 and create a virtual environment.
3. Using pip install numpy scipy (conda might also be ok but I did not try)
4. Using .whl file to install pytorch (pip or conda might also be ok but I did not try)
5. Using pip install Tensorly Tensorly-Torch (conda might also be ok but I did not try)
6. pip install neuraloperator
7. Here is the key! If you check the site-packages of your virtual env, you will find the folder of neuralop is almost empty. So you can download the folder neuralop from github and replace it by hand.
8. Should work by now.

dhpitt:
Fixed with updates for 1.0.2

PR #505, opened 2024-12-28
Fix pyproject.toml to enable pip install

takuya4880:
Thank you for mentaining a nice library.

When we `pip install neuraloperator`, currently we cannot install the main body of the library and we just get:
```
root@971493d5a2e0:/# ls /usr/local/lib/python3.13/site-packages/neuralop
__init__.py  __pycache__  utils.py
```

This is caused by the `tool.setuptoold.packages.find` part of `pyproject.toml` setting. This PR fix it. We just need `*` at the end.
cf. https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#setuptools-specific-configuration

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ requires = ["setuptools>=64"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools.packages.find]
-include = ["neuralop"]
+include = ["neuralop*"]
 
 # Dynamic versioning from neuralop.__init__
 [tool.setuptools.dynamic]


takuya4880:
Also reported at item 7 of https://github.com/neuraloperator/neuraloperator/issues/321#issuecomment-2560508945

JeanKossaifi:
Thanks a lot for the PR @takuya4880, it's much appreciated! 
I'm seeing this as I've just merged #506. Is there a preference between neuralop* and neuralop.*?

takuya4880:
Thanks for merging #506. That's almost the same as this PR.
>Is there a preference between neuralop* and neuralop.*

In this case, both works and fine for me :+1:

Issue #500, opened 2024-12-22
channel_dim in pt_dataset


jrhip:
Fairly sure this is user error, but there isn't a 3D example to check against. 

I have 2 channel 3D data in shape `4,2,64,64,64` (with a batch size of 4, for example)

In pt_dataset, channel_dim=1 inserts a dimension of size 1 at 1, yielding

`4,1,2,64,64,64`

When this is passed into the network (UNO), I get an error in conv.py (Torch):

`return F.conv1d(`

`RuntimeError: Given groups=1, weight of size [256, 2, 1], expected input[4, 1, 524288] to have 2 channels, but got 1 channels instead`

524288 is 64x64x64x2. 

What is the correct shape for multi channel 3D data? What is channel_dim for/why does it insert a dimension?

dhpitt:
Thanks for raising this issue @James-F1dev . PTDataset is for our example 2d datasets saved as `.pt` files. It should definitely be fixed to accommodate 3d data, and i'll make a fix shortly.

dhpitt:
Fixed by #509

PR #509, opened 2025-01-03
(fix) Add argument to make unsqueezing channel dimension optional in `PTDataset`

dhpitt:
Raised by #500. This fix makes the PTDataset compatible with both single-channel squeezed data (as in our example problems) and multi-channel datasets that are not squeezed.

### DIFF
--- a/neuralop/data/datasets/pt_dataset.py
+++ b/neuralop/data/datasets/pt_dataset.py
@@ -117,6 +117,9 @@ class PTDataset:
         output_data_dims = data["y"].ndim - 2
         if channels_squeezed:
             output_data_dims += 1
+            channel_slice = []
+        else:
+            channel_slice = [slice(None)]
         # convert None and 0 to 1
         if not input_subsampling_rate:
             output_subsampling_rate = 1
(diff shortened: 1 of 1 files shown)


neuralop/data/datasets/pt_dataset.py
```diff
@@ -76,6 +77,10 @@ def __init__(self,
             rate at which to subsample each output dimension, by default None
         channel_dim : int, optional
             dimension of saved tensors to index data channels, by default 1
+        channels_squeezed : bool, optional
+            whether the channel dim of saved tensors is squeezed in storage
```
JeanKossaifi, line 81:
```suggestion
            If the channels dim is 1, whether that is explicitly kept in the saved tensor. 
            If not, we need to unsqueeze it to explicitly have a channel dim. 
```

neuralop/data/datasets/pt_dataset.py
```diff
@@ -96,7 +101,9 @@ def __init__(self,
 
         # optionally subsample along data indices
         ## Input subsampling 
-        input_data_dims = data["x"].ndim - 1
+        input_data_dims = data["x"].ndim - 2 # batch and channels
+        if channels_squeezed:
+            input_data_dims += 1
```
JeanKossaifi, line 106: That's a little convoluted - could we potentially just do the unsqueeze first?

JeanKossaifi (commented):
Looks good, thank you.

neuralop/data/datasets/pt_dataset.py
dhpitt: Yep, changed

PR #513, opened 2025-01-07
add secret for test pypi (remove deprecated trusted publisher config)

dhpitt:
@JeanKossaifi : the simplest way for us to fix the Test-PyPI part of our action is to add a secret to Secrets --> Actions named `TEST_PYPI_API_KEY`, which I've changed the workflow to use.


JeanKossaifi:
Isn't that the old way? I had updated it to use their newer API.

dhpitt:
Closing, stale since we removed test-pypi from deployments

PR #520, opened 2025-01-15
update disco conv for new filter basis funcs

dhpitt:


neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -764,6 +813,7 @@ def __init__(
         in_shape: Tuple[int],
         out_shape: Tuple[int],
         kernel_shape: Union[int, List[int]],
+        filter_basis: str="piecewise_linear",
```
bonevbs, line 816: shall we call this `basis_type` for consistency with torch-harmonics?

bonevbs (approved):
LGTM apart from the naming suggestion. Feel free to merge

neuralop/layers/discrete_continuous_convolution.py
dhpitt: Sure

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -123,7 +147,10 @@ def _precompute_convolution_filter_matrix(grid_in,
     r = torch.sqrt(diffs[0] ** 2 + diffs[1] ** 2)
     phi = torch.arctan2(diffs[1], diffs[0]) + torch.pi
 
-    idx, vals = kernel_handle(r, phi)
+    if isinstance(basis_type, str):
+        # if a basis name is provided, instantiate it with the provided kernel shape
+        basis_type = basis_type_classes[basis_type](kernel_shape)
```
JeanKossaifi, line 152: What would be the alternative? For the user to pass directly a class, instantiated with the correct kernel shape? In this case, I'd keep only the str option.

neuralop/layers/discrete_continuous_convolution.py
```diff
@@ -416,7 +447,10 @@ class DiscreteContinuousConvTranspose2d(DiscreteContinuousConv):
         Can also pass a string to generate a regular (tensor) grid.
         For exact options see ``torch_harmonics.quadrature``.
     kernel_shape: Union[int, List[int]]
-        kernel shape. Expects either a single integer for isotropic kernels or two integers for anisotropic kernels
+        kernel shape. Expects either a single integer for isotropic kernels
+        or two integers for anisotropic kernels
```
JeanKossaifi, line 451: The formatting is weird. Also made me realize this description isn't super helpful. Can we explain the two terms and add more detail (or better, an example) for each cases?

JeanKossaifi (commented):
Looks good to me, aside from the two minor comments.

neuralop/layers/discrete_continuous_convolution.py
dhpitt: Done

neuralop/layers/discrete_continuous_convolution.py
dhpitt: Updated

PR #525, opened 2025-01-20
Make `domain_padding_mode` symmetric by default in FNO

dhpitt:


neuralop/layers/padding.py
```diff
@@ -5,29 +5,28 @@
...
+    This class works for any input resolution. Expects inputs of shape
+   ` `(batch-size, channels, d1, ...., dN)``
     """
 
     def __init__(
         self,
-        domain_padding,
-        padding_mode="one-sided",
+        domain_padding: Union[float, list],
+        padding_mode: str="symmetric",
```
JeanKossaifi, line 29: Let us be more specific if we are to use type hints.
```suggestion
        padding_mode: Literal["symmetric", "one-side"] = "symmetric",
```

JeanKossaifi (commented):
Looks good - just a small suggested for more specific hints.

neuralop/layers/padding.py
dhpitt: Done

PR #527, opened 2025-01-27
Reimplement `segment_csr` natively with tensor ops for speed

dhpitt:
Timing shows the new segment-csr, which uses broadcasting instead of einsum, to be up to **10 times faster**!


neuralop/layers/segment_csr.py
```diff
@@ -54,38 +60,23 @@ def segment_csr(
...
         else:
-            batched = False
             point_dim = 0
 
         # if batched, shape [b, n_out, channels]
         # otherwise shape [n_out, channels]
         output_shape = list(src.shape)
-        n_out = indptr.shape[point_dim] - 1
+        n_out = indptr.shape[0] - 1
+        n_reps = indptr[1:] - indptr[:-1]
```
JeanKossaifi, line 71: Shall we use a more informative name for indptr and add some more comments on what we do?

neuralop/layers/segment_csr.py
```diff
@@ -54,38 +60,23 @@ def segment_csr(
...
+
         output_shape[point_dim] = n_out
 
-        out = torch.zeros(output_shape, device=src.device)
+        indices = torch.arange(n_out).to(indptr.device).repeat_interleave(n_reps)
+        out = torch.zeros(output_shape, device=src.device).index_add_(point_dim, indices, src)
```
JeanKossaifi, line 76: We might want to add some more comments here or maybe break in two to make these clearer? Since the indices are fixed (1..n_out), perhaps we can do this without the repeat interleaves - e.g with broadcasting, need to think about it more.

neuralop/layers/segment_csr.py
```diff
@@ -54,38 +60,23 @@ def segment_csr(
...
+        if reduce == 'mean':
+            # scale the outputs by number of reduced neighbors, add eps to avoid div by zero
+            scale = (1 / (n_reps + eps)).unsqueeze(0).T
+            out = out * scale
```
JeanKossaifi, line 80: Seems odd to just have one case in the if. Isn't this the most efficient place to add quadrature? If we assume they are already in the features, wouldn't it be the same for the mean scaling factor?

JeanKossaifi (commented):
I like the refactoring, much clearer to pull the code within segment_csr!

neuralop/layers/segment_csr.py
dhpitt: Without the if, the code implements `reduce=="sum"` by default.

neuralop/layers/segment_csr.py
dhpitt: We could use broadcasting via a mask but that would add memory overhead.

JeanKossaifi:
Wonderful @dhpitt, let me know when it's ready!

dhpitt:
I used this code in practice as a test during development of #546 and found that as with neighbor search, the native method uses an insane amount of memory in real-world cases. Dropping until I can think of a better fix.

PR #536, opened 2025-02-11
Codano Model Ignores Second Layer in Mixer

MichalBravansky:


dhpitt:
Thanks for this catch @MichalBravansky ! @ashiq24 confirmed this fix is correct. Merging once tests pass

PR #538, opened 2025-02-12
Codano Bugfix

ashiq24:
Minor bug fix for variable extension mechanism. Added corresponding test cases.


dhpitt:
Thank you for the catch @ashiq24 . Merging.

PR #539, opened 2025-02-12
(fix) Allow calling `Trainer.train_one_epoch` before `Trainer.train` has been called

### LINKED ISSUE #537: `train_one_batch` requires `self.optimizer` which `Trainer` does not have
Hi there,

The function `train_one_batch` needs access to the `optimizer` attribute in this line

https://github.com/neuraloperator/neuraloperator/blob/3513714df6bf51e7aefc58fb059f27c6bf96a95e/neuralop/training/trainer.py#L407

but the class `Trainer` does not have this `optimizer` attribute

https://github.com/neuraloperator/neuraloperator/blob/3513714df6bf51e7aefc58fb059f27c6bf96a95e/neuralop/training/trainer.py#L25

Therefore, it results in the following error

```
>>> trainer.train_one_epoch(epoch, train_loader, train_loss)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/ubuntu/.local/lib/python3.10/site-packages/neuralop/training/trainer.py", line 263, in train_one_epoch
    loss = self.train_one_batch(idx, sample, training_loss)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/neuralop/training/trainer.py", line 407, in train_one_batch
    self.optimizer.zero_grad(set_to_none=True)
AttributeError: 'Trainer' object has no attribute 'optimizer'
```

Thanks in advance for your time to look at this.

dhpitt:
Fixes #537

### DIFF
--- a/neuralop/training/incremental.py
+++ b/neuralop/training/incremental.py
@@ -79,7 +79,7 @@ class IncrementalFNOTrainer(Trainer):
         if self.incremental_grad:
             self.grad_explained()
 
-    def train_one_epoch(self, epoch, train_loader, training_loss):
+    def train_one_epoch(self, epoch, train_loader, training_loss, optimizer=None, scheduler=None, regularizer=None):
         """train_one_epoch inherits from the base Trainer's method
             and adds the computation of the incremental-FNO algorithm
             before returning the training epoch's metrics. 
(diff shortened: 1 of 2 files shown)


JeanKossaifi:
What would be the usecase? Do we _want_ users to call that directly instead of letting the trainer do it?
If so do we need the class really? Might just be a collection of functions/or some could be classmethods.

I do like the clarity of passing model, opt and scheduler explicitly - should we add it everywhere?

ziyiyin97:
The potential use case could be that a user wants to run some customized callback function after each epoch. Food for thoughts.

JeanKossaifi:
I agree. We can close this and open an issue.

PR #540, opened 2025-02-13
Add hdiv norm

AleCorintis:
Added H(div) norm loss and its test

### DIFF
--- a/neuralop/losses/data_losses.py
+++ b/neuralop/losses/data_losses.py
@@ -417,6 +417,243 @@ class H1Loss(object):
         """
         return self.rel(y_pred, y, quadrature=quadrature)
 
+class HdivLoss(object):
+    """
+    HdivLoss provides the Hdiv Sobolev norm between
+    two d-dimensional discretized functions.
+
+    .. note :: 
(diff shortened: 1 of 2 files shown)


dhpitt:
Thank you for the contribution and the nicely formatted PR. Could you provide some motivation for when/why this loss is used? Thanks!

AleCorintis:
Thank you @dhpitt for reviewing my PR! 
The H(div) norm is useful for problems where the solution lies in a divergence-free function space. This typically appears in fluid dynamics, electromagnetics, and other PDE-constrained problems where enforcing the divergence condition is critical.
In particular, the H(div) norm loss ensures that the learned solution respects the conservation laws encoded in divergence constraints, which can lead to more physically accurate models, especially when solving PDEs like the incompressible Navier-Stokes equations, Maxwell’s equations, or Darcy’s flow in porous media.

Let me know if you’d like more details or any modifications! Thanks again for your time.

Also, would you like me to mark this PR as ready for review instead of a draft?

dhpitt:
This is an excellent contribution. Thanks @AleCorintis ! Merging.

PR #545, opened 2025-02-21
fix: Ensure weights match precision mode in SpectralConv

sebulo:
Previously weights remained in full precision even when using 'half' or 'mixed' precision modes. Now:
- Weights are initialized with correct dtype (torch.chalf) for half/mixed precision modes
- AdamW optimizer updated to handle complex32 (chalf) tensors by replacing addcmul_ with basic operations, as addcmul_ is not supported for chalf dtype.   Note: Optimizer states match gradient precision. Future work should evaluate if forcing full precision in optimizer would provide better training stability while maintaining memory benefits of half precision in forward pass.

### DIFF
--- a/neuralop/layers/spectral_convolution.py
+++ b/neuralop/layers/spectral_convolution.py
@@ -335,15 +335,23 @@ class SpectralConv(BaseSpectralConv):
 
         tensor_kwargs = decomposition_kwargs if decomposition_kwargs is not None else {}
 
-        # Create/init spectral weight tensor
-
+        # Determine weight dtype based on fno_block_precision
+        self.weight_dtype = {
+            "full": torch.cfloat,
+            "half": torch.chalf,
+            "mixed": torch.chalf
+        }.get(fno_block_precision, torch.cfloat)
+        
+        # Create/init spectral weight tensor with specified precision
         if factorization is None:
-            self.weight = torch.tensor(weight_shape, dtype=torch.cfloat)
+            self.weight = torch.empty(weight_shape, dtype=self.weight_dtype)
+            with torch.no_grad():
+                self.weight.normal_(0, init_std)
         else:
             self.weight = FactorizedTensor.new(weight_shape, rank=self.rank, 
                                      factorization=factorization, fixed_rank_modes=fixed_rank_modes,
-                                     **tensor_kwargs, dtype=torch.cfloat) 
-        self.weight.normal_(0, init_std)
+                                     **tensor_kwargs, dtype=self.weight_dtype)
+            self.weight.normal_(0, init_std)
         
         self._contract = get_contract_fun(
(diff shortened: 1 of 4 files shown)


dhpitt:
These are great changes @sebulo , thank you for making the fix.

dhpitt (approved):

neuralop/layers/spectral_convolution.py
```diff
@@ -335,15 +335,23 @@ def __init__(
...
         tensor_kwargs = decomposition_kwargs if decomposition_kwargs is not None else {}
 
-        # Create/init spectral weight tensor
-
+        # Determine weight dtype based on fno_block_precision
+        self.weight_dtype = {
+            "full": torch.cfloat,
+            "half": torch.chalf,
+            "mixed": torch.chalf
+        }.get(fno_block_precision, torch.cfloat)
```
JeanKossaifi, line 343: Looks good though, wouldn't we want to fail loudly rather than silently use cfloat in case of wrong type? 
I'd first define the dict then catch any KeyError and write an explicit message for users to help them fix the issue.

neuralop/layers/spectral_convolution.py
```diff
@@ -435,8 +444,9 @@ def forward(
             out_dtype = torch.chalf
         else:
             out_dtype = torch.cfloat
+            
         out_fft = torch.zeros([batchsize, self.out_channels, *fft_size],
-                              device=x.device, dtype=out_dtype)
+                            dtype=out_dtype, device=device)  # Use stored device
```
JeanKossaifi, line 449: Why the switch here? Wouldn't we want to always be on the same device as x?

neuralop/layers/spectral_convolution.py
```diff
@@ -482,24 +492,22 @@ def forward(
...
+        
         if self.order > 1:
             out_fft = torch.fft.fftshift(out_fft, dim=fft_dims[:-1])
         
         if self.complex_data:
-            x = torch.fft.ifftn(out_fft, s=mode_sizes, dim=fft_dims, norm=self.fft_norm)
+            out_fft = torch.fft.ifftn(out_fft, s=mode_sizes, dim=fft_dims, norm=self.fft_norm)
         else:
-            x = torch.fft.irfftn(out_fft, s=mode_sizes, dim=fft_dims, norm=self.fft_norm)
+            out_fft = torch.fft.irfftn(out_fft, s=mode_sizes, dim=fft_dims, norm=self.fft_norm)
```
JeanKossaifi, line 508: Why the change here? I don't think it should change anything for memory usage since we're not doing the operation inplace. We can always call x.add_ below if needed.

JeanKossaifi (commented):
Looks great overall, thanks for the fix @sebulo! Just left a few comments in the code.

PR #547, opened 2025-02-25
remove zarr required dep for compatibility with HPC clusters

dhpitt:
Installations on new clusters with NVIDIA Grace CPUs fail on installing `numcodecs`, which is a dependency of `zarr` . Our `zarr_dataset` is currently unused in the main build, so for now let's remove the dependency and make it optional for users who wish to use the `zarr_dataset`


JeanKossaifi (approved):
Looks good to me, thanks @dhpitt!

PR #551, opened 2025-03-06
fix(burgers): following changes to make burgers runnable

alundilong:
1. dataset: provide burgers_train_16.pt burgers_test_16.pt (shape changes)
2. burgers.py: set channel_squeezed to False otherwise add extra dimension
3. __init__.py: add load_min_burgers_1dtime otherwise function cannot be found
4. pt_dataset.py: dimension estimation should use latest dimension
5. train_burgers.py: find the dataset based on config.yaml
6. burgers_config.yaml: fix path to dataset


dhpitt:
Thank you for the PR and for opening #552 . Your changes are along the right track but will not replicate the original `load_burgers_1dtime` method (which isn't your fault, the method was just removed accidentally). I've opened #557 to address these issues properly.

PR #553, opened 2025-03-07
Reintroduce the use_channel_mlp option

vduruiss:
Reintroduce the option to activate or deactivate the ChannelMLP (by default True so existing codes still work)

I have encountered examples where FNO works significantly better without the ChannelMLP

I think there might have been discussions to rename it as ChannelMixing ? 
This could be the opportunity to do it at the same time

(Also cleaned up some outdated default values in the docstrings)


JeanKossaifi:
I agree - not sure why we removed the option. Looks good to me.

JeanKossaifi:
Merging.

PR #554, opened 2025-03-07
Numerical Stability of Relative Losses

vduruiss:
Adding an epsilon to the denominators of relative losses to avoid numerical instability when `ynorm` is close to 0

Currently set to `eps=1e-8` since it seems to be a fairly common value in torch/scipy/numpy

### DIFF
--- a/neuralop/losses/data_losses.py
+++ b/neuralop/losses/data_losses.py
@@ -52,6 +52,8 @@ class LpLoss(object):
 
             ``LpLoss`` always reduces over the spatial dimensions according to ``self.measure``.
             `reduction` only applies to the batch and channel dimensions.
+    eps : float, optional
+        small number added to the denominator for numerical stability when using the relative loss
 
     Examples
     --------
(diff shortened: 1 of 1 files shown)


neuralop/losses/data_losses.py
```diff
@@ -64,6 +64,7 @@ def __init__(self, d=1, p=2, measure=1., reduction='sum'):
 
         self.d = d
         self.p = p
+        self.eps = 1e-8
```
dhpitt, line 67: Should this be parameterized?

dhpitt:
Thanks for the PR @vduruiss ! One small question about whether it would be worth parameterizing `self.eps`

vduruiss:
@dhpitt We can certainly parametrize it, either as an argument of the `__init__()` for the loss class, or as an input parameter of the `rel()` function, whichever you think would be best practice

JeanKossaifi:
Thanks! I would keep it in the __init__ so the loss can then be used as a function, without the need for additional arguments, e.g. 
```python
train_loss = Loss(all_args_here)
...
loss = train_loss(ground_truth, prediction)
```

dhpitt:
I agree, that sounds good

JeanKossaifi:
Awesome, merging, thanks you!

PR #556, opened 2025-03-08
Finite Differences for Point Clouds in Arbitrary Dimensions

vduruiss:
I wrote a code for non-uniform finite differences on arbitrary point clouds in 2D, and  thought it might eventually be useful to others, so I generalized it to arbitrary dimensions and cleaned it up.

The formulation of the approach is written in details in the appendix of the Mollified GNO paper for the 2D case. It obtains the FD stencil weights by solving appropriate linear least squares problem given a fixed number of nearest neighbors. It can suffer from ill-conditioning so I added the option to regularize the least squares problem.

I included tests in 1D and 2D to check that it's accurate within 5% L2 relative error (usually is around 1% on these examples), and commented codes to plot the resulting derivatives. I'm including the test plots below for convenience

image

image


dhpitt:
Fantastic PR @vduruiss . Especially nice that you added documentation and tests. This would make an excellent addition to the library. Sending to @JeanKossaifi for another review.

vduruiss:
I added the example numerical solver codes using structured central finite differences. 

I included the gifs for reference, but we should probably remove them before merging.

vduruiss:
Following @JeanKossaifi 's suggestion, I added a radius cutoff for the neighbors taken into account in the finite differences.

It mostly helps avoid taking too many neighbors in regions where the point cloud is not very dense. For instance in the example below, I used a low resolution (32 points) but a relatively too large number of neighbors (10) for FD. The radius cutoff of 0.05 reduced the number of neighbors effectively used in the FD computations (typically to 3-6 neighbors) and improved performance

Reference

By taking a large number of neighbors and then tuning the radius, we can also essentially make the number of neighbors specified irrelevant and have something that aligns more with the neural operator paradigm.


Note that I hardcoded a minimum value of neighbors of 3 (including the point itself)

Also, note that the radius cutoff is implemented as a mask which is non-differentiable.

JeanKossaifi:
Thanks @vduruiss, this looks great! Could we make the examples in the Sphinx gallery format, in https://github.com/neuraloperator/neuraloperator/tree/main/examples (we can add an `fd_solver` folder there), so they get rendered in the website: https://neuraloperator.github.io/dev/auto_examples/index.html

neuralop/losses/finite_diff.py
vduruiss, line 1: Lines 79-82: Update docstring so it says "central_diff_3d computes derivatives 
    df(x,y,z)/dx, df(x,y,z)/dy, and df(x,y,z)/dz, for f(x,y,z) defined 
    on a regular 3d grid using finite-difference"

```
def central_diff_3d(x, h, fix_x_bnd=False, fix_y_bnd=False, fix_z_bnd=False):
    """central_diff_3d computes derivatives 
    df(x,y,z)/dx and df(x,y,z)/dy for f(x,y,z) defined 
    on a regular 2d grid using finite-difference
 ```

vduruiss: It might be useful to have an additional argument for the central_diff functions to specify which dimension to differentiate with respect to, and avoid having to compute all derivatives when only one is needed. For instance in the Poisson equation, we still compute the u_xy and u_yx derivatives which takes unnecessary time and memory

vduruiss:
> Thanks @vduruiss, this looks great! Could we make the examples in the Sphinx gallery format, in https://github.com/neuraloperator/neuraloperator/tree/main/examples (we can add an `fd_solver` folder there), so they get rendered in the website: https://neuraloperator.github.io/dev/auto_examples/index.html

@JeanKossaifi  I added .ipynb tutorial versions of the codes, but I don't know how to use Sphinx so they rendered in the website

JeanKossaifi:
Thanks @vduruiss!

Mainly we want the files to be named plot_xyz, i.e. filename starting with `plot_`: [syntax](https://sphinx-gallery.github.io/stable/syntax.html), or alternatively [syntax for Jupyter notebooks](https://sphinx-gallery.github.io/stable/tutorials/index.html) and an [example](https://sphinx-gallery.github.io/stable/auto_examples/plot_0_sin.html).
For [animations](https://sphinx-gallery.github.io/stable/auto_examples/plot_8_animations.html), we need to turn the option on.

dhpitt:
Updated in #576

PR #560, opened 2025-03-20
(fix) Proper kernel size for morlet basis, add tests

dhpitt:
@samaonline found an issue with DISCO layers when `basis_type='morlet'`. I added basis type parameterization to the unit tests and uncovered a few bugs introduced during the update to `torch_harmonics==0.7.4`.

* `kernel_shape` is now always 2d
* `kernel_size` is computed differently for `'morlet'` basis


samaonline:
Looks good to me! thanks

PR #563, opened 2025-03-21
Load state weights only

dhpitt:


JeanKossaifi:
Looks good - though the tests fail.

dhpitt:
Fixed the issues in the doc and in `neuralop.models.UQNO` causing tests to fail

JeanKossaifi (changes requested):
Thanks @dhpitt, this fixes serialization issue. 
I'd add a function though, e.g. `get_nonlinearity` instead of indexing the dictionary `nonlinearity_modules[non_linearity]`. The function can still internally use the dict but we can at least catch the `KeyError` and print a useful message to user.

JeanKossaifi:
Was this ready to merge @dhpitt ?

dhpitt:
No, this was pending some kind of fix to the sub module saving procedure

PR #572, opened 2025-03-31
(fix) set use_scatter=False in segmentcsr tests

dhpitt:
Test native behavior. The defaults work out here on the runner but this will fix breaking tests on local machines


JeanKossaifi:
Looks good to me - should we aim to parametrize this down the line and have a check for whether it's available or not?

dhpitt:
I think there's nothing stopping us from adding this currently

dhpitt:
Updated PR to parametrize this in all model/layer tests that use a GNO. I think the setting should always be false in `test_segment_csr`, because when `use_scatter` is True, we'd just be testing a function imported from another module.

dhpitt:
Tests break due to issue fixed in #575

PR #574, opened 2025-04-07
Cleanup Base Model

JeanKossaifi:
Only override load_state_dict and state_dict so our objects are compatible with regular PyTorch saving/loading out of the box. I would argue we should remove load/save_checkpoint completely and always use `torch.save(model.state_dict())` or `Model.from_checkpoint(...)` if needed.

### DIFF
--- a/config/burgers_config.py
+++ b/config/burgers_config.py
@@ -1,0 +1,99 @@
+from typing import Any, List, Optional
+
+
+from zencfg import ConfigBase
+
+
+
+
+
+class Distributed(ConfigBase):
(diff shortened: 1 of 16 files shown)


JeanKossaifi:
Tests need to be updated as now models are always self-contained and metadata is in the main state_dict. Also passing explicitly checkpoint_path to match pytorch instead of path + name separately.

dhpitt (approved):
Looks great. Thanks for this update @JeanKossaifi !

dhpitt:
closing due to supsersede by #563

PR #576, opened 2025-04-11
Finite-difference on point clouds, plus interactive examples

dhpitt:


JeanKossaifi:
Great, thank you both @vduruiss @dhpitt, merging!

PR #584, opened 2025-04-20
Feat/layers laplace conv

sinapordanesh:
# ✨ Laplace Neural Operator (LNO) ✨
This PR introduces **Laplace‑based Spectral Convolutions** in 1‑D, 2‑D, and 3‑D, along with unit tests and a small utility for shape enforcement.

**Primary references**

| Resource | Link |
|----------|------|
| Paper | <https://arxiv.org/pdf/2303.10528> |
| Original implementation | <https://github.com/qianyingcao/Laplace-Neural-Operator> |

---

## 1  |  Why a series of small PRs?
Implementing Laplace and Wavelet kernels touches multiple layers of the library (kernels → blocks → models).  
To keep each review focused and mergeable, I’m submitting a **bottom‑up sequence** of PRs:

1. **Spectral kernels** (Laplace ✓, Wavelet → upcoming)  
2. **Block‑level integration** (FNOBlocks / XNOBlocks)  
3. **Model‑level wiring** (FNO, TFNO, etc.)  
4. **Docs & gallery example refresh**

Each PR contains **fully‑documented code, helper utilities, unit tests, and inline typing**.  
I don’t expect immediate approval; instead, I’m committed to an iterative review process and will revise as many times as needed to meet project standards.

---

## 2  |  What’s in this PR

| File | Purpose |
|------|---------|
| `neuralop/layers/spectral_convolution_laplace.py` | New `SpectralConvLaplace{1,2,3}D` layers following core API conventions (FFT shape‑handling, `modes`, bias, complex weights). |
| `neuralop/layers/shape_enforcer.py` | Utility to ensure inverse FFT outputs match expected spatial dims (PyTorch handles this for FFT; Laplace kernels need an explicit helper). |
| `neuralop/layers/tests/test_spectral_convolution_laplace.py` | Unit tests for forward pass, mode truncation, etc. |
| `neuralop/layers/tests/test_shape_enforcer.py` | Edge‑case tests for cropping/padding logic. |

---

## 3  |  Review notes & open questions

* **API surface** – I tried to mirror `SpectralConv` naming (`in_channels`, `out_channels`, etc.), and consider all those arguemtns in each new spectral convolution classes, although not all of them being used. Just keep them for future developement. 
* **ShapeEnforcer location** – currently in `neuralop.layers`; relocate if there’s a utilities sub‑package you’d rather use.

---

## 4  |  Additional references
For a detailed comparison between the Laplace kernels introduced in this PR and the original paper’s implementation, see **LNO.md**:

[LNO.md](https://github.com/user-attachments/files/19823392/LNO.md)

The note documents each kernel, explains the adaptations made for `neuralop`, and provides accuracy checks confirming numerical parity with the reference code.

---

*Thank you for your time and feedback. I’m available to iterate until the code fits seamlessly into* **neuraloperator** *and its CI.*

### DIFF
--- a/neuralop/layers/shape_enforcer.py
+++ b/neuralop/layers/shape_enforcer.py
@@ -1,0 +1,90 @@
+import torch
+import torch.nn.functional as F
+from typing import Optional, Tuple, List
+
+
+class ShapeEnforcer(torch.nn.Module):
+    """
+    A PyTorch module that enforces a specific output shape on the tensor
+    by cropping or padding the spatial/temporal dimensions.
+
+    This class is useful in scenarios where you need consistent output shapes
+    after operations like spectral or wavelet transforms, where dimensions might
+    not match the desired size naturally.
+
+    Parameters
+    ----------
+    start_dim : int, optional
+        The dimension index at which to start enforcing the shape.
+        For example, with typical 2D data `[B, C, H, W]`, `start_dim=2`.
+        For 1D data `[B, C, L]`, `start_dim=2`. Default is 2.
+
+    Methods
+    -------
+    forward(x: torch.Tensor, output_shape: Optional[List[int]] = None) -> torch.Tensor
+        Enforces the desired shape on the input tensor.
+    """
+
+    def __init__(self, start_dim: int = 2):
+        """
+        Initialize the ShapeEnforcer module.
+
+        Parameters
+        ----------
+        start_dim : int, optional
+            The dimension index at which to start enforcing the shape.
+            Default is 2.
+        """
+        super(ShapeEnforcer, self).__init__()
+        self.start_dim = start_dim
+
+    def forward(
+        self, x: torch.Tensor, output_shape: Optional[List[int]] = None
+    ) -> torch.Tensor:
+        """
+        Forward method to enforce the desired shape on the input tensor.
+
+        Parameters
+        ----------
+        x : torch.Tensor
+            Input tensor of shape `[B, C, D1, D2, ..., DN]`.
+        output_shape : List[int], optional
+            The desired shape of the tensor's spatial/temporal dimensions,
+            e.g., `[D_out1, D_out2, ...]`. If None, the tensor is returned unchanged.
+
+        Returns
+        -------
+        torch.Tensor
+            Tensor with shapes from `self.start_dim` onward forced to match `output_shape`.
+        """
+        if output_shape is None:
+            return x  # No shape enforcement required
+
+        # -- 1) Crop if needed --
+        # Build slicing for each dimension
+        slices = [slice(None)] * self.start_dim  # keep [B, C, ...] as is
+        for i, size_desired in enumerate(output_shape):
+            current_size = x.shape[self.start_dim + i]
+            # If current is bigger, crop
+            if current_size > size_desired:
+                slices.append(slice(0, size_desired))
+            else:
+                slices.append(slice(0, current_size))
+        x = x[tuple(slices)]
+
+        # -- 2) Pad if needed --
+        # F.pad expects a pad-list in reverse order: [dimN_left, dimN_right, dimN-1_left, ...]
+        pad_list = []
+        for i in reversed(range(len(output_shape))):
+            size_desired = output_shape[i]
+            current_size = x.shape[self.start_dim + i]
+            if current_size < size_desired:
+                pad_amount = size_desired - current_size
+                # Pad only on the "end" (right side).
+                pad_list.extend([0, pad_amount])
+            else:
+                pad_list.extend([0, 0])
+        if any(pad_list):
+            x = F.pad(x, pad_list, mode="constant", value=0.0)
+
+        return x
--- a/neuralop/layers/spectral_convolution_laplace.py
+++ b/neuralop/layers/spectral_convolution_laplace.py
@@ -1,4 +1,3 @@
-
 import torch
 import torch.nn as nn
 import numpy as np
@@ -12,22 +11,19 @@ from .base_spectral_conv import BaseSpectralConv
 Number = Union[int, float]
 
 
-def _compute_dt(
-    shape, 
-    start_points: List=None, 
-    end_points: List=None
-    ):
+def _compute_dt(shape, start_points: List = None, end_points: List = None):
     """
     Compute uniform spacing (dt) for each dimension based on domain lengths, step sizes,
     start points, and end points. Defaults to a unit domain if not specified.
 
     Parameters:
+    ----------
     shape (Sequence[int]): The shape of the input excluding batch and channel, i.e. (d_1, d_2, ..., d_n).
-    step_sizes (Sequence[float], optional): Step sizes for each dimension. Defaults to shape-based uniform spacing.
     start_points (Sequence[float], optional): Start points for each dimension. Defaults to 0.0 for all dimensions.
     end_points (Sequence[float], optional): End points for each dimension. Defaults to 1.0 for all dimensions.
 
     Returns:
+    -------
     dt_list (Sequence[float]): A list of spacings, one per dimension.
     grid (List[torch.Tensor]): A list of grid points for each dimension based on the spacing and domain.
     """
@@ -40,13 +36,18 @@ def _compute_dt(
 
     # Validate that start_points and end_points match the number of dimensions
     if len(start_points) != dim or len(end_points) != dim:
-        raise ValueError("Start points and end points must match the number of input dimensions ({dim}).")
+        raise ValueError(
+            "Start points and end points must match the number of input dimensions ({dim})."
+        )
 
     # Compute domain lengths from start and end points
     domain_lengths = [end_points[i] - start_points[i] for i in range(dim)]
 
     # Generate grid points for each dimension using torch.linspace
-    grid = [torch.linspace(start_points[i], end_points[i], steps=shape[i]) for i in range(dim)]
+    grid = [
+        torch.linspace(start_points[i], end_points[i], steps=shape[i])
+        for i in range(dim)
+    ]
 
     # Compute dt directly from the grid
     dt_list = [(grid[i][1] - grid[i][0]).item() for i in range(dim)]
@@ -54,15 +55,64 @@ def _compute_dt(
     return dt_list, grid
 
 
-
 # ====================================
 #  Laplace layer: pole-residue operation is used to calculate the poles and residues of the output
 # ====================================
 class SpectralConvLaplace1D(BaseSpectralConv):
+    """Implements 1D Laplace-based Spectral Convolution.
+
+    This class implements a spectral convolution layer using pole-residue formulation
+    for 1D data. Instead of using standard Fourier transforms, it uses pole-residue
+    operations to calculate the response in the frequency domain.
+
+    Parameters
+    ----------
+    in_channels : int
+        Number of input channels
+    out_channels : int
+        Number of output channels
+    n_modes : int or int tuple
+        Number of modes to use for pole-residue calculation
+    complex_data : bool, optional
+        whether data takes on complex values in the spatial domain, by default False
+    max_n_modes : int tuple or None, default is None
+        Maximum number of modes to keep in the Laplace layer
+    bias : bool, default is True
+        whether to include bias term
+    separable : bool, default is False
+        whether to use separable implementation
+    resolution_scaling_factor : float or float list, optional
+        Factor to scale the output resolution
+    xno_block_precision : str, default is "full"
+        Precision to use for the convolution block
+    rank : float, optional
+        Rank parameter, by default 0.5
+    factorization : str or None, optional
+        Type of factorization to use
+    implementation : str, default is "reconstructed"
+        Implementation approach for the convolution
+    fixed_rank_modes : bool, default is False
+        Whether to use fixed rank modes
+    decomposition_kwargs : dict, optional
+        Additional parameters for decomposition
+    init_std : str or float, default is "auto"
+        Standard deviation for weight initialization
+    fft_norm : str, default is "forward"
+        Normalization method for FFT
+    device : torch.device, optional
+        Device to use for computation
+    linspace_steps : list, optional
+        Number of steps for each dimension
+    linspace_startpoints : list, optional
+        Start points for each dimension
+    linspace_endpoints : list, optional
+        End points for each dimension
+    """
+
     def __init__(
-        self, 
-        in_channels, 
-        out_channels, 
+        self,
+        in_channels,
+        out_channels,
         n_modes,
         complex_data=False,
         max_n_modes=None,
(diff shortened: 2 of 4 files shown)


neuralop/layers/tests/test_spectral_convolution_laplace.py
```diff
@@ -0,0 +1,317 @@
+# MIT License
+# Copyright (c) 2024 Saman Pordanesh
```
dhpitt, line 2: I would remove. Authors don't own individual files in the library

neuralop/layers/spectral_convolution_laplace.py
```diff
@@ -0,0 +1,979 @@
...
+    if len(start_points) != dim or len(end_points) != dim:
+        raise ValueError(
+            "Start points and end points must match the number of input dimensions ({dim})."
+        )
+
+    # Compute domain lengths from start and end points
+    domain_lengths = [end_points[i] - start_points[i] for i in range(dim)]
+
+    # Generate grid points for each dimension using torch.linspace
+    grid = [
```
dhpitt, line 47: In `neuraloperator`, we provide a module for performing grid embeddings that you can pass directly into your model (`neuralop.layers.embeddings.GridEmbeddingND`)

neuralop/layers/spectral_convolution_laplace.py
```diff
@@ -0,0 +1,979 @@
...
+
+    def __init__(
+        self,
+        in_channels,
+        out_channels,
+        n_modes,
+        complex_data=False,
+        max_n_modes=None,
+        bias=True,
+        separable=False,
```
dhpitt, line 120: A lot of these parameters seem unused.

neuralop/layers/spectral_convolution_laplace.py
```diff
@@ -0,0 +1,979 @@
...
+from typing import List, Optional, Tuple, Union
+from .resample import resample
+from ..utils import validate_scaling_factor
+from .shape_enforcer import ShapeEnforcer
+from .base_spectral_conv import BaseSpectralConv
+
+Number = Union[int, float]
+
+
+def _compute_dt(shape, start_points: List = None, end_points: List = None):
```
dhpitt, line 14: Better to use the provided `gridembedding` layer

neuralop/layers/spectral_convolution_laplace.py
```diff
@@ -0,0 +1,979 @@
...
+    # Compute dt directly from the grid
+    dt_list = [(grid[i][1] - grid[i][0]).item() for i in range(dim)]
+
+    return dt_list, grid
+
+
+# ====================================
+#  Laplace layer: pole-residue operation is used to calculate the poles and residues of the output
+# ====================================
+class SpectralConvLaplace1D(BaseSpectralConv):
```
dhpitt, line 61: As a general note, it would be much better if you could find a way to refactor this logic for a dimension-agnostic case. From checking your code, nothing jumps out at me as impossible to refactor into more abstract terms, though I may be missing something.

dhpitt (changes requested):
Thanks for kicking this off, @sinapordanesh . Thanks for following our conventions for commenting style as well!

I left a preliminary review with a few notes, now passing to @JeanKossaifi . My main concerns are:

1. Is there a reason the laplace convs are implemented specifically for each dimension? Over time, this makes the code more difficult to maintain and adds class proliferation. Unless there are specific elements of the computation that can't be refactored, I would suggest refactoring so that the dimension is abstracted. 
2. I don't really get the purpose of the `ShapeEnforcer`; I don't see a corresponding module in the original LNO code, and I don't see any mentions of cropping or padding in the code or the original paper. We're concerned this may impact the method's discretization-invariance.

neuralop/layers/shape_enforcer.py
```diff
@@ -0,0 +1,90 @@
+import torch
+import torch.nn.functional as F
+from typing import Optional, Tuple, List
+
+
+class ShapeEnforcer(torch.nn.Module):
```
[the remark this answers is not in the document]
sinapordanesh: No, it's because of integrating with the package, not originally from the paper. 
This is a copy of what parameter `s` does at:
`x = torch.fft.ifftn(out_fft, s=mode_sizes, dim=fft_dims, norm=self.fft_norm)`

`torch.fft` makes many tasks more manageable for implementing fourier-based transformations, but as Laplace and wavelet transforms are not supported in Torch, we need to implement some features from scratch, like this feature to shape the output `x` as the input `x`.

Do you think it makes sense?

neuralop/layers/spectral_convolution_laplace.py
sinapordanesh: Yes, considering all you've used in the original spectral_conv.py classes, for possible future feature adding. 

Do you like them to be removed for now?

neuralop/layers/tests/test_shape_enforcer.py
```diff
@@ -0,0 +1,110 @@
+# MIT License
+# Copyright (c) 2024 Saman Pordanesh
+
+import pytest
+import torch
+import numpy as np
+from neuralop.layers.shape_enforcer import ShapeEnforcer
+
+
+@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
sinapordanesh: Done, check it out after updating my pull request.

neuralop/layers/spectral_convolution_laplace.py
sinapordanesh: check this out [1c398d9](https://github.com/neuraloperator/neuraloperator/pull/584/commits/1c398d94bed117f62615096055105516db05905e)

(`GridEmbeddingND` wasn't that flexible to give me the dt directly. Would make the code more complicated.)

neuralop/layers/spectral_convolution_laplace.py
sinapordanesh: check this out [1c398d9](https://github.com/neuraloperator/neuraloperator/pull/584/commits/1c398d94bed117f62615096055105516db05905e)

(`GridEmbeddingND` wasn't that flexible to give me the dt directly. Would make the code more complicated.)

neuralop/layers/spectral_convolution_laplace.py
dhpitt: Parameters should only exist if they are used in the code. If/when they are needed, we can add them

neuralop/layers/spectral_convolution_laplace.py
```diff
@@ -0,0 +1,979 @@
...
+
+        # Transform to frequency domain and compute frequency grid
+        alpha = torch.fft.fft(x, dim=-1)
+        lambda0 = (
+            torch.fft.fftfreq(t.shape[0], dt, device=alpha.device) * 2 * np.pi * 1j
+        )
+        lambda1 = lambda0.unsqueeze(-1).unsqueeze(-1).unsqueeze(-1)
+
+        # Extract pole and residue weights from the combined weight tensor
+        weights_pole = self.weight[:, :, :modes1].view(
```
dhpitt, line 266: Would it be simpler to store the pole and residue weights separately?

neuralop/layers/spectral_convolution_laplace.py
```diff
@@ -0,0 +1,979 @@
...
+        tx = shape[1].to(x.device)
+        ty = shape[2].to(x.device)
+        dtz = dt_list[0]  # this can be time dimension, instead of Z dimension
+        dtx = dt_list[1]
+        dty = dt_list[2]
+
+        # Transform to frequency domain and compute frequency grids
+        alpha = torch.fft.fftn(x, dim=[-3, -2, -1])
+        omega1 = (
+            torch.fft.fftfreq(tz.shape[0], dtz, device=alpha.device) * 2 * np.pi * 1j
```
dhpitt, line 867: These fftfreq operations could be chained using iterables/loops to abstract away dimensions in this part

neuralop/layers/tests/test_shape_enforcer.py
```diff
@@ -5,103 +5,134 @@
 
 
 @pytest.mark.parametrize(
-    "test_id, start_dim, input_shape, output_shape, description",
+    "start_dim, input_shape, output_shape",
```
dhpitt, line 8: This isn't what I meant. I mean that these tests should happen in 1d, 2d, and 3d, and you should also parameterize the index of the start/stop dimensions explicitly, then compute the proper shapes inside the tests.

neuralop/layers/shape_enforcer.py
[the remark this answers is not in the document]
dhpitt: I see. In that case, I wonder if it might be better to implement the laplace and wavelet transforms independently, then allow users to choose their transform type (e.g. fourier vs. laplace vs. wavelet) inside the convolution. What do you think, @JeanKossaifi ?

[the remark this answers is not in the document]
sinapordanesh: > I see. In that case, I wonder if it might be better to implement the laplace and wavelet transforms independently, then allow users to choose their transform type (e.g. fourier vs. laplace vs. wavelet) inside the convolution. What do you think, @JeanKossaifi ?

Yes, that’s already how I’ve structured it. The wavelet implementation hasn’t been pushed yet. As noted in the main PR comment, I’m splitting changes into separate pull requests to keep each one focused and more manageable to review. We can finish this PR for LNO first, and I’ll introduce additional features afterward.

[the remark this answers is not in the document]
sinapordanesh: > Also, have you done any empirical testing with output scaling to confirm that the layers perform as expected?

Can you give an example of the kinds of testing you think of? A test case on FNO would help.

neuralop/layers/spectral_convolution_laplace.py
sinapordanesh: Ok, sure
I'll remove extra parameters here.

neuralop/layers/tests/test_shape_enforcer.py
sinapordanesh: This should address your point [d54a405](https://github.com/neuraloperator/neuraloperator/pull/584/commits/d54a405ef9badc41a9cbecb44ade85734ec1c5da)

neuralop/layers/tests/test_shape_enforcer.py
```diff
@@ -5,103 +5,134 @@
...
+    if x.shape[3] < output_shape[1]:
+        assert torch.all(y[:, :, :, x.shape[3]:] == 0)
 
-    # Test case specific assertions
+
+@pytest.mark.parametrize(
+    "start_dim, input_shape, output_shape, description",
+    [
+        (2, (2, 3, 10, 15), [10, 15], "Same shape - no changes needed"),
+        (2, (2, 3, 10, 15), None, "None output shape - return input unchanged"),
```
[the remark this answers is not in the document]
sinapordanesh: This should address your point [d54a405](https://github.com/neuraloperator/neuraloperator/pull/584/commits/d54a405ef9badc41a9cbecb44ade85734ec1c5da)

neuralop/layers/spectral_convolution_laplace.py
sinapordanesh: Let me start remastering the Laplace convolution classes to have an abstract version like the one we have for FNO. In the best case, we should try to have N-dimensional, but at least we can try to have 1- 3D on a single class. 
I'll address this point there.

neuralop/layers/spectral_convolution_laplace.py
sinapordanesh: Let me start remastering the Laplace convolution classes to have an abstract version like the one we have for FNO. In the best case, we should try to have N-dimensional, but at least we can try to have 1- 3D on a single class.

dhpitt (commented):
These changes are great, especially the general dimension-agnostic laplace convolution. Thanks! I would suggest removing the original 1, 2, and 3-D layers and performing all your tests on the new layer. Also added a few small comments.

dhpitt:
One note on the correctness test: A relative tolerance of 5e-1 is really high. If the outputs do not pass a lower tolerance, e.g. 1e-5 or 1e-7, this means there is a numerical error in your method.

sinapordanesh:
That's right, but in my investigation, I found the original LNO a little bit buggy. More confident in the N-dim kernel, which follows a clear path. 
However, I'll inspect further to find what's causing these differences, and report or fix it on either side.

neuralop/layers/spectral_convolution_laplace.py
```diff
@@ -104,15 +103,22 @@ class SpectralConvLaplace(BaseSpectralConv):
         Domain start points for each dimension. Defaults to 0.0 for all.
     linspace_endpoints : list of float, optional
         Domain end points for each dimension. Defaults to 1.0 for all.
+        
+    References
+    -----------
+    .. [1] :
+
```
dhpitt: nit: remove these lines

dhpitt: e.g. 

```
.. [1] : Author, Paper name.

**Comment by sinapordanesh** [comment]
@JeanKossaifi There was a typo in the test script, causing failures on all tests. 
It should be resolved now.

File: neuralop/layers/spectral_convolution_laplace.py [hunk 13]
```diff
@@ -0,0 +1,415 @@
...
+        Domain start points for each dimension. Defaults to 0.0 for all.
+    linspace_endpoints : list of float, optional
+        Domain end points for each dimension. Defaults to 1.0 for all.
+        
+    .. [1] : Cao, Q., Goswami, S., and Karniadakis, G. E. "LNO: Laplace Neural Operator
+        for Solving Differential Equations" (2023). arXiv preprint arXiv:2303.10528,
+        https://arxiv.org/pdf/2303.10528.
+    
+    
+
```
JeanKossaifi: Few new lines here.

neuralop/layers/tests/test_spectral_convolution_laplace.py
```diff
@@ -0,0 +1,51 @@
...
+        assert list(res.shape[2:]) == list(expected_output_shape[2:])
+    
+    else:
+        # Forward pass
+        res = conv(x)
+        
+        # Check output shape
+        assert res.shape[0] == batch_size
+        assert res.shape[1] == out_channels
+        assert list(res.shape[2:]) == [12]*dim
```
JeanKossaifi: Is there any test we could run besides just checking the shape?

JeanKossaifi (commented):
Thank you @sinapordanesh, great contribution!

Might be useful to run black or some other tool for pep8, etc. 
Also would be great to have some doc, perhaps an example? We can also add the new block to the API.

neuralop/layers/spectral_convolution_laplace.py
sinapordanesh: Added a brief documentation from the abstract of the LNO paper. Let me know if you'd like to see more detailed docs.

neuralop/layers/tests/test_spectral_convolution_laplace.py
sinapordanesh: Thanks for reminding me of this. I added a few more test assertions to cover more class features. 
Tried to match what you've implemented in test_spectral_conv.py
The LNO class has fewer arguments to be tested, and mainly the output shape. 

I've added weights complex value assurance, output real value, and testing domain description for each dimension on the LNO initialisation (important as we need to specify the domain for each problem, based on the physical domain)

sinapordanesh:
> Might be useful to run black or some other tool for pep8, etc.
> Also would be great to have some doc, perhaps an example? We can also add the new block to the API.

For pep8 style, ran black command for all four files contained in this PR. 

Regarding block for LNO and WNO, and main classes like LNO.py, we discussed that it's a better practice first to merge spectral convolutions, then add blocks and main classes as separate PRs, to avoid complex PRs.
I'll implement those as soon as we finalise the accepted versions of each spectral convolution.

aaprasad:
Is this going to be merged any time soon? Would love to try it out!

PR #586, opened 2025-04-22
Escape characters compatible with python >= 3.12

acotino-ignitioncomputing:
Minor correction in docstrings: changed escape characters to prevent 'SyntaxError' in python versions >= 3.12


dhpitt:
Thanks for opening this fix, @acotino-ignitioncomputing !

PR #592, opened 2025-05-17
Start updating config

JeanKossaifi:


dhpitt:
Merging as two separate commits to retain the history showing how configs were converted.

PR #593, opened 2025-05-20
Config to zencfg

dhpitt:


config/navier_stokes_config.py
```diff
@@ -0,0 +1,47 @@
...
+        n_train=10000,
+        train_resolution=128,
+        n_tests=[2000],
+        test_resolutions=[128],
+        test_batch_sizes=[8],
+        encode_input=True,
+        encode_output=True,
+    )
+    patching: PatchingConfig = PatchingConfig()
+    wandb: WandbConfig = WandbConfig(
```
JeanKossaifi, line 42: No need for these params.

config/navier_stokes_config.py
```diff
@@ -0,0 +1,47 @@
...
+    opt: OptimizationConfig = OptimizationConfig(
+        n_epochs=600,
+        learning_rate=3e-4,
+        training_loss="h1",
+        weight_decay=1e-4,
+        scheduler="StepLR",
+        step_size=100,
+        gamma=0.5,
+    )
+    data: NavierStokesDatasetConfig = NavierStokesDatasetConfig(
```
JeanKossaifi, line 31: Let's put all the values for NS in the main one I think, we don't need to override it here. If you want to be able to inspect, just put the whole NavierStokesDatasetConfig declaration in this file.

config/opt.py
```diff
@@ -0,0 +1,63 @@
+from typing import Optional, Literal
+from zencfg import ConfigBase
+
+class OptimizationConfig(ConfigBase):
```
JeanKossaifi, line 4: I thought we decided to not have docstrings for these. Also they're not parameters technically.

neuralop/models/base_model.py
```diff
@@ -207,21 +207,21 @@ def get_model(config):
     model : nn.Module
         the instanciated module
     """
-    arch = config["arch"].lower()
-    config_arch = config.get(arch)
+    arch = config.model['model_arch'].lower()
```
JeanKossaifi, line 210: Why don't we pass the model_config only, directly?

JeanKossaifi (changes requested):
Thanks @dhpitt!
Seems the tests no longer pass. Left some comments in the code. Could you rebase for clean history once it's fixed?

PR #610, opened 2025-06-05
Enable "batch_norm" in FNO class

klockwood19:
Added "batch_norm" as a normalization option in the FNO class constructor. This option appears in the FNO_blocks constructor but is absent in the FNO constructor


JeanKossaifi:
Thank you @klockwood19 for the PR, it looks good to me, merging.

PR #611, opened 2025-06-06
Loss Reweighing Schemes + PINO Trainer

vduruiss:
Added 2 loss reweighing schemes (Relobralo and SoftAdapt) which don't require gradient evaluations.

Made a version that works with the Trainer, and then also added a slightly modified version of the code from NVIDIA PhysicsNeMo which is more convenient to use with an explicit training loop (i.e. without the Trainer)

Added a PINOTrainer subclass of the Trainer which helps with printing and logging individual loss weights and loss values.

Added training example scripts for Burgers with the Trainer and without the Trainer

Added some comments on Github in parts of the code that need more scrutiny regarding how we want to deal with the fact that different losses take different inputs (this is more easily and explicitly dealt with when using an explicit training loop, whereas the abstraction of the Trainer makes it harder)

### DIFF
--- a/scripts/train_burgers_pino.py
+++ b/scripts/train_burgers_pino.py
@@ -11,10 +11,6 @@ from neuralop.training import setup, AdamW, PINOTrainer
 from neuralop.utils import get_wandb_api_key, count_model_params, get_project_root
 from neuralop.losses.meta_losses import Relobralo_for_Trainer, SoftAdapt_for_Trainer
 
-#
-# This script uses config.opt.loss_aggregator to select the loss aggregator:
-#   'relobralo' (default) or 'softadapt'
-#
 
 # Read the configuration
 config_name = "default"
@@ -26,7 +22,7 @@ from config.burgers_pino_config import Default
 config = cfg_from_commandline(Default)
 config = config.to_dict()
 
-# Set-up distributed communication, if using
+# Set-up distributed communication
 device, is_logger = setup(config)
 
 # Set up WandB logging
@@ -58,7 +54,7 @@ if config.wandb.log and is_logger:
     wandb.init(**wandb_init_args)
 else: 
     wandb_init_args = None
-# Make sure we only print information when needed
+
 config.verbose = config.verbose and is_logger
 
 # Print config to screen
@@ -91,6 +87,7 @@ optimizer = AdamW(
     weight_decay=config.opt.weight_decay,
 )
 
+# Create the scheduler
 if config.opt.scheduler == "ReduceLROnPlateau":
     scheduler = torch.optim.lr_scheduler.ReduceLROnPlateau(
         optimizer,
@@ -181,7 +178,7 @@ trainer = PINOTrainer(
     wandb_log = config.wandb.log
 )
 
-# Log parameter count
+# Log number of parameters
 if is_logger:
     n_params = count_model_params(model)
 


neuralop/losses/meta_losses.py
```diff
@@ -87,3 +92,427 @@ def __str__(self):
...
+            # Prepare arguments for loss function
+            call_args = {}
+            if 'pred' in sig.parameters or 'y_pred' in sig.parameters:
+                call_args['pred' if 'pred' in sig.parameters else 'y_pred'] = pred_field
+            if 'y' in sig.parameters or 'truth' in sig.parameters:
+                call_args['y' if 'y' in sig.parameters else 'truth'] = truth_field
+            call_args.update(kwargs)
+            
+            # Compute loss
+            field_loss = loss_fn(**call_args)
```
vduruiss, line 163: There should be a better way to do that. This is quite specific to the Burgers equation considered in the example.

This has to do with how we deal with the fact that different losses take different inputs. We could make all the losses more consistent in the names of their arguments, which would make this generalize.

Or we can go for a strategy where every single loss takes all the exact same inputs, even if most of them are unused (which seems to be what was done earlier with the WeightedSumLoss

neuralop/losses/meta_losses.py
```diff
@@ -87,3 +92,427 @@ def __str__(self):
...
```
vduruiss, line 272: Same comment as for Relobralo

There should be a better way to do that. This is quite specific to the Burgers equation considered in the example.

This has to do with how we deal with the fact that different losses take different inputs. We could make all the losses more consistent in the names of their arguments, which would make this generalize.

Or we can go for a strategy where every single loss takes all the exact same inputs, even if most of them are unused (which seems to be what was done earlier with the WeightedSumLoss

dhpitt: What about modifying `PINOTrainer.train_one_batch` so that the call signatures match what you expect/adding index variables or whatever's necessary there?

neuralop/losses/meta_losses.py
```diff
@@ -87,3 +92,427 @@ def __str__(self):
...
+                    losses = {key: weights[key] * losses[key] for key in losses.keys()}
+                    return losses
+
+            return weigh_losses
+
+        self.weigh_losses = weigh_losses_initialize(self.weights)
+
+
+
+class SoftAdapt(Aggregator):
```
dhpitt, line 366: would be helpful to have a simple correctness test for each of these if you can think of one

neuralop/losses/meta_losses.py
```diff
@@ -87,3 +92,427 @@ def __str__(self):
...
+    """
```
dhpitt, line 367: also the param docstring should go here, not in `__init__`

neuralop/losses/meta_losses.py
```diff
@@ -87,3 +92,427 @@ def __str__(self):
...
+
+class Relobralo(Aggregator):
+    """
+    Relative loss balancing with random lookback
+    Reference: "Bischof, R. and Kraus, M., 2021.
+    Multi-Objective Loss Balancing for Physics-Informed Deep Learning.
+    arXiv preprint arXiv:2110.09813."
+    """
+
+    def __init__(
```
dhpitt, line 438: Please add an np docstring here too

neuralop/training/trainer.py
```diff
@@ -790,3 +790,47 @@ def checkpoint(self, save_dir):
                 print(f"[Rank 0]: saved training state to {save_dir}")
 
        
+
+class PINOTrainer(Trainer):
+    """
+    Trainer subclass that logs per-field loss values and weights at the end of each epoch.
+    Works with meta-losses like Relobralo that expose latest_loss_record, latest_weight_record, and latest_total_loss attributes.
+    """
+    def train_one_epoch(self, epoch, train_loader, training_loss):
```
dhpitt, line 799: Like mentioned before - might it be easier to modify `PINOTrainer.train_one_batch` as well/instead of `train_one_epoch`? That way you'd have more control over the loss fn's signature and you wouldn't need to assume a fully abstract loss.

vduruiss:
Made a simpler PR at [PR618](https://github.com/neuraloperator/neuraloperator/pull/618) to add the schemes to the library without the PINOTrainer.

I will come back to the PINOTrainer later on, follow @dhpitt's suggestions, and work on it once we also have sampling/resampling strategies ready.

vduruiss:
@JeanKossaifi  We can close this PR

PR #612, opened 2025-06-12
enable offload

wdlctc:
This PR introduces activation offloading support for Fourier Neural Operator (FNO) training in neuralop, targeting reduced GPU memory consumption during forward and backward passes.

1. New Activation Offloading Module (neuralop/training/offload.py):

Defines enable_activation_offload_for_FNO and supporting functions.

Wraps key forward passes (FNO, FNOBlocks, SpectralConv) with torch.autograd.graph.save_on_cpu(pin_memory=True) for CPU offloading of saved activations during training.

2. Training Script for Offloading (scripts/train_offload.py):

New training script demonstrating end-to-end training with activation offloading.

Compatible with distributed training, WandB logging, and multiresolution datasets (e.g., Darcy Flow).

Preserves model configuration and training loop from standard train.py but optionally enables memory-efficient execution.

### DIFF
--- a/neuralop/training/offload.py
+++ b/neuralop/training/offload.py
@@ -1,0 +1,228 @@
+import math
+import types
+import torch
+
+from typing import List, Optional, Tuple, Union
+
+        
+def enable_activation_offload_for_FNO(FNO):
+    def offload_forward(self, x, output_shape=None, **kwargs):
+        """FNO's forward pass
+        
+        1. Applies optional positional encoding
+    
+        2. Sends inputs through a lifting layer to a high-dimensional latent space
+    
+        3. Applies optional domain padding to high-dimensional intermediate function representation
+    
+        4. Applies `n_layers` Fourier/FNO layers in sequence (SpectralConvolution + skip connections, nonlinearity) 
+    
+        5. If domain padding was applied, domain padding is removed
+    
+        6. Projection of intermediate function representation to the output channels
+    
+        Parameters
+        ----------
+        x : tensor
+            input tensor
+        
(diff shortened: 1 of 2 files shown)


neuralop/training/offload.py
```diff
@@ -0,0 +1,228 @@
...
+                x = self.non_linearity(x)
+    
+            return x
+
+    FNOBlocks.forward_with_postactivation = types.MethodType(forward_with_postactivation, FNOBlocks)
+    for conv in FNOBlocks.convs:
+        enable_activation_offload_for_SpectralConv(conv)
+
+
+def enable_activation_offload_for_SpectralConv(SpectralConv):
```
dhpitt, line 110: I wonder if this might be better as a module in `neuralop.layers`, or as a param to the original `SpectralConv`

neuralop/training/offload.py
```diff
@@ -0,0 +1,228 @@
...
+    
+    def forward(
+        self, x: torch.Tensor, output_shape: Optional[Tuple[int]] = None
+    ):
+        with torch.autograd.graph.save_on_cpu(pin_memory=True):
```
dhpitt, line 115: This looks like the only thing that needs to be changed, right?

dhpitt (commented):
This is really cool @wdlctc , thank you for opening! Since the helper function totally reimplements `SpectralConv.forward` and assumes full access to the method, I think it'll be much more maintainable if it directly interfaces with the forward calls themselves instead of living as helper functions in `neuralop.training`.

PR #614, opened 2025-06-20
Add Multiwavelet Neural Operator

Shixuan01:
This version updates the Multiwavelet Neural Operator, adds mwt.py in the neuralop/models/ directory, and adds mwt_utils.y in neuralop/ as a set of mwt utility functions.

### DIFF
--- a/neuralop/models/mwt.py
+++ b/neuralop/models/mwt.py
@@ -1,0 +1,917 @@
+import torch
+import numpy as np
+import torch.nn as nn
+import torch.nn.functional as F
+from torch import Tensor
+
+
+from typing import List, Tuple, Union, Literal
+import math
+
+from ..mwt_utils import get_filter
+
+class MWT(nn.Module):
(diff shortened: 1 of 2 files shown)


neuralop/models/mwt.py
```diff
@@ -0,0 +1,917 @@
...
+        self.c = c
+        self.n_layers = n_layers
+        self.L = L
+        self.base = base
+        
+        # Determine alpha based on dimensionality
+        if self.n_dim == 1:
+            alpha = n_modes[0]
+            ich_multiplier = c * k
+            MWT_CZ_class = MWT_CZ1d
```
dhpitt, line 109: It would be really preferable if the key operations could be refactored to be dim-agnostic.

neuralop/mwt_utils.py
```diff
@@ -0,0 +1,190 @@
+import torch
```
dhpitt, line 1: These functions should go in a layer/submodule file

PR #617, opened 2025-06-27
Corrected spelling and grammar in FNO theory guide

piano-miles:
Corrected spelling and grammar and improved phrasing in the FNO theory guide documentation. These changes improve readability without altering technical content.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,6 @@
 # weights and biases
-.wandb/
+.wandb/ 
 wandb_api_key.txt
-scripts/wandb/
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
@@ -78,20 +77,5 @@ docs/_build/
 # PyBuilder
 target/
 
-# IPython Notebook
+#Ipython Notebook
 .ipynb_checkpoints
-
-# VSCode
-.vscode
-
-# linting
-.mypy_cache
-
-# Venv
-.venv
-
-# Checkpoints
-checkpoints/
-
-# Logs
-logs/
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 # weights and biases
-.wandb/ 
+.wandb/
 wandb_api_key.txt
+scripts/wandb/
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
@@ -77,5 +78,20 @@ docs/_build/
 # PyBuilder
 target/
 
-#Ipython Notebook
+# IPython Notebook
 .ipynb_checkpoints
+
+# VSCode
+.vscode
+
+# linting
+.mypy_cache
+
+# Venv
+.venv
+
+# Checkpoints
+checkpoints/
+
+# Logs
+logs/
--- a/config/README.md
+++ b/config/README.md
@@ -1,0 +1,1 @@
+- `training_loss`: Loss function used during training (e.g., `"h1"` combines L2 loss and gradient penalty).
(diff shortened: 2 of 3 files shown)


neuralop/models/fno.py
```diff
@@ -156,14 +156,14 @@ class FNO(BaseModel, name='FNO'):
...
-    Li, Z. et al. "Fourier Neural Operator for Parametric Partial Differential 
+    Li, Z. et al. "Fourier Neural Operator for Parametric Partial Differential
         Equations" (2021). ICLR 2021, https://arxiv.org/pdf/2010.08895.
 
     """
 
     def __init__(
         self,
-        n_modes: Tuple[int],
+        n_modes: Tuple[int, ...],
```
JeanKossaifi, line 166: Why did you add the Ellipsis here? I don't think we want to support any other type that tuples of ints?

neuralop/models/fno.py
```diff
@@ -184,21 +184,23 @@ def __init__(
         domain_padding_mode: Literal['symmetric', 'one-sided']="symmetric",
         fno_block_precision: str="full",
         stabilizer: str=None,
-        max_n_modes: Tuple[int]=None,
+        max_n_modes: Tuple[int, ...]=None,
```
JeanKossaifi, line 187: Same here

neuralop/models/fno.py
```diff
@@ -184,21 +184,23 @@ def __init__(
...
         factorization: str=None,
         rank: float=1.0,
         fixed_rank_modes: bool=False,
         implementation: str="factorized",
-        decomposition_kwargs: dict=dict(),
+        decomposition_kwargs: dict=None,
```
JeanKossaifi, line 192: Thank you, this is much better.

JeanKossaifi (commented):
This looks good, thank you @piano-miles! I just have a question about one of the types.

neuralop/models/fno.py
piano-miles: `Tuple[int]` represents a strictly length-1 tuple with an element of type `int`, while `Tuple[int, ...]` represents a tuple of any length with the homogeneous type `int`. This is something I wish Python made more clear. The specifics for type hints are in the original type hint proposal as well as the documentation for the module.

The [PEP 484](https://peps.python.org/pep-0484/) outlines type hints, and states:

> Arbitrary-length homogeneous tuples can be expressed using one type and ellipsis, for example `Tuple[int, ...]`. (The `...` here are part of the syntax, a literal ellipsis.)

This is also echoed in the official documentation for the [typing module](https://docs.python.org/3/library/typing.html):

> To denote a tuple which could be of any length, and in which all elements are of the same type T, use tuple[T, ...]

The documentation also provides this example of "incorrect" usage:

```py
# Error: the type annotation indicates a tuple of length 1,
# but ``z`` has been assigned to a tuple of length 3
z: tuple[int] = (1, 2, 3)
```

neuralop/models/fno.py
piano-miles: Same as explained above ^^^

neuralop/models/fno.py
piano-miles: Note that the lowercase `tuple` syntax was introduced in Python 3.9 via [PEP 585](https://peps.python.org/pep-0585/), and it achieves the same purpose as `typing.Tuple`.

piano-miles: I'll also note that it is a bit different with lists because they are mutable. So `List[int]` would be a `list` of `int`s of any (variable) length. The author might've carried over the syntax for a `list` when using a `tuple`.

PR #618, opened 2025-07-01
PINO Reweighing Schemes

vduruiss:
Adding SoftAdapt / Relobralo and a PINO example

This is a simplified version of [PR611](https://github.com/neuraloperator/neuraloperator/pull/611) where the schemes are used without the main Trainer.

I will come back to building the PINOTrainer later on based on @dhpitt's good suggestions, once we also have sampling/resampling strategies ready. 

I addressed some of the other comments made in [PR611](https://github.com/neuraloperator/neuraloperator/pull/611) regarding docstrings.


config/wandb.py
```diff
@@ -26,10 +26,10 @@ class WandbConfig(ConfigBase):
         to W&B, if logging to W&B (``log = True``)
 
     """
-    log: bool = False
+    log: bool = True
     entity: Optional[str] = None
-    project: Optional[str] = None
+    project: Optional[str] = "Burgers-Pino"
```
dhpitt: I wouldn't change this default

dhpitt:
Thanks @vduruiss , this is an excellent addition to the library.

PR #625, opened 2025-07-23
Add grf generation with RBF kernel and Matern kernel

i207M:


i207M:
Thanks @JeanKossaifi for reviewing! Yeah I do plan to add some documentation for it. Please allow some time.

Issue #623, opened 2025-07-18
Dependencies (torch_harmonics) error when installing from source


niemiaszek:
I tried installing from source, as I had an issue with Darcy-Flow dataset download on pip version. I was hit with error with import from `torch_harmonics`, when simply `import neuralop`.
```
File "/home/<>/.pixi/envs/default/lib/python3.13/site-packages/neuralop/layers/discrete_continuous_convolution.py", line 13, in <module>
    from torch_harmonics.convolution import _compute_support_vals_isotropic, _compute_support_vals_anisotropic
ImportError: cannot import name '_compute_support_vals_isotropic' from 'torch_harmonics.convolution' (/home/<>/.pixi/envs/default/lib/python3.13/site-packages/torch_harmonics/convolution.py)
```
`torch_harmonics` wasn't installed at first with requirements. When installed, it still wasn't working with above issue. l looked it up and it looks like `troch_harmonics` removed the imported function with 0.7.4 version, which is recommended in requirements of neuraloperator extras.

Installing 0.7.3 fixes the issue.

PR #627, opened 2025-08-01
Update torch-harmonics dependency to use specific GitHub repository U…

chrisvoncsefalvay:
…RL to fix torch_harmonics import related errors with filter_basis.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,7 +4,7 @@ configmypy
 zencfg
 tensorly
 tensorly-torch
-torch-harmonics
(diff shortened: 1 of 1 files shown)


chrisvoncsefalvay:
Fixes #623

JeanKossaifi:
@chrisvoncsefalvay - do you still have an issue with this or should we close?

chrisvoncsefalvay:
I think I've rerun this a few times now and the issue seems to have fixed itself so we can close this as a 'won't do'! Thanks!

PR #630, opened 2025-08-03
Remove one-sided padding and Fourier continuation

vduruiss:


JeanKossaifi:
Looks great, thank you!

PR #632, opened 2025-08-05
ZenCFG Update + Small config fixes

vduruiss:
Updating the zencfg function names following the recent commit https://github.com/JeanKossaifi/zencfg/commit/62d7f1aba9b0d55b004303f431855e0d73d45eea in the ZenCFG library 

Please @JeanKossaifi check that this would not break anything else

### DIFF
--- a/neuralop/data/datasets/car_cfd_dataset.py
+++ b/neuralop/data/datasets/car_cfd_dataset.py
@@ -77,7 +77,7 @@ class CarCFDDataset(MeshDataModule):
                                         root=root_dir)
         super().__init__(
             root_dir=root_dir,
-            item_dir_name='',
+            item_dir_name='data/',
             n_train=n_train,
             n_test=n_test,
             query_res=query_res,


scripts/train_burgers.py
```diff
@@ -14,7 +14,7 @@
 # Read the configuration
 config_name = "default"
 # Read the configuration
-from zencfg import cfg_from_commandline, cfg_from_nested_dict
```
JeanKossaifi, line 17: Let's just use make_config_from_cli instead of the old name

neuralop/tests/test_model_from_config.py
```diff
@@ -9,15 +9,15 @@
...
-from zencfg import cfg_from_flat_dict
+from zencfg.from_dict import make_config_from_flat_dict
 import sys 
 from .test_config import TestConfig
 
 
 def test_from_config():
     """Test forward/backward from a config file"""
     # Read the configuration
-    config = cfg_from_flat_dict(TestConfig, {})
```
JeanKossaifi, line 20: We can just use make_config(TestConfig)

JeanKossaifi (commented):
I left some comments, thanks @vduruiss !

PR #635, opened 2025-08-06
A documentation on how to apply cpu offloading

cwwangcal:
A documentation on how to apply cpu offloading


scripts/cpuoffload.md
dhpitt, line 1: Looks like there is a lot redundant between this and the new docs. I would delete this

dhpitt (approved):
This is a very helpful tip, thanks @cwwangcal ! Approved once we remove the `.md` file in scripts.

PR #636, opened 2025-08-07
Add Multiwavelet Neural Operator

shixuanli01:
Thank you very much for your previous reply and suggestions, including refactoring the layers to be dimension-agnostic and follow the modular design of the FNO currently in the library. In this version, we have integrated the previous mwt_utils.py into mwt.py (if it is not very suitable, we can also integrate it into the layer/submodule), and integrated the multidimensional MWT. All updates are now concentrated in /neuraloperator/neuralop/models/mwt.py, which hopefully makes the structure more concise! We will also update test_mwt.py and related explanatory files later, and hope they will be helpful!


neuralop/models/mwt.py
```diff
@@ -0,0 +1,666 @@
...
+from torch import Tensor
+from typing import List, Tuple, Union, Literal
+import math
+
+from functools import partial
+from scipy.special import eval_legendre
+from sympy import Poly, legendre, Symbol, chebyshevt
+
+class WaveletUtils:
+    """Unified wavelet utility class containing wavelet transform functions for all dimensions"""
```
dhpitt: Some docs here would be helpful. Please refer to the contributors guide for guidance on how to format in-line docstrings

neuralop/models/mwt.py
```diff
@@ -0,0 +1,666 @@
...
+            x = x.permute(0, 2, 3, 1)
+        elif self.n_dim == 3:
+            x = x.permute(0, 2, 3, 4, 1)
+        
+        x = self.output_proj(x)
+        x = x.view(original_shape)
+        return x
+
+
+class SparseKernelFT(nn.Module):
```
dhpitt: These layers belong in `neuralop/layers`

neuralop/models/mwt.py
```diff
@@ -0,0 +1,666 @@
...
+            x = torch.fft.irfftn(out_ft, s=(Nx, Ny, T))
+            x = x.permute(0, 2, 3, 4, 1)
+            x = F.relu(x)
+            x = self.output_proj(x)
+            x = x.view(original_shape)
+        
+        return x
+
+
+class MWT_CZ(nn.Module):
```
dhpitt: same comments re: docstring, and moving layers to the layers module

neuralop/models/mwt.py
```diff
@@ -0,0 +1,666 @@
...
+        
+        for i in range(ns - 1 - self.L, -1, -1):
+            x = x + Us[i]
+            x = torch.cat((x, Ud[i]), -1)
+            x = self.even_odd_reconstruction(x)
+        
+        return x
+
+
+class MWT(nn.Module):
```
dhpitt: This is great. Let's think of a more descriptive name than MWT, perhaps MultiWaveletNO or MWNO?

neuralop/models/mwt.py
```diff
@@ -0,0 +1,666 @@
...
+        elif self.n_dim == 3:
+            x = x.view(B, Nx, Ny, T, -1)
+        
+        x = self.projection(x)
+        
+        if self.out_channels == 1:
+            return x.squeeze(-1)
+        return x
+
+"""Compatible MWT Model"""
```
dhpitt: We don't necessarily need these classes

dhpitt (commented):
Thank you so much for your improved PR @Shixuan01 . This is great. I have a few stylistic notes to make this library-ready:
1. Wherever possible, please add unit tests (e.g. for models, in `neuralop/models/tests/test_xx.py` to ensure that data passes through models in the way that we expect. Where possible, if you can test the correctness of sub-operations like the transform or the basis, these will make great tests to prevent regressions when we change the code. Check `neuralop/layers/tests` for examples. 
2. All functions and classes should have Numpy-formatted docstrings, detailed [here](https://neuraloperator.github.io/dev/dev_guide/documentation.html) in our dev guide. This will make it easier for users to understand and for future maintainers to learn and maintain. 
3. The individual layers and utils belong in `neuralop/layers` to indicate that they are submodules that can be composed into an MWT model. 
4. Let's think of a slightly more descriptive name for the model that indicates that it's a full operator, perhaps MWNO?

Overall this is very solid and will make a great contribution to the library. Thank you!

neuralop/models/mwno.py
```diff
@@ -0,0 +1,200 @@
...
+
+from functools import partial
+from scipy.special import eval_legendre
+from sympy import Poly, legendre, Symbol, chebyshevt
+
+
+from ..layers.mwno_block import MWNO_CZ
+
+class MWNO(nn.Module):
+    """
```
dhpitt: Nit: could you add a reference to the MWNO paper and a short description of the model?

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,610 @@
...
+            x = self.output_proj(x)
+            x = x.view(original_shape)
+        
+        return x
+
+
+
+class MWNO_CZ(nn.Module):
+    """
+    Dimension-agnostic MWNO core layer
```
dhpitt: nit: a short desc + reference would also be helpful here

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,610 @@
...
+    """
+
+    def __init__(self, k=3, alpha=5, L=0, c=1, base='legendre', n_dim=1, initializer=None, **kwargs):
```
dhpitt, line 1230: Could you add a docstring w/arg meanings ?

neuralop/models/mwno.py
```diff
@@ -0,0 +1,200 @@
...
+        Returns
+        -------
+        output : tensor
+            Output tensor with shape:
+            - 1D: (B, N, out_channels) or (B, N) if out_channels=1
+            - 2D: (B, Nx, Ny, out_channels) or (B, Nx, Ny) if out_channels=1
+            - 3D: (B, Nx, Ny, T, out_channels) or (B, Nx, Ny, T) if out_channels=1
+        """
+        
+        x = self.lifting(x)
```
dhpitt, line 444: nit: Some of this dimension handling logic could be cleaned up / wrapped in `assert`s so that errors are a little more informative. e.g. throw an error if the shapes/dims are wrong

neuralop/models/mwt.py
```diff
@@ -0,0 +1,666 @@
...
+            x = F.relu(x)
+            x = self.output_proj(x)
+            x = x.view(original_shape)
+        
+        return x
+
+
+class MWT_CZ(nn.Module):
+    """Dimension-agnostic MWT core layer"""
+
```
dhpitt: Another small note; this will work better as a layer

dhpitt (commented):
Looks really good, thank you so much for the improvements. Could you please:

1. respond to the small nits re: docstrings
2. refactor `neuralop.models.mwt` --> `neuralop.layers.mwt`
3. add corresponding unit tests for the MWT and mwno block
Once these changes are in this will be ready to go.

neuralop/models/mwt.py
shixuanli01: Thanks for mentioning this, it was a redundant module and we have removed it.

dhpitt (commented):
These changes look good to me, thank you @Shixuan01 for responding to the feedback!

dhpitt (approved):

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,701 @@
...
+        This function evaluates a polynomial defined by coefficients phi_c,
+        but sets the value to zero outside the interval [lb, ub].
+        
+        Parameters
+        ----------
+        phi_c : array_like
+            Polynomial coefficients
+        x : array_like
+            Points at which to evaluate the polynomial
+        lb : float, optional
```
JeanKossaifi: could we just call them lower_bound and upper_bound or min and max for clarity?

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,701 @@
...
+    def phi_(phi_c, x, lb=0, ub=1):
+        """
+        Evaluate a polynomial basis function with compact support.
+        
+        This function evaluates a polynomial defined by coefficients phi_c,
+        but sets the value to zero outside the interval [lb, ub].
+        
+        Parameters
+        ----------
+        phi_c : array_like
```
JeanKossaifi: What about just `coefs_`?

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,701 @@
...
+        x : array_like
+            Points at which to evaluate the polynomial
+        lb : float, optional
+            Lower bound of the support interval, default 0
+        ub : float, optional
+            Upper bound of the support interval, default 1
+            
+        Returns
+        -------
+        array_like
```
JeanKossaifi: Would be good to specify the shape (e.g. `x.shape[0]`, ...)

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,701 @@
...
+        Evaluate a polynomial basis function with compact support.
+        
+        This function evaluates a polynomial defined by coefficients phi_c,
+        but sets the value to zero outside the interval [lb, ub].
+        
+        Parameters
+        ----------
+        phi_c : array_like
+            Polynomial coefficients
+        x : array_like
```
JeanKossaifi, line 64: Can we specify the shape depending on dim?

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,701 @@
...
+        -----
+        For Legendre basis:
+        - Scaling functions are normalized Legendre polynomials on [0,1]
+        - Wavelets are constructed via orthogonalization against scaling functions
+        
+        For Chebyshev basis:
+        - Scaling functions are normalized Chebyshev polynomials
+        - Wavelets have compact support on [0,0.5] and [0.5,1] respectively
+        """
+        x = Symbol('x')
```
JeanKossaifi: Do we need to go through Sympy for this? Couldn't we directly return the discretized evaluations, instead of generating a symbolic version and then evaluating?

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,701 @@
...
+                    PHI0[ki, kpi] = (wm * phi[ki](2*x_m) * phi[kpi](2*x_m)).sum() * 2
+                    PHI1[ki, kpi] = (wm * phi[ki](2*x_m-1) * phi[kpi](2*x_m-1)).sum() * 2
+                    
+            PHI0[np.abs(PHI0)<1e-8] = 0
+            PHI1[np.abs(PHI1)<1e-8] = 0
+
+        for matrix in [H0, H1, G0, G1]:
+            matrix[np.abs(matrix)<1e-8] = 0
+            
+        return H0, H1, G0, G1, PHI0, PHI1
```
JeanKossaifi: The function and docstring are a little hard to follow - could we add a more intuitive explanation in the doctoring and use more descriptive name for the variables (we also typically name variables e.g. `h_0` and reserve Capital Case for classes.

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,701 @@
...
+            x = x.view(B, Nx, Ny, T, -1).permute(0, 4, 1, 2, 3)
+        
+        x = self.conv(x)
+        
+        if self.n_dim == 1:
+            x = x.permute(0, 2, 1)
+        elif self.n_dim == 2:
+            x = x.permute(0, 2, 3, 1)
+        elif self.n_dim == 3:
+            x = x.permute(0, 2, 3, 4, 1)
```
JeanKossaifi, line 628: minor: is there a way to avoid the permuting back and forth?

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,701 @@
...
+            out_ft[:, :, -l1:, -l2:, :l3] = torch.einsum("bixyz,ioxyz->boxyz",
+                x_fft[:, :, -l1:, -l2:, :l3], self.weights4[:, :, :l1, :l2, :l3])
+            
+            x = torch.fft.irfftn(out_ft, s=(Nx, Ny, T))
+            x = x.permute(0, 2, 3, 4, 1)
+            x = F.relu(x)
+            x = self.output_proj(x)
+            x = x.view(original_shape)
+        
+        return x
```
JeanKossaifi, line 855: This doesn't have a doctoring and hard to follow currently - could we reuse the existing spectral conv by factoring out the reshapes? Don't we need a rfftshift to index like this as well?

JeanKossaifi (commented):
This is a great addition to the library @Shixuan01, thank you for the PR!

My main comments are on clarify and readability of the code: right now it is hard for a maintainer/user to directly use the code and/or modify it. It would be great to clarify it by using descriptive names and adding more docstrings and comments!

neuralop/layers/mwno_block.py
shixuanli01: Thanks for the suggestion, we have made the changes.

neuralop/layers/mwno_block.py
shixuanli01: Yes, we use numpy instead of Sympy to make the calculation faster.

neuralop/layers/mwno_block.py
shixuanli01: Thanks for pointing this out, we've unified our coding style and variables and expanded the documentation a lot to explain the code.

neuralop/layers/mwno_block.py
shixuanli01: Thanks for pointing this out. We have reflected this in the new version of the code and modified the documentation and unnecessary operations.

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,1551 @@
...
+    def forward(self, x):
+        """
+        Apply multiwavelet transform: decompose → transform → reconstruct.
+
+        This is the main processing pipeline that:
+        1. Decomposes input into multiple scales (wavelet pyramid)
+        2. Applies learnable transformations at each scale
+        3. Reconstructs output from all scales
+
+        Processing flow:
```
dhpitt, line 1520: These are extremely helpful docstrings. I would suggest putting them down beneath the params heading so that the parameters section renders first in IDEs

neuralop/models/mwno.py
```diff
@@ -0,0 +1,621 @@
...
+    >>> y = model_3d(x)  # (8, 32, 32, 20) - channel squeezed
+
+    Notes
+    -----
+    - Spatial dimensions must be powers of 2 for wavelet decomposition
+    - For 3D, only first two dimensions are decomposed (time preserved)
+    - ReLU activation applied between MWNO layers (except after last layer)
+    - Output channel dimension auto-squeezed if out_channels=1
+
+    References
```
dhpitt, line 107: nit: can you use the RST formatted heading as in the FNO's model file?

neuralop/layers/mwno_block.py
shixuanli01: Thank you! I have moved the place for those senteces.

neuralop/models/mwno.py
shixuanli01: Yes, I have standardized our format.

vduruiss:
Thank you @Shixuan01 for your hard work on this PR!  I'm starting my review now

neuralop/models/tests/test_mwno.py
```diff
@@ -0,0 +1,86 @@
...
+from ..mwno import MWNO
+
+
+@pytest.mark.parametrize("n_dim", [1, 2, 3])
+@pytest.mark.parametrize("alpha", [3, 5, 8])
+@pytest.mark.parametrize("k", [3, 4])
+@pytest.mark.parametrize("c", [1, 2])
+@pytest.mark.parametrize("n_layers", [2, 3])
+@pytest.mark.parametrize("L", [0, 1])
+@pytest.mark.parametrize("base", ["legendre"])
```
vduruiss: Could we split these tests? 

It is a very comprehensive test function, but it is currently running 144 times because of all the possible configurations. This can significantly slow down the time it takes to deploy the library and run the pytests.

Maybe we can split it in 2 or 3 test functions that explore varying all the arguments but without testing every single possible configuration?

neuralop/models/__init__.py
```diff
@@ -10,5 +10,6 @@
 from .uqno import UQNO
 from .fnogno import FNOGNO
 from .gino import GINO
-from .codano import CODANO
+from .mwno import MWNO
+
 from .base_model import get_model
```
vduruiss: Could you rebase to the latest version of the library? 

Some of these models are not in the library anymore (FNO1d, FNO2d, FNO3d, TFNO1d, TFNO2d, TFNO3d)

neuralop/models/mwno.py
```diff
@@ -0,0 +1,580 @@
...
+    Other parameters
+    ----------------
+    alpha : int, optional
+        Alternative way to specify modes (all dimensions use same value).
+        Only used if n_modes is not provided.
+        Example: alpha=12 with n_dim=2 → modes=(12, 12).
+    n_dim : int, optional
+        Spatial dimensionality (1, 2, or 3).
+        Only needed if using alpha parameter.
+        Inferred from n_modes if n_modes is a tuple.
```
vduruiss: Maybe we can remove these since they don't seem to provide more functionality than what n_modes does?

neuralop/models/mwno.py
```diff
@@ -0,0 +1,580 @@
...
+    title={Coupled Multiwavelet Operator Learning for Coupled Differential Equations},
+    author={Xiongye Xiao and Defu Cao and Ruochen Yang and Gaurav Gupta and Gengshuo Liu and Chenzhong Yin and Radu Balan and Paul Bogdan},
+    booktitle={The Eleventh International Conference on Learning Representations },
+    year={2023},
+    url={https://openreview.net/forum?id=kIo_C6QmMOM}
+
+    """
+
+    # Default hidden dimension when projection_channels=0
+    DEFAULT_PROJECTION_HIDDEN = 128
```
vduruiss: We can probably remove those since the `projection_channels` is set to 128 by default

neuralop/models/mwno.py
```diff
@@ -0,0 +1,580 @@
...
+
+        Returns
+        -------
+        nn.Module
+            Projection layer (two-layer MLP with ReLU)
+        """
+        if projection_channels > 0:
+            hidden_dim = projection_channels
+        else:
+            hidden_dim = self.DEFAULT_PROJECTION_HIDDEN
```
vduruiss: We can simply set it to `projection_channels` if we remove `self.DEFAULT_PROJECTION_HIDDEN`

neuralop/models/mwno.py
```diff
@@ -0,0 +1,580 @@
...
+        alpha : int, optional
+            Uniform modes for all dimensions
+        n_dim : int, optional
+            Number of spatial dimensions
+
+        Returns
+        -------
+        n_modes : Tuple[int, ...]
+            Parsed modes per dimension
+        n_dim : int
```
vduruiss: We can probably simplify significantly this whole parsing function if we remove the alpha+ndim for specifying the modes

neuralop/models/mwno.py
```diff
@@ -0,0 +1,580 @@
...
+            Input feature dimension
+        channel_multiplier : int
+            Output dimension (c * k^n_dim)
+        lifting_channels : int
+            Hidden dimension (0 for direct linear projection)
+
+        Returns
+        -------
+        nn.Module
+            Lifting layer (Linear or Sequential MLP)
```
vduruiss, line 245: Could we use the existing `ChannelMLP` class [(link)](https://github.com/neuraloperator/neuraloperator/blob/main/neuralop/layers/channel_mlp.py) ?

You can then create the lifting layer as in the init of `FNO` class

```
self.lifting = ChannelMLP(
            in_channels=lifting_in_channels,
            out_channels=self.hidden_channels,
            hidden_channels=self.lifting_channels,
            n_layers=2,
            n_dim=self.n_dim,
            non_linearity=non_linearity,
        )
        if self.complex_data:
            self.lifting = ComplexValued(self.lifting)
```

neuralop/models/mwno.py
```diff
@@ -0,0 +1,580 @@
...
+            Input dimension (c * k^n_dim)
+        out_channels : int
+            Output feature dimension
+        projection_channels : int
+            Hidden dimension (0 uses default)
+
+        Returns
+        -------
+        nn.Module
+            Projection layer (two-layer MLP with ReLU)
```
vduruiss: Could we use the existing `ChannelMLP` class [(link)](https://github.com/neuraloperator/neuraloperator/blob/main/neuralop/layers/channel_mlp.py) ?

You can then create the projection layer as in the init of `FNO` class

```
## Projection layer
        self.projection = ChannelMLP(
            in_channels=self.hidden_channels,
            out_channels=out_channels,
            hidden_channels=self.projection_channels,
            n_layers=2,
            n_dim=self.n_dim,
            non_linearity=non_linearity,
        )
        if self.complex_data:
            self.projection = ComplexValued(self.projection)

```

neuralop/models/mwno.py
```diff
@@ -0,0 +1,580 @@
...
+            out_channels: int = 1,
+            k: int = 3,
+            c: int = 1,
+            n_layers: int = 3,
+            L: int = 0,
+            lifting_channels: int = 0,
+            projection_channels: int = 128,
+            base: str = 'legendre',
+            initializer: Optional[Callable] = None,
+            **kwargs
```
vduruiss: We can remove the **kwargs for the model initialization, to make sure users don't pass in arguments that do nothing and expect different results

neuralop/models/mwno.py
vduruiss: We can essentially re-use the code from the `init` of the `FNO` class for the mode specification

neuralop/models/mwno.py
```diff
@@ -0,0 +1,580 @@
...
+
+        elif self.n_dim == 2:
+            batch_size, height, width = x.shape[0], x.shape[1], x.shape[2]
+            return x.view(batch_size, height, width, -1)
+
+        elif self.n_dim == 3:
+            batch_size, height, width, time_steps = (
+                x.shape[0], x.shape[1], x.shape[2], x.shape[3]
+            )
+            return x.view(batch_size, height, width, time_steps, -1)
```
vduruiss: I think we might be able to replace all of this by 

`return x.flatten(-2, -1)`

neuralop/models/mwno.py
```diff
@@ -0,0 +1,580 @@
...
+            return x.view(batch_size, height, width, self.c, self.k ** 2)
+
+        elif self.n_dim == 3:
+            batch_size, height, width, time_steps, lifted_channels = x.shape
+            expected_channels = self.c * self.k ** 2
+            assert lifted_channels == expected_channels, (
+                f"Lifting produced {lifted_channels} channels, "
+                f"expected {expected_channels} (c={self.c} × k²={self.k ** 2})"
+            )
+            return x.view(batch_size, height, width, time_steps, self.c, self.k ** 2)
```
vduruiss: I think this could probably be replaced by

```
    # basis: k for 1D, k**2 for 2D/3D
    basis = self.k if self.n_dim == 1 else self.k ** 2
    expected = self.c * basis
    assert x.shape[-1] == expected, (
        f"Lifting produced {x.shape[-1]} channels, expected {expected} "
        f"(c={self.c} × basis={basis})"
    )
    # split last dim into (c, basis)
    return x.unflatten(-1, (self.c, basis))
```

neuralop/models/mwno.py
```diff
@@ -0,0 +1,580 @@
...
+            - 3D: (batch, height, width, time, out_channels) or (batch, height, width, time) if out_channels=1
+
+        Examples
+        --------
+        >>> model = MWNO(alpha=12, n_dim=2, in_channels=1, out_channels=1)
+        >>> x = torch.randn(16, 64, 64, 1)
+        >>> y = model(x)  # shape: (16, 64, 64) - channel squeezed
+        """
+        # Validate input
+        self._validate_input_shape(x)
```
vduruiss: I think we should find an alternative to calling this for every single forward pass. This is definitely helpful for debugging and making sure the input is passed properly, but this will get called thousands or millions of times during training unnecessarily

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,1553 @@
...
+
+        How it works:
+        1. Start with orthogonal polynomials (Legendre or Chebyshev) on [0,1]
+        2. These become our "scaling functions" (phi) - they capture smooth, low-frequency content
+        3. Create "wavelet functions" (psi) through orthogonalization - they capture details/edges
+        4. Wavelets are split into two halves: psi1 for [0, 0.5] and psi2 for [0.5, 1]
+
+        The result is a complete basis where:
+        - phi functions = "blur" or "smooth approximation" at a given scale
+        - psi functions = "details" or "differences" between scales
```
vduruiss: Can we move this to the top of the docstring for the function?

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,1553 @@
...
+
+    Use case: Apply spatial mixing to wavelet detail coefficients, capturing
+    local patterns and textures in the spatial domain.
+
+    The layer automatically handles 1D, 2D, and 3D inputs by:
+    1. Reshaping input to (batch, channels, *spatial_dims)
+    2. Applying appropriate Conv{1,2,3}d layer
+    3. Projecting back to original channel structure
+    4. Reshaping to original format
+
```
vduruiss: Can we move this to the top of the docstring for the class?

neuralop/layers/mwno_block.py
vduruiss: Let's move the description back to the top of the docstring, I think it's what we have been doing more consistently across the library. Sorry for the back and forth

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,1553 @@
...
+        if self.n_dim == 1:
+            batch_size, n_points, num_channels, input_channels = x.shape
+            num_scales = math.floor(np.log2(n_points))
+        elif self.n_dim == 2:
+            batch_size, nx, ny, num_channels, input_channels = x.shape
+            num_scales = math.floor(np.log2(nx))
+        elif self.n_dim == 3:
+            batch_size, nx, ny, time_steps, num_channels, input_channels = x.shape
+            num_scales = math.floor(np.log2(nx))
+
```
vduruiss, line 1554: Does this mean the number of scales is always determined by the resolution in the first dimension? 
Does this lead to issues when dealing with cases where the dimensions have significantly different resolutions?

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,1553 @@
...
+            # (B, Nx, Ny, c, k²) → (B, c*k², Nx, Ny)
+            batch_size, nx, ny, num_channels, k_sq = x.shape
+            x = x.view(batch_size, nx, ny, -1)  # (B, Nx, Ny, c*k²)
+            x = x.permute(0, 3, 1, 2)  # (B, c*k², Nx, Ny)
+
+        elif self.n_dim == 3:
+            # (B, Nx, Ny, T, c, k²) → (B, c*k², Nx, Ny, T)
+            batch_size, nx, ny, time_steps, num_channels, k_sq = x.shape
+            x = x.view(batch_size, nx, ny, time_steps, -1)  # (B, Nx, Ny, T, c*k²)
+            x = x.permute(0, 4, 1, 2, 3)  # (B, c*k², Nx, Ny, T)
```
vduruiss: I think we can update with this
```

        if self.n_dim == 1:
            # (B, N, c, k) -> (B, c*k, N)
            batch_size, n_points, num_channels, k = x.shape
            # move (c,k) next to batch: (B, c, k, N), then merge c and k
            x = x.permute(0, 2, 3, 1).contiguous().reshape(batch_size, num_channels * k, n_points)

        elif self.n_dim == 2:
            # (B, Nx, Ny, c, k²) -> (B, c*k², Nx, Ny)
            batch_size, nx, ny, num_channels, k_sq = x.shape
            x = x.permute(0, 3, 4, 1, 2).contiguous().reshape(batch_size, num_channels * k_sq, nx, ny)

        elif self.n_dim == 3:
            # (B, Nx, Ny, T, c, k²) -> (B, c*k², Nx, Ny, T)
            batch_size, nx, ny, time_steps, num_channels, k_sq = x.shape
            x = x.permute(0, 4, 5, 1, 2, 3).contiguous().reshape(batch_size, num_channels * k_sq, nx, ny, time_steps)

```

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,1553 @@
...
+        x = self.conv(x)
+
+        # Reshape back
+        x = self._reshape_from_conv(x, original_shape)
+
+        # Project to original channels
+        x = self.output_proj(x)
+
+        # Restore exact original shape
+        x = x.view(original_shape)
```
vduruiss: Can replace with
`x = x.reshape(original_shape)`

`reshape` is usually safer than `view`

neuralop/models/mwno.py
```diff
@@ -0,0 +1,580 @@
...
+
+        # Build network layers
+        self.lifting = self._build_lifting_layer(
+            in_channels, channel_multiplier, lifting_channels
+        )
+
+        self.mwno_layers = nn.ModuleList([
+            MWNO_CZ(
+                k=k,
+                alpha=alpha,
```
vduruiss: It seems that the current code only allows to use the same number of modes for every dimension. Is there a real constraint leading to this design? Otherwise it would be good to have the flexibility of choosing more or less modes along certain dimensions, since dynamics can have different scales along different dimensions and the resolution could also be different along different dimensions

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,1553 @@
...
+
+        elif self.n_dim == 2:
+            # (B, Nx, Ny, c, k²) → (B, c*k², Nx, Ny)
+            batch_size, nx, ny, num_channels, k_sq = x.shape
+            x = x.view(batch_size, nx, ny, -1).permute(0, 3, 1, 2)
+
+        elif self.n_dim == 3:
+            # (B, Nx, Ny, T, c, k²) → (B, c*k², Nx, Ny, T)
+            batch_size, nx, ny, time_steps, num_channels, k_sq = x.shape
+            x = x.view(batch_size, nx, ny, time_steps, -1).permute(0, 4, 1, 2, 3)
```
vduruiss: Can probably replace with 

  ```
        if self.n_dim == 1:
            # (B, N, c, k) -> (B, c*k, N)
            batch_size, n_points, num_channels, k = x.shape
            # (B, N, c, k) -> (B, c, k, N) -> (B, c*k, N)
            x = x.permute(0, 2, 3, 1).contiguous().reshape(batch_size, num_channels * k, n_points)
        elif self.n_dim == 2:
            # (B, Nx, Ny, c, k²) -> (B, c*k², Nx, Ny)
            batch_size, nx, ny, num_channels, k_sq = x.shape
            # (B, Nx, Ny, c, k²) -> (B, c, k², Nx, Ny) -> (B, c*k², Nx, Ny)
            x = x.permute(0, 3, 4, 1, 2).contiguous().reshape(batch_size, num_channels * k_sq, nx, ny)

        elif self.n_dim == 3:
            # (B, Nx, Ny, T, c, k²) -> (B, c*k², Nx, Ny, T)
            batch_size, nx, ny, time_steps, num_channels, k_sq = x.shape
            # (B, Nx, Ny, T, c, k²) -> (B, c, k², Nx, Ny, T) -> (B, c*k², Nx, Ny, T)
            x = x.permute(0, 4, 5, 1, 2, 3).contiguous().reshape(batch_size, num_channels * k_sq, nx, ny, time_steps)
```

File: neuralop/layers/mwno_block.py [hunk 40]
```diff
@@ -0,0 +1,1553 @@
...
+        return x
+
+
+class SparseKernelFT(nn.Module):
+    """
+    Sparse Fourier Transform Kernel for Multiwavelet Networks.
+
+    This layer operates in the frequency domain, applying learned transformations
+    to a sparse subset of Fourier modes. This is computationally efficient and
+    particularly effective for capturing global patterns.
```
vduruiss, line 670: This whole class seems to be doing almost exactly the same thing as the `SpectralConv` class ([link](https://github.com/neuraloperator/neuraloperator/blob/main/neuralop/layers/spherical_convolution.py))

Could we leverage the existing `SpectralConv` class instead so we don't have duplicate codes written in different ways?

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,1553 @@
...
+    >>> layer_3d = MWNO_CZ(k=3, alpha=5, L=0, c=1, n_dim=3)
+    >>> field = torch.randn(8, 32, 32, 20, 1, 9)
+    >>> output = layer_3d(field)
+    >>> assert output.shape == field.shape
+
+    Multiwavelet Neural Operator Core Z-transform Layer.
+
+    This layer is the heart of the Multiwavelet Neural Operator (MWNO) architecture.
+    It works like a hierarchical image pyramid, but with learnable transformations
+    at each level.
```
vduruiss: Can we move this to the top of the docstring?

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,1553 @@
...
+        # Reshape and apply nonlinearity + projection
+        x = self._reshape_from_fft(x, original_shape)
+        x = F.relu(x)
+        x = self.output_proj(x)
+        x = x.view(original_shape)
+
+        return x
+
+
+class MWNO_CZ(nn.Module):
```
vduruiss: Maybe we can rename this `MWNO_Block` or `MWNO_CZ_Block`, more similar to have we have been calling other layers in the library

neuralop/models/mwno.py
```diff
@@ -0,0 +1,580 @@
...
+        For scalar fields: in_channels=1. For vector fields: in_channels=n_components.
+    out_channels : int
+        Number of output function channels. Determined by the problem.
+        Set to 1 for scalar outputs (will auto-squeeze).
+    k : int, optional
+        Wavelet basis size (number of polynomial basis functions). Default: 3
+        k=2: Piecewise linear wavelets, k=3: Cubic wavelets, k=4: Quartic wavelets.
+    c : int, optional
+        Number of parallel wavelet channels. Default: 1
+        Increases model capacity: total wavelet features = c * k^n_dim.
```
vduruiss, line 41: Can you specify what are typical ranges you would recommend for this parameter?

neuralop/models/tests/test_mwno.py
shixuanli01: Thank you for your suggestion; we have simplified the test cases.

neuralop/models/mwno.py
shixuanli01: Thank you!

neuralop/models/mwno.py
shixuanli01: Thank you! Accepted.

neuralop/layers/mwno_block.py
shixuanli01: After reviewing the code, we found some differences in the details between the two code snippets, so we decided to keep the previous design.

neuralop/models/mwno.py
shixuanli01: Thank you for your question, but currently, the code only allows the same number of modes for every dimension. We will improve this feature later.

neuralop/layers/mwno_block.py
shixuanli01: We have modified our block name. Thank you for your advice!

neuralop/models/mwno.py
vduruiss: These (alpha and n_dim) are not in the `__init__ `anymore. If that's intentional, we can remove them from the doctoring so users are not confused

Let's make sure the docstring reflects only the parameters that the user can specify in the `__init__ `

neuralop/models/mwno.py
```diff
@@ -0,0 +1,461 @@
...
+        initializer : callable
+            Initialization function with signature: initializer(weight) -> None
+        """
+        for module in self.modules():
+            if isinstance(module, nn.Linear):
+                initializer(module.weight)
+
+    def _validate_input_shape(self, x: torch.Tensor) -> None:
+        """
+        Validate input tensor shape and dimensions.
```
vduruiss: I don't think this function is used anywhere

neuralop/models/mwno.py
```diff
@@ -0,0 +1,461 @@
...
+        Number of coarsest decomposition levels to skip. Default: 0
+        Reduces computation by stopping wavelet decomposition early.
+        L=0: Full decomposition to coarsest scale. L=1: Stop 1 level before coarsest.
+    lifting_channels : int, optional
+        Hidden dimension for lifting layer. Default: 0
+        If 0: Direct linear lifting (fast).
+        If > 0: Two-layer MLP (in → lifting_channels → wavelet_space).
+    projection_channels : int, optional
+        Hidden dimension for projection layer. Default: 128
+        If 0: Uses default two-layer MLP with hidden_dim=128.
```
vduruiss: The `projection_channels=0` option does not seem to be implemented. Maybe we can remove that option? Or simply use the ChannelMLP already in the library

neuralop/models/mwno.py
```diff
@@ -0,0 +1,461 @@
...
+    fluctuations and supports accurate modeling of complex dynamical systems governed
+    by partial differential equations.
+
+    Main Parameters
+    ---------------
+    n_modes : Tuple[int, ...] or int
+        Fourier modes to retain in each dimension.
+        The dimensionality of the MWNO is inferred from len(n_modes).
+        If int: Creates (alpha,) for 1D, (alpha, alpha) for 2D, etc.
+        Must provide either n_modes or alpha (not both).
```
vduruiss: Maybe can remove mention of alpha since we will remove it

vduruiss: Might be good to remove alpha completely everywhere and just use n_modes to avoid having two variables for the same thing

neuralop/models/mwno.py
```diff
@@ -0,0 +1,461 @@
...
+
+    By leveraging orthogonality, compact support, and multi-resolution properties of
+    multiwavelets, it effectively captures both global structures and local irregularities.
+    This makes it particularly well-suited for learning operators from data with high
+    fluctuations and supports accurate modeling of complex dynamical systems governed
+    by partial differential equations.
+
+    Main Parameters
+    ---------------
+    n_modes : Tuple[int, ...] or int
```
vduruiss: With the current implementation, `n_modes` cannot really be a tuple. It seems the code only takes `n_modes[0]` and then uses it everywhere. 

Either we allow for the number of modes to be different along different dimensions, or otherwise we update the docstring and an assertion at the beginning of the model definition to make it clear n_modes but be a single integer

neuralop/models/mwno.py
```diff
@@ -0,0 +1,461 @@
...
+    -----
+    - Spatial dimensions must be powers of 2 for wavelet decomposition
+    - For 3D, only first two dimensions are decomposed (time preserved)
+    - ReLU activation applied between MWNO layers (except after last layer)
+    - Output channel dimension auto-squeezed if out_channels=1
+
+    Input Shapes:
+        - 1D: (batch, n_points, in_channels) where n_points must be a power of 2
+        - 2D: (batch, height, width, in_channels) where height, width must be powers of 2
+        - 3D: (batch, height, width, time, in_channels) where height, width must be powers of 2
```
vduruiss, line 87: Definitely need to add an assert and throw an error if one of the spatial resolutions is not a power of 2

neuralop/models/mwno.py
```diff
@@ -0,0 +1,461 @@
...
+        Parameters
+        ----------
+        x : torch.Tensor
+            Input function values
+            - 1D: (batch, n_points, in_channels)
+            - 2D: (batch, height, width, in_channels)
+            - 3D: (batch, height, width, time, in_channels)
+
+            Requirements:
+            - Spatial dimensions must be powers of 2
```
vduruiss, line 419: Definitely need to add an assert and throw an error if one of the spatial resolutions is not a power of 2

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,1553 @@
...
+
+                    # Stretched version
+                    poly_stretched = cheb_poly(Chebyshev([-1, 4]))
+                    coeffs_2x = poly_stretched.coef * np.sqrt(2) * 2 / np.sqrt(np.pi)
+                    scaling_2x_coeffs[basis_idx, :len(coeffs_2x)] = coeffs_2x
+
+            # For Chebyshev, wavelets are handled differently (compact support)
+            phi = [partial(cls.phi_, scaling_coeffs[i, :]) for i in range(k)]
+            psi1 = [partial(cls.phi_, np.zeros(k), lb=0, ub=0.5) for i in range(k)]
+            psi2 = [partial(cls.phi_, np.zeros(k), lb=0.5, ub=1) for i in range(k)]
```
vduruiss: `phi` does not take these `lb` and `ub` arguments anymore. Please update the names and make sure everything runs smoothly

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,1553 @@
...
+                    poly_on_unit_interval = cheb_poly(Chebyshev([-1, 2]))
+                    coeffs = poly_on_unit_interval.coef * 2 / np.sqrt(np.pi)
+                    scaling_coeffs[basis_idx, :len(coeffs)] = coeffs
+
+                    # Stretched version
+                    poly_stretched = cheb_poly(Chebyshev([-1, 4]))
+                    coeffs_2x = poly_stretched.coef * np.sqrt(2) * 2 / np.sqrt(np.pi)
+                    scaling_2x_coeffs[basis_idx, :len(coeffs_2x)] = coeffs_2x
+
+            # For Chebyshev, wavelets are handled differently (compact support)
```
vduruiss: It does not seem like the Chebyshev version runs correctly. With the below code, it seems that psi1 and psi2 are just zeros. Make sure to run the code with the Chebyshev version and make sure it works properly

neuralop/layers/mwno_block.py
```diff
@@ -0,0 +1,1553 @@
...
+            self.weights1[:, :, :num_modes_x, :num_modes_y]
+        )
+
+        # Negative frequencies in x: [-num_modes_x, ..., -1]
+        # In FFT layout, these are at indices [nx-num_modes_x:nx]
+        output_fft[:, :, -num_modes_x:, :num_modes_y] = torch.einsum(
+            "bixy,ioxy->boxy",
+            x_fft[:, :, -num_modes_x:, :num_modes_y],
+            self.weights2[:, :, :num_modes_x, :num_modes_y]
+        )
```
vduruiss: These could be overlapping and problematic if alpha is too large. This is why I would recommend using the SpectralConv class directly, or starting from there

neuralop/models/mwno.py
```diff
@@ -0,0 +1,461 @@
...
+        Only used if n_modes is not provided.
+        Example: alpha=12 with n_dim=2 → modes=(12, 12).
+    n_dim : int, optional
+        Spatial dimensionality (1, 2, or 3).
+        Only needed if using alpha parameter.
+        Inferred from n_modes if n_modes is a tuple.
+    L : int, optional
+        Number of coarsest decomposition levels to skip. Default: 0
+        Reduces computation by stopping wavelet decomposition early.
+        L=0: Full decomposition to coarsest scale. L=1: Stop 1 level before coarsest.
```
vduruiss, line 51: Would be good to add an assert that L needs to be smaller than `num_scales`

neuralop/models/tests/test_mwno.py
```diff
@@ -0,0 +1,78 @@
...
+
+from ..mwno import MWNO
+
+
+@pytest.mark.parametrize("n_modes", [[16], (12,12),(8,8,8)])
+@pytest.mark.parametrize("k", [4])
+@pytest.mark.parametrize("c", [4, 16])
+@pytest.mark.parametrize("n_layers", [3])
+@pytest.mark.parametrize("L", [1])
+@pytest.mark.parametrize("base", ["legendre"])
```
vduruiss: Need to test more cases, and definitely need to test chebyshev. If it becomes a number of configs that is too large, we can split it in 2-3 test functions

vduruiss:
Thank you for the updates @Shixuan01 !

I have added a few more comments. There seems to be a few remaining issues in the code, especially for the Chebyshev option which currently does not work

It would also be nice to have an example script for the multiwavelet NO. That could also help make sure the code is working correctly, both for the Chebyshev and Legendre cases. 

You could add examples of (tuned) MWNO hyperparameters (maybe one for Chebyshev and one for Legendre) in the config files ([here](https://github.com/neuraloperator/neuraloperator/tree/main/config)) and see if it works when loaded by the training script for NS equations:
[https://github.com/neuraloperator/neuraloperator/blob/main/scripts/train_navier_stokes.py](https://github.com/neuraloperator/neuraloperator/blob/main/scripts/train_navier_stokes.py)

neuralop/models/mwno.py
shixuanli01: Thanks — we now use ChannelMLP for projection (same pattern as FNO), with projection_channels=0 mapped to hidden width 128 so the documented default is actually applied; forward permutes to channels-first for the MLP and back.

neuralop/models/mwno.py
shixuanli01: Dropped redundant alpha / duplicate self.alpha; the public API is n_modes only (with optional int shorthand for 1D). MWNOBlock now takes n_modes as well; internal kernels still use the first entry as the mode cutoff, documented in Notes.

neuralop/models/mwno.py
shixuanli01: Good point. The public API now requires n_modes to be one int plus explicit n_dim, with validation in __init__, and the docs no longer imply a tuple. Anisotropic mode counts would need deeper changes in SparseKernelFT; we can leave that for a later PR if desired.

neuralop/models/mwno.py
shixuanli01: Added _validate_spatial_resolution at the start of forward: we require power-of-two sizes on the wavelet axes (1D: n_points; 2D: H,W; 3D: H,W only) and raise ValueError with an explicit message otherwise, plus a small unit test for the 2D case.

neuralop/layers/mwno_block.py
shixuanli01: phi_ uses lower_bound / upper_bound, not lb / ub. Updated the Chebyshev partial(...) calls accordingly so the wavelet callables match the current signature.

neuralop/layers/mwno_block.py
shixuanli01: You were right: Chebyshev psi1/psi2 were zero polynomials. We now convert the Chebyshev scaling rows to power basis (so the same Gram–Schmidt + monomial integrals as Legendre are valid), reuse that Gram–Schmidt for Chebyshev wavelet coefficients, and build psi1/psi2 with phi_ and lower_bound/upper_bound on [0,0.5] / [0.5,1]. Added tests to assert non-zero wavelets and a finite Chebyshev filter bank.

neuralop/layers/mwno_block.py
shixuanli01: Thanks — we were indeed using m = min(α, nx//2+1) on the full FFT axis, so [:m] and [-m:] could overlap when 2m > nx. We now cap symmetric blocks at nx//2 per side (and treat nx≤1 as DC-only with no negative slice, since [-0:] aliases the whole axis). The last rfft dimension still uses n//2+1. We added a short note pointing to SpectralConv’s handling of FFT redundancy; wiring SparseKernelFT to SpectralConv directly would be a larger API/weight-shape change if we want that refactor later. This would require a complete structural modification.

neuralop/models/mwno.py
shixuanli01: Added a ValueError in MWNO.forward when L >= num_scales (with num_scales = floor(log2(spatial_size)) on the wavelet axis), documented the constraint on L for both MWNO and MWNOBlock, and a small unit test.

neuralop/models/tests/test_mwno.py
shixuanli01: Thanks — I split the parametrized smoke test into test_mwno_legendre and test_mwno_chebyshev with the same grid (n_dim/n_modes, c, n_layers ∈ {1,3}, L ∈ {0,1,2}, k=4) and factored the forward/backward + grad checks into a small helper so we don’t duplicate code. If 72 parametrized runs is too heavy for CI we can trim (e.g. drop L=2 or one of n_layers).

shixuanli01:
Hi, is there any more code I need to improve?

PR #637, opened 2025-08-07
[BUGFIX] Fix TypeError caused by len() on None in CODANO

sjh97:
### 🐛 Bug Description

In the `CODANO` class constructor, the following assertion is currently written as:

```python
assert (
    len(n_heads) == n_layers or n_heads is None
), "number of Attention head for all layers are not given"
assert (
    len(per_layer_scaling_factors) == n_layers
    or per_layer_scaling_factors is None
), "scaling for all layers are not given"
assert (
    len(attention_scaling_factors) == n_layers
    or attention_scaling_factors is None
), "attention scaling for all layers are not given"
```

However, if `n_heads`, `per_layer_scaling_factors`, or `attention_scaling_factors`  are None, this leads to a TypeError:
```python
TypeError: object of type 'NoneType' has no len()
```

✅ Fix Summary
This PR fixes the bug by changing the order of conditions in the assertion statements to check for None first. This avoids evaluating len() on a NoneType object.

Updated code:
```python
assert (
     n_heads is None or len(n_heads) == n_layers
), "number of Attention head for all layers are not given"
assert (
    per_layer_scaling_factors is None or
    len(per_layer_scaling_factors) == n_layers
), "scaling for all layers are not given"
assert (
    attention_scaling_factors is None or
    len(attention_scaling_factors) == n_layers
), "attention scaling for all layers are not given"
```

This change aligns with the later logic in the code, where default values are set when these arguments are None.

🧪 Added Test File
To verify the fix and provide a simple example usage of CODANO, I added a new notebook:
`test.ipynb`

This notebook includes minimal code to instantiate the model with default parameters and ensures that the bug no longer occurs.

Thanks for reviewing this PR!

### DIFF
--- a/neuralop/models/codano.py
+++ b/neuralop/models/codano.py
@@ -195,15 +195,15 @@ class CODANO(nn.Module):
         self.n_layers = n_layers
         assert len(n_modes) == n_layers, "number of modes for all layers are not given"
         assert (
-            len(n_heads) == n_layers or n_heads is None
+             n_heads is None or len(n_heads) == n_layers
         ), "number of Attention head for all layers are not given"
         assert (
+            per_layer_scaling_factors is None or
             len(per_layer_scaling_factors) == n_layers
-            or per_layer_scaling_factors is None
         ), "scaling for all layers are not given"
         assert (
+            attention_scaling_factors is None or
             len(attention_scaling_factors) == n_layers
-            or attention_scaling_factors is None
         ), "attention scaling for all layers are not given"
         if use_positional_encoding:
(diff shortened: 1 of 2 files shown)


sjh97:
Thanks for the feedback! I've removed `test.ipynb` and updated the test_codano.py. Let me know if there's anything else I should adjust.

vduruiss:
Thank you for bringing that up @sjh97 ! I did not realize there was already a PR on this and opened up a new one to respond to an issue bringing up this problem.

[PR 710](https://github.com/neuraloperator/neuraloperator/pull/710) essentially applies the same fix as the one you proposed. 

Thank you!

PR #640, opened 2025-08-14
Clarifying central differences code

vduruiss:
Just clarifying that central differences by default assume periodic BCs

Also making the periodic and non-periodic cases more distinct, instead of doing the periodic BCs, and then overwriting for the non-periodic case

### DIFF
--- a/neuralop/losses/finite_diff.py
+++ b/neuralop/losses/finite_diff.py
@@ -3,12 +3,21 @@ import torch
 finite_diff.py implements utilities for computing derivatives via finite-difference method
 """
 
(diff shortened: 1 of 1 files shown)


vduruiss:
Maybe we could also change the names `fix_x_bnd`,   `fix_y_bnd` , ... to  something like `periodic_in_x` , `periodic_in_y`, ...

I did not make that change yet since many codes might be relying on the current version with these argument names

dhpitt (approved):
Good changes @vduruiss , thank you. We definitely could rename the variables for clarity but I don't think it's necessary

JeanKossaifi:
Thanks @vduruiss ! I like the name suggestion too, it makes it a lot clearer what the arguments do!

PR #641, opened 2025-08-22
Adding CONTRIBUTING.md

vduruiss:
Adding a `CONTRIBUTING.md` as suggested by [pytorch-fdn](https://github.com/pytorch-fdn/ecosystem/issues/40#issuecomment-3194703526) 

Similar to the contribution guide from the website + a few additions

Added CODE_OF_CONDUCT.md file pointing to the standard Linux code of conduct

Also moved a few things from README to CONTRIBUTING.md to reduce redundancy

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -61,7 +61,7 @@ Participants are expected to:
 
 ## Incident Procedure
 
-To report a violation or appeal a decision, please contact: [Jean Kossaifi](https://github.com/JeanKossaifi)
+To report a violation or appeal a decision, please contact: [neuraloperator@outlook.com](mailto:neuraloperator@outlook.com)
 
 Include relevant details (including links, screenshots, etc.). Every effort will be made to maintain a safe, collegial, and fair environment. The project reserves the right to take action, including removing individuals from participation, as necessary.
 


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,5 @@
+# Code of Conduct
+
+This project follows the [Linux Foundation Code of Conduct](https://lfprojects.org/policies/code-of-conduct/).
+
+All participants are expected to uphold this Code of Conduct to ensure a friendly and welcoming environment for everyone.
```
JeanKossaifi: Thanks @vduruiss - this is great.
I wonder if shouldn't have one source of truth though where possible and perhaps keep CONTRIBUTING to a minimum but point to the webpage with more details (or vice-versa).

vduruiss: Do you mean have the full code of conduct written out somewhere instead of referencing the Linux one?

JeanKossaifi: Yes. What do you think? Also means we have a clear reference point if the link gets updated over time, we can actively look at these changes and make sure everyone is aware of the changes as opposed to passively assuming members stay up to date.

vduruiss: @JeanKossaifi  I updated the code of conduct to only keep relevant parts from the Linux one

Let me know if there are other things you would like to emphasize

PR #643, opened 2025-09-01
Finite Differences and Fourier Differentiation

vduruiss:
Adding 2d and 3d Fourier differentiation (also more efficient by evaluating derivatives jointly within a single FFT-IFFT)

Made classes for 1d/2d/3d  Fourier differentiation and central differences, with more sub-functions for derivatives

These are first drafts, we can decide if we want to write these classes differently

Also added more comprehensive tests and examples

### DIFF
--- a/examples/data_gen/plot_burgers_2d_solver.py
+++ b/examples/data_gen/plot_burgers_2d_solver.py
@@ -1,0 +1,113 @@
+"""
+.. _burgers_2d_fd_vis :
+
+A Numerical Solver for Burgers' Equation in 2 Dimensions 
+========================================================
+An intro to our loss module's finite difference utility demonstrating
+its use to create a simple numerical solver for Burgers' equation in 2d.
+"""
+
+# %%
+# Import the library
+# ------------------
+# We first import our `neuralop` library and required dependencies.
+import torch
+import numpy as np
+import matplotlib.pyplot as plt
+import matplotlib.animation as animation
+from neuralop.losses.differentiation import FiniteDiff  
+
+device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
+
+# %%
+# Defining our problem
+# --------------------
+# We aim to solve the 2D viscous Burger's equations:
+#
+# :math: `u_t + u \cdot u_x + v \cdot u_y = \nu (u_xx + u_yy)`
+#
+# :math: `v_t + u \cdot v_x + v \cdot v_y = \nu (v_xx + v_yy)`
+
+## Simulation parameters
+Lx, Ly = 2.0, 2.0   # Domain lengths
+nx, ny = 64, 64   # Grid resolution
+T = 1    # Total simulation time
+dt = 0.001  # Time step
+nu = 0.04   # Viscosity
+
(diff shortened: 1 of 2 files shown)


neuralop/losses/finite_diff.py
```diff
@@ -29,198 +28,651 @@ def central_diff_1d(x, h, periodic_in_x=True):
...
+    return fd2d.dx(x), fd2d.dy(x)
+
+
+def central_diff_3d(x, h, periodic_in_x=True, periodic_in_y=True, periodic_in_z=True):
+    """
+    Backward compatibility function for central_diff_3d.
+    Creates a FiniteDiff3D instance and returns dx, dy, dz.
+    """
+    fd3d = FiniteDiff3D(h=h, periodic_in_x=periodic_in_x, periodic_in_y=periodic_in_y, periodic_in_z=periodic_in_z)
+    return fd3d.dx(x), fd3d.dy(x), fd3d.dz(x)
```
vduruiss: I kept the old functions like this for the moment, but we can also remove them

neuralop/losses/fourier_diff.py
```diff
@@ -1,95 +1,762 @@
...
+        """
+        self.L = L
+        self.use_FC = use_FC
+        self.FC_d = FC_d
+        self.FC_n_additional_pts = FC_n_additional_pts
+        self.low_pass_filter_ratio = low_pass_filter_ratio
+
+    
+    def compute_multiple_derivatives(self, u, orders):
+        """
```
JeanKossaifi: Nitpick: there seems to be trailing spaces - maybe running black/pep8 on the file would be useful.

neuralop/losses/fourier_diff.py
```diff
@@ -1,95 +1,762 @@
...
-    for certain functions and parameter combinations.
+    A class for computing 1D Fourier derivatives with Fourier continuation support.
+    
+    Provides methods for computing 1D Fourier derivatives,
+    with optional Fourier continuation for handling non-periodic functions.
+    """
+    
+    def __init__(self, 
+                 L=2*torch.pi, 
+                 use_FC=False, FC_d=4,
```
JeanKossaifi: I tend to prefer using small letters for non-classes variables, would tend to use use_fc, fc_degree, etc. 

I'd use fc_degree to differentiate degree VS dim, though should be somewhat obvious from the context.

JeanKossaifi (commented):
@vduruiss - amazing PR, as always!

Few notes: 
* I'd run black (or some other formatter) to get rid of extra lines, and other minor stylistic things
* I wonder whether we could add a class that works for any 1 <= order <= 3, e.g. FourierDiff that then calls the correct one. Very optional
* It would be great I think to add a section in the user guide on the basic math behind this - I would normally suggest docstring, but here that would mean repeating it for each number of dimensions (1D, ..., 3D). 

What do you think?

neuralop/losses/finite_diff.py
vduruiss: @JeanKossaifi Should I remove these?

PR #644, opened 2025-09-01
FC Class

vduruiss:
Adding a FC Class and writing FCLegendre and FCGram as subclasses to simplify the code and remove redundancy


JeanKossaifi:
Looks great - would love to include the code to actually generated these continuations (we can use e.g. float128 if needed), we could run it once during setup of the package and cache them.

JeanKossaifi:
Looks good to me, thank you for the MR and the changes @vduruiss ! Happy to merge unless you want to add anything else.

vduruiss:
@JeanKossaifi Should be ready to merge!

PR #645, opened 2025-09-01
Add scale-consistency training

zongyi-li:
Implement the scale-consistency training from paper https://arxiv.org/abs/2507.18813
We added four files; no existing filed is modified.
- `neuralop/losses/scale_consistency_losses.py` the loss scale_consistency by compare global domain and sub domains.
- `neuralop/data/transforms/rescale.py` rescaling transforms that zoom in to subdomains.
- `neuralop/training/scale_consistency.py` the subclass trainer that pass the model to the loss.
- `example/training/plot_scale_consistency_FNO_darcy.py` an example of scale consistency training on boundary-valued darcy problem.


dhpitt:
This is great, thanks so much @zongyi-li ! Comments to follow but this looks really good. 

One note: the online runner does not haver access to the 128x128 darcy-flow dataset by default as it is not pre-included in the library; we can either download it (e.g. call `DarcyDataset(..., download = True)` or use a smaller example.

neuralop/data/transforms/rescale.py
```diff
@@ -0,0 +1,372 @@
...
+        initial condition, a tensor of shape (batch, channels, X, Y)
+    y: torch.Tensor
+        boundary condition, a tensor of shape (batch, channels, X, Y)
+
+    Returns
+    -------
+    x: torch.Tensor
+        the combined input tensor with boundary conditions added in channels 1,...,4
+    """
+    unsqueeze_x = x.ndim == 3  # check whether batch dimension is missing
```
dhpitt: Perhaps instead of inferring the batch dynamically, it would be simpler to have the forward call `DarcyExtractBC(x,y,batched=True)`?

neuralop/data/transforms/rescale.py
```diff
@@ -0,0 +1,372 @@
...
+    unsqueeze_y = y.ndim == 3
+
+    # add batch dimension if there was no batch dimension
+    if unsqueeze_x:
+        x = x.unsqueeze(0)
+    if unsqueeze_y:
+        y = y.unsqueeze(0)
+
+    # extract BC
+    Nx, Ny = y.shape[-2], y.shape[-1]
```
dhpitt, line 33: Some documentation of what the BC is/what this extracts might be helpful. Perhaps one line

neuralop/data/transforms/rescale.py
```diff
@@ -0,0 +1,372 @@
...
+        initial condition, a tensor of shape (batch, channels, X, T)
+    y: torch.Tensor
+        boundary condition, a tensor of shape (batch, channels, X, T)
+
+    Returns
+    -------
+    x: torch.Tensor
+        the combined input tensor with boundary conditions added in channels 1 and 2
+    """
+    if y.ndim > 3:
```
dhpitt: Same notes here re: inferring dimensions and documenting the BC

neuralop/data/transforms/rescale.py
```diff
@@ -0,0 +1,372 @@
...
+
+    x = y[:, 0, :]  # x (batch, x)
+    x = x.reshape(-1, 1, 1, S).repeat(1, 1, T, 1)  # x (batch, 1, T, x)
+    x = torch.cat([x, boundary], dim=1)  # x (batch, 1+2, T, x)
+    return x
+
+
+def HelmholtzExtractBC(x, y):
+    """
+    Extract boundary conditions for Helmholtz Equation: 2D boundary value problem.
```
dhpitt, line 84: Same as above

neuralop/data/transforms/rescale.py
```diff
@@ -0,0 +1,372 @@
...
+    def __call__(self, x, y):
+        return self.forward(x, y)
+
+    def forward(self, x, y):
+        if torch.rand(1) < self.p:
+            x, y = F.hflip(x), F.hflip(y)
+        if torch.rand(1) < self.p:
+            x, y = F.vflip(x), F.vflip(y)
+        if torch.rand(1) < self.p:
+            x, y = x.transpose(-1, -2), y.transpose(-1, -2)
```
dhpitt, line 177: is this a flip and a rotation? Seems like a composition/superset of `RandomRotation`

neuralop/data/transforms/rescale.py
```diff
@@ -0,0 +1,372 @@
...
+        return self.forward(x, y)
+
+    def forward(self, x, y):
+        if torch.rand(1) < self.p:
+            x, y = x.transpose(-1, -2), y.transpose(-1, -2)
+        x = ExtractBC(x, y)
+        return x, y
+
+
+class RandomCropResize:
```
dhpitt: Is there a specific reason we can't reshape and use `torchvision.transforms` directly?

neuralop/losses/scale_consistency_losses.py
```diff
@@ -0,0 +1,141 @@
...
+        y_small = transform_xy.crop(y, i, j, h, w)
+        x_small = ExtractBD(x_small, y_small)
+
+        if group_action is not None:
+            x_small, y_small = group_action(x_small, y_small)
+
+        if type == "darcy":
+            y_small_ = model(x_small)
+        else:
+            y_small_ = model(x_small, re)
```
dhpitt, line 110: This forward call seems like it might belong better in the custom trainer's training loop

neuralop/training/scale_consistency.py
```diff
@@ -0,0 +1,79 @@
...
+                loss = training_loss(out, **sample)
+        else:
+            loss = training_loss(out, **sample)
+
+        # self-consistency term (optional)
+        if self.selfconsistency_loss is not None and self.selfconsistency_weight != 0.0:
+            model = self._unwrap_model()
+            if self.mixed_precision:
+                with torch.autocast(device_type=self.autocast_device_type):
+                    sc_loss = self.selfconsistency_loss(model, sample["x"], loss_fn=training_loss, y=sample["y"])
```
dhpitt, line 70: As above; might make more sense if things like the model's forward call went here, and as little as possible of the computation belonged in the loss function

dhpitt (commented):
Thanks again for this contribution, @zongyi-li. A few minor notes on style and clarity for future maintainability but overall this looks very good. Passing on to @JeanKossaifi for review.

JeanKossaifi:
This looks good, thanks @zongyi-li for the PR and @dhpitt for the review.

My main question is whether there shouldn't be some more separation between the concepts (the algorithm) and the problem specific implementation (for Darcy, Burgers, etc). Perhaps this is just a matter of refactoring a little. What do you think?

neuralop/data/transforms/rescale.py
zongyi-li: Thanks for the suggestion! I just added `batched=True` and remove the dimension checking lines.

neuralop/losses/scale_consistency_losses.py
zongyi-li: Yea we added a custom trainer, but I feel it will also be beneficial to have a loss, which we can use in other trainers in the future.

PR #646, opened 2025-09-02
Removing subclasses FNO*d, TFNO*d

vduruiss:
Removing the subclasses FNO1d, FNO2d, FNO3d, TFNO1d, TFNO2d, TFNO3d

I think some people are still using these, but they have been depreciated and not maintained to keep up with latest changes (e.g.  lifting_channels vs. lifting_channel_ratio)

This can be misleading because some of the parameters specified for these subclasses are not actually enforced when passed to the main FNO class, which means users may end up working with something different from what they want

I suggest we remove them completely, unless there is another reason to keep them


JeanKossaifi:
I agree, they were only here for legacy reasons, to support the code from the original paper. Perhaps TFNO could be an alias for FNO even and we could harmonize the defaults.

neuralop/models/fno.py
```diff
@@ -648,7 +432,38 @@ def partialclass(new_name, cls, *args, **kwargs):
...
+    >>>
+    >>> # Equivalent FNO model with explicit factorization:
+    >>> model = FNO(n_modes=(12, 12), in_channels=1, out_channels=1, hidden_channels=64,
+    ...             factorization="Tucker", rank=0.1)
+    """
+
+    def __init__(self, *args, **kwargs):
+        kwargs.setdefault("factorization", "Tucker")
+        kwargs.setdefault("rank", 0.1)
+        super().__init__(*args, **kwargs)
```
vduruiss, line 479: @JeanKossaifi  Is this better?

vduruiss:
@JeanKossaifi  I currently left the possibility to use kwargs for the forwards as this might be necessary for the Trainer's logic in certain cases. For instance, when using multiple losses, each individual loss takes all the arguments but only a few are used for each individual loss

JeanKossaifi:
That's a great PR and a huge quality of life improvement for users, thank you @vduruiss! Looks good to me, merging!

PR #647, opened 2025-09-03
BUG: Using a non-tuple sequence for multi-dim indexing is deprecated

StillJosh:
Indexing tensors with a list is deprecated and will result in errors starting with PyTorch 2.9. 

Currently, this yields a warning for a minimal example like 
```
from neuralop.models import FNO
import torch

operator = FNO(n_modes=(16, 16), hidden_channels=64,
                in_channels=3, out_channels=1)

data = torch.randn(2, 3, 64, 64)

output = operator(data)
```

```
UserWarning: Using a non-tuple sequence for multidimensional indexing is deprecated and will be changed in pytorch 2.9; use x[tuple(seq)] instead of x[seq]. In pytorch 2.9 this will be interpreted as tensor index, x[torch.tensor(seq)], which will result either in an error or a different result (Triggered internally at [...]/torch/csrc/autograd/python_variable_indexing.cpp:312.)
  out_fft[slices_x] = self._contract(x[slices_x], weight, separable=self.separable)
```
due to indexing the weight tensor in FNO with a list of slices, depending on the input data. 

Replaced indexing lists with tuples. Kept lists only where mutability was required and converted them to tuples at indexing time.

### DIFF
--- a/neuralop/data/datasets/pt_dataset.py
+++ b/neuralop/data/datasets/pt_dataset.py
@@ -122,6 +122,7 @@ class PTDataset:
         train_input_indices = [slice(0, n_train, None)] + [slice(None, None, rate)\
                                                            for rate in input_subsampling_rate]
         train_input_indices.insert(channel_dim, slice(None))
+        train_input_indices = tuple(train_input_indices)
         x_train = x_train[train_input_indices]
         
         y_train = data["y"].clone()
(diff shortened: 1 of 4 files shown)


dhpitt:
Thank you for making this fix! Approved.

JeanKossaifi:
Thank you, merging!

PR #648, opened 2025-09-04
Divergence-free Spectral Projection

vduruiss:
This is code for enforcing the divergence-free constraint in 2d vector field.

This is based on our [ICML Workshop Paper](https://openreview.net/forum?id=Zvxm14Rd1F)

I updated completely our earlier projection code to use an explicit formula for the linear least-squares problem for the divergence-free constraint, instead of keeping the memory-heavy and time-consuming pinv

@mliuschi  can also have a look

Note that the example and test are based on the new differentiation code from [PR643](https://github.com/neuraloperator/neuraloperator/pull/643)

### DIFF
--- a/neuralop/layers/spectral_projection.py
+++ b/neuralop/layers/spectral_projection.py
@@ -1,0 +1,151 @@
+import torch
+import numpy as np
+import torch.nn.functional as F
+
+
+def spectral_projection_divergence_free(u, domain_size, constraint_modes):
+    """
+    Apply spectral projection to make velocity field divergence-free.
+    
+    This method implements a Helmholtz-Hodge projection in the spectral domain
+    to project velocity fields onto the divergence-free subspace.
+    
+    
+    References:
+    -----------
+    The method is based on the spectral projection approach described in:
+    
+        [1] Towards Enforcing Hard Physics Constraints in Operator Learning Frameworks
+        V. Duruisseaux, M. Liu-Schiaffini, J. Berner, and A. Anandkumar
+        ICML 2024 AI for Science Workshop
+        https://openreview.net/pdf?id=Zvxm14Rd1F
+        
+        [2] Enforcing physical constraints in CNNs through differentiable PDE layer. 
+        C. M. Jiang, K. Kashinath, P. Prabhat, and P. Marcus
+        ICLR 2020 Workshop on Integration of Deep Neural Models and Differential Equations, 2020.
+        https://openreview.net/pdf?id=q2noHUqMkK
+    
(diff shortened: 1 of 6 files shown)


neuralop/layers/spectral_projection.py
```diff
@@ -0,0 +1,140 @@
...
+
+def spectral_projection_divergence_free(u, domain_size, constraint_modes):
+    """
+    Apply spectral projection to make velocity field divergence-free.
+    
+    This method implements a Helmholtz-Hodge projection in the spectral domain
+    to project velocity fields onto the divergence-free subspace.
+    
+    
+    References:
```
JeanKossaifi: This is a really great doctoring!

Do all the sections actually get rendered? 
I would also put Parameters and Return first for consistency.

vduruiss: The `plot_spectral_projection.py` example gets rendered correctly

vduruiss: Fixed the docstring!

neuralop/layers/spectral_projection.py
```diff
@@ -0,0 +1,143 @@
...
+        neg_idx_h = torch.arange(height - constraint_modes[0]//2, height, 1)
+        idx_h = torch.cat((pos_idx_h, neg_idx_h)).to(device)
+        u_ft = torch.index_select(u_ft, 2, idx_h)
+    
+    # Extract lower width modes where the constraint is applied
+    if width != constraint_modes[1]:
+        pos_idx_w = torch.arange(0, constraint_modes[1]//2, 1)
+        neg_idx_w = torch.arange(width - constraint_modes[1]//2, width, 1)
+        idx_w = torch.cat((pos_idx_w, neg_idx_w)).to(device)
+        u_ft = torch.index_select(u_ft, 3, idx_w)
```
JeanKossaifi: One option would be to index both dims jointly as we do in the main spectral conv, after an fftshift?

vduruiss: Updated!

```  # Extract lower modes for both dimensions where the constraint is applied
    if (height != constraint_modes[0]) or (width != constraint_modes[1]):
        u_ft = torch.fft.fftshift(u_ft, dim=(2, 3))
        u_ft = u_ft[:, :, 
                    (height - constraint_modes[0])//2 : (height - constraint_modes[0])//2 + constraint_modes[0],
                    (width - constraint_modes[1])//2 : (width - constraint_modes[1])//2 + constraint_modes[1]]
        u_ft = torch.fft.ifftshift(u_ft, dim=(2, 3))

PR #653, opened 2025-09-26
Fix missing **kwargs in FNO1d, FNO2d & FNO3d to allow passing arguments like positional_embedding

mabetr:
Hello there,

In my project I needed to use FNO1d with positional_embedding=None, and it was not working. I noticed the **kwargs argument was not passed to super().__init__(...) calls. I wrote a test, proved it wasn't working, fixed it, test is passing.

Let me know if instead of passing **kwargs you'd like to pass a specific list of arguments (or feel free to do this change on your own).

Marine

### DIFF
--- a/neuralop/models/fno.py
+++ b/neuralop/models/fno.py
@@ -463,6 +463,7 @@ class FNO1d(FNO):
             implementation=implementation,
             decomposition_kwargs=decomposition_kwargs,
             domain_padding=domain_padding,
+            **kwargs
         )
         self.n_modes_height = n_modes_height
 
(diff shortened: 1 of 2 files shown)


vduruiss:
Hi Marine, thank you for reaching out!

FNO1d, FNO2d, FNO3d were the earlier versions we had for the FNO in the different dimensions. We eventually combined them all together in a single FNO class. Since then, certain updates to the library were not propagated to these older classes, so they are now outdated and could be misleading. 

We will soon update the library and the documentation to remove these deprecated classes so they are not misused.

To instantiate an FNO in 1d, 2d, 3d, you can simply use the FNO class and specify a tuple with the number of modes for every dimension. For instance

`model = FNO(n_modes=(64,), ...)` will instantiate a FNO in 1d
`model = FNO(n_modes=(64, 64), ...)` will instantiate a FNO in 2d
`model = FNO(n_modes=(64, 64, 64), ...)` will instantiate a FNO in 3d

Let us know if you have further questions!

PR #654, opened 2025-09-28
Add OTNO, and an example of ShapeNet-Car dataset

Xinyi-Li-4869:
This PR introduces optimal transport neural operator (OTNO) with an example of ShapeNet-Car dataset.

The main classes are:
1.  OTDataModule: generate OT data from physical mesh
2. OTNO: model implementation

For car dataset example, I added files according to GINO. The main files are:
- neuraloperator/config/otno_carcfd_config.py
- neuraloperator/neuralop/data/datasets/car_ot_dataset.py
- neuraloperator/scripts/train_otno_carcfd.py

Paper link: https://arxiv.org/abs/2507.20065
Original code: https://github.com/Xinyi-Li-4869/OTNO


neuralop/models/otno.py
```diff
@@ -0,0 +1,101 @@
+import torch
+import torch.nn as nn
+
+from neuralop.models import FNO
+from neuralop.layers.channel_mlp import ChannelMLP as NeuralopMLP
```
vduruiss: Probably better to keep the original `ChannelMLP` name

neuralop/models/otno.py
```diff
@@ -0,0 +1,101 @@
...
+            factorization=None,
+            rank=1,
+            joint_factorization=False,
+            fixed_rank_modes=False,
+            implementation="factorized",
+            decomposition_kwargs=dict(),
+            domain_padding=None,
+            SpectralConv=SpectralConv,
+            **kwargs
+    ):
```
vduruiss, line 25: Most of these are the original FNO arguments. Maybe it's not worth redeclaring them, unless you want to emphasize different default values for certain parameters?

vduruiss: Related to that, since OTNO inherits from FNO, it would be good to check that every argument of the FNO can be used with every possible option. If not, it might be good to have an explicit way of handling the FNO arguments or options that cannot be passed to the OTNO

neuralop/models/otno.py
```diff
@@ -0,0 +1,101 @@
...
+            **kwargs
+        )
+
+        self.projection = NeuralopMLP(
+            in_channels=self.hidden_channels,
+            out_channels=out_channels,
+            hidden_channels=self.hidden_channels*projection_channel_ratio,
+            non_linearity=non_linearity,
+            n_dim=1,
+        )
```
vduruiss: Is this needed? This seems to be the same projection code as for the FNO, unless I am missing something. 

Why do you need to specify n_dim=1?

neuralop/models/otno.py
```diff
@@ -0,0 +1,101 @@
...
+
+    @property
+    def device(self):
+        """Returns the device that the model is on."""
+        return self.device_indicator_param.device
+
+    # trans: (1, in_channels, n_s_sqrt, n_s_sqrt)
+    def forward(self, trans, ind_dec, **kwargs):
```
vduruiss: Can probably change `trans` to `x` ?

vduruiss: `ind_dec` are the indices for the decoding?

vduruiss: Might be good to add a few lines of comments to explain the forward and what arguments it takes (and their shapes as well)

neuralop/models/otno.py
vduruiss: Also related to that, you could add a test_otno.py file similar to the ones for the other architectures, to make sure it works with other options 

[https://github.com/neuraloperator/neuraloperator/blob/main/neuralop/models/tests/](https://github.com/neuraloperator/neuraloperator/blob/main/neuralop/models/tests/)


These test codes in the current version of the library have actually not be maintained perfectly as we added more options, so it might be better to look at the updated test codes I have in an existing PR

[https://github.com/neuraloperator/neuraloperator/tree/4fdaec0fb0069a3affa73f727027ac390efed71c/neuralop/models/tests](https://github.com/neuraloperator/neuraloperator/tree/4fdaec0fb0069a3affa73f727027ac390efed71c/neuralop/models/tests)

scripts/run_test.py
```diff
@@ -0,0 +1,16 @@
...
+            n_total=2,
+        )'''
+from neuralop.data.datasets.car_ot_dataset import CarOTDataset
+
+if __name__ == "__main__":
+    data_module = CarOTDataset(
+            root_dir="D:/python_code/data/car-pressure-data",
+            n_train=2,
+            n_test=1,
+        )
```
vduruiss: Probably better to move a more comprehensive test script of the dataset  to `neuralop/data/datasets/tests`

neuralop/data/datasets/data/ot_expand3.0_reg1e-06_num3.pt
vduruiss, line 1: Maybe we can find a more descriptive name for the data file?

config/models.py
```diff
@@ -194,3 +194,18 @@ class GINO_Poisson2d(GINOConfig):
     fno_factorization: Optional[Any] = None
     fno_rank: float = 0.8
     fno_domain_padding: float = 0.0
+
+class OTNO_Small3d(ModelConfig):
+    model_arch: str = "otno"
+    n_modes: List[int] = [24, 24]
+    hidden_channels: int = 120
```
vduruiss: `hidden_channels = 120`  looks pretty big for me for a `Small3d` model. Maybe we can have it work decently with a smaller model?

scripts/run_test.py
```diff
@@ -0,0 +1,16 @@
...
+    data_module = OTDataModule(
+            root_dir="D:/python_code/data/car-pressure-data",
+            item_dir_name="data",
+            n_total=2,
+        )'''
+from neuralop.data.datasets.car_ot_dataset import CarOTDataset
+
+if __name__ == "__main__":
+    data_module = CarOTDataset(
+            root_dir="D:/python_code/data/car-pressure-data",
```
vduruiss: The `root_dir` probably need to be updated so it works for everyone

scripts/train_otno_carcfd.py
```diff
@@ -0,0 +1,167 @@
...
+    wandb.log({'time_to_distance': data_module.time_to_distance}, commit=False)
+
+trainer.train(
+              train_loader=train_loader,
+              test_loaders={'':test_loader},
+              optimizer=optimizer,
+              scheduler=scheduler,
+              training_loss=train_loss_fn,
+              #eval_losses={config.opt.testing_loss: test_loss_fn, 'drag': DragLoss},
+              eval_losses={config.opt.testing_loss: test_loss_fn},
```
vduruiss: Only keep one of these, maybe the most general one?

neuralop/data/datasets/ot_datamodule.py
```diff
@@ -0,0 +1,188 @@
...
+        item_dir_name : Union[str, Path]
+            directory in which individual item subdirs are stored
+        n_total : int, optional
+            hard limit on number of total examples
+            if n_total is greater than the actual number
+            of total examples available, nothing is changed
+        expand_factor : float, optional
+            Scale factor to map physical mesh size to latent mesh size (e.g., torus/sphere).
+        reg : float, optional
+            Regularization coefficient for the Sinkhorn algorithm.
```
vduruiss: Could be useful to add one more line for each them to mention how to choose these parameters and what can be reasonable numbers beyond the default values

neuralop/data/datasets/ot_datamodule.py
```diff
@@ -0,0 +1,188 @@
...
+            mesh.compute_vertex_normals()
+            normal = torch.from_numpy(np.asarray(mesh.vertex_normals).astype(np.float32()))
+            pressure = np.load(
+                str(data_dir / ('press_' + ind.zfill(3) + '.npy'))
+            ) # (3682,)
+            #pressure = np.concatenate((pressure[0:16], pressure[112:]), axis=0) #(3586,) this step moved to car_ot_dataset
+            pressure = torch.from_numpy(pressure.astype(np.float32()))
+
+            # OT
+            n_t = len(target) # number of target samples
```
vduruiss, line 124: Can remove these comments if not needed

neuralop/data/datasets/car_ot_dataset.py
```diff
@@ -0,0 +1,153 @@
...
+class CarOTDataset(OTDataModule):
+    """
+    Parameters
+    ----------
+    root_dir : Union[str, Path]
+        root directory at which data is stored.
+    n_train : int, optional
+        Number of training instances to load, by default 1
+    n_test : int, optional
+        Number of testing instances to load, by default 1
```
vduruiss: Maybe specify max number of samples in the dataset?

neuralop/data/datasets/car_ot_dataset.py
```diff
@@ -0,0 +1,153 @@
...
+    expand_factor : float, optional
+        Scale factor to map physical mesh size to latent mesh size (e.g., torus/sphere).
+    reg : float, optional
+        Regularization coefficient for the Sinkhorn algorithm.
```
vduruiss, line 31: Maybe add slightly more details about how to choose these parameters, and what are good values to try beyond the default ones

neuralop/data/datasets/car_ot_dataset.py
```diff
@@ -0,0 +1,153 @@
...
+            n_total=n_train+n_test,
+            expand_factor=expand_factor, 
+            reg=reg,
+            device=device,
+        )
+        
+        # process data list to remove specific vertices from pressure to match number of vertices
+        for i, item_data in enumerate(self.data):
+            press = item_data['press']
+            self.data[i]['press'] = torch.cat((press[0:16], press[112:]), axis=0)
```
vduruiss: Make sure to verify these hardcoded numbers provide the most general version possible

neuralop/data/datasets/car_ot_dataset.py
```diff
@@ -0,0 +1,153 @@
...
+        Load the saved Car-OT dataset we package along with our module.
+
+        See `neuralop.data.datasets.ot_datamodule` for more detailed references
+        """
+        n_total = n_train + n_test
+        data = torch.load(get_project_root() / "neuralop/data/datasets/data" / f"ot_expand{expand_factor}_reg{reg}_num{n_total}.pt")
+        # process data list to remove specific vertices from pressure to match number of vertices
+        for i, item_data in enumerate(data):
+            press = item_data['press']
+            data[i]['press'] = torch.cat((press[0:16], press[112:]), axis=0)
```
vduruiss: Same as earlier comment, make sure these hardcoded value s provide the most general setting. It might be good to use a variable name to specify these numbers, just so it is clearer what they are. This could make maintaining that code easier in the future

neuralop/models/otno.py
Xinyi-Li-4869: Thanks, have revised to keep the original ChannelMLP name.

neuralop/models/otno.py
Xinyi-Li-4869: Thanks for reminding me of the test code. Have simplified the hyperparameters of OTNO, remaining only important ones. And added test codes for OTNO. Find the `positional_embedding` having problem in previous version, and fixed it. Now the test code run successfully.

neuralop/models/otno.py
Xinyi-Li-4869: I have removed this now. It is need for an old version, n_dim=1 is also for that old release of NO library.

neuralop/models/otno.py
Xinyi-Li-4869: Have change `trans` to `x`, and yes, ind_dec is the indices for decoding. I have also added comments in forward. Thanks for suggestions!

scripts/run_test.py
Xinyi-Li-4869: Have removed

config/models.py
Xinyi-Li-4869: Thanks, I have revised to smaller model with `n_modes=[16, 16], hidden_channel=64`

scripts/run_test.py
Xinyi-Li-4869: Thanks, have revised

scripts/train_otno_carcfd.py
Xinyi-Li-4869: Sure, revised to keep only the L2 error which is most general

neuralop/data/datasets/car_ot_dataset.py
Xinyi-Li-4869: Have added details

neuralop/data/datasets/car_ot_dataset.py
Xinyi-Li-4869: In car_cfd_dataset.py, it also uses this. These numbers depend on the dataset, there's probably no general settings. If we update the shapenet dataset, we can remove this, but I don't have access to this published dataset

neuralop/data/datasets/ot_datamodule.py
Xinyi-Li-4869: Have added

neuralop/data/datasets/car_ot_dataset.py
Xinyi-Li-4869: If the n_train is bigger than the max number of samples, the OTDataModule will proceed to the full dataset and print the number of available train and test samples. So I think here is no much need to restate it.

neuralop/data/datasets/data/ot_expand3.0_reg1e-06_num3.pt
Xinyi-Li-4869: A more descriptive name would definitely be better — I just don’t have a good idea at the moment. Thanks for the suggestion!

examples/models/car_to_torus.gif
vduruiss, line 1: Let's remove the .gif, it should be generated automatically for the tutorial

examples/models/torus_to_car.gif
vduruiss, line 1: Let's remove the .gif, it should be generated automatically for the tutorial

examples/models/plot_OTNO_car_cfd.py
vduruiss, line 1: Looks good! Can we modify slightly so it does not save the gifs?
We can either have it so the .gif files are never created, or otherwise remove them at the end of the script.

The script below had an example of an animation. Maybe that's something you could adapt for the current animations? It might also be easier to run it without adding more imports/modules that need to be installed

https://neuraloperator.github.io/dev/auto_examples/data_gen/plot_diffusion_advection_solver.html

neuralop/models/tests/test_otno.py
```diff
@@ -0,0 +1,76 @@
...
+# data parameters
+n_s_sqrt = 18
+n_target_points = 100
+
+@pytest.mark.parametrize("use_mlp", [True, False])
+@pytest.mark.parametrize("norm", [None, "group_norm", "instance_norm"])
+@pytest.mark.parametrize("factorization", [None, "tucker"])
+@pytest.mark.parametrize("n_layers", [2, 4])
+@pytest.mark.parametrize("positional_embedding", [None, "grid"]) 
+@pytest.mark.parametrize("domain_padding", [0.0, 0.2, 0.4])
```
vduruiss: This is good, but it will run the code 144 times to test every single possible combination.

Can we break this into 2-3 test functions? This way it runs less combinations and would not require as much time to run for test/deployment

doc/requirements_doc.txt
```diff
@@ -5,4 +5,10 @@ sphinx
...
 sphinx-design
 tensorly_sphinx_theme
-torchtnt
\ No newline at end of file
+torchtnt
+pyvista
+imageio
+imageio[ffmpeg]
+imageio[pyav]
+pillow
```
vduruiss: See the comments for the tutorial. Maybe we can have the animations show up without requiring all these modules

Xinyi-Li-4869:
I’ve removed the GIFs and rewritten the content following the animation example, so I also removed the related packages from the documentation environment requirements. I’ve also reduced the test functions.

Please check again @vduruiss  — thanks so much!

JeanKossaifi:
Looks good to me as well, thank you both @Xinyi-Li-4869 @vduruiss !

PR #672, opened 2025-10-18
Added example script for showcasing sinusoidal embeddings

gluca99:


gluca99:
Latest change modified the previous example script such that it could be rendered in HMTL

PR #685, opened 2025-10-25
Added RNO and recurrent_layers from PR 248, updated to be compatible with upstream

rwhan:
Add RNO model with FNO-compatible parameter interface

- Implement RNO (Recurrent Neural Operator) with parameter alignment to FNO
- Add recurrent_layers.py with RNO_cell and RNO_layer components  

Notable updates from PR 248:
- Implement lifting_channel_ratio and projection_channel_ratio in RNO (previously lifting_channels and projection_channels)
- Add resolution_scaling_factor (replacing output_scaling_factor) for super-resolution
- Update domain padding to match FNO implementation
- Add comprehensive parameter documentation matching FNO style
- Update test suite to use new parameter interface

### DIFF
--- a/neuralop/models/rno.py
+++ b/neuralop/models/rno.py
@@ -345,7 +345,13 @@ class RNO(BaseModel, name='RNO'):
         if self.complex_data:
             self.projection = ComplexValued(self.projection)
 
-    def forward(self, x, init_hidden_states=None, return_hidden_states=False):
+    def forward(
+        self,
+        x,
+        init_hidden_states=None,
+        return_hidden_states=False,
+        keep_states_padded=False,
+    ):
         """
         Forward pass for the Recurrent Neural Operator.
 
@@ -360,31 +366,23 @@ class RNO(BaseModel, name='RNO'):
             shape (batch, hidden_channels, *spatial_dims_h), where spatial_dims_h are
             the unpadded spatial dimensions at each layer (accounting for resolution
             scaling). If domain_padding is enabled, hidden states will be automatically
-            padded like the input x. If None, all hidden states are initialized internally.
+            padded like the input x, unless keep_states_padded is True.
+            If None, all hidden states are initialized internally.
         return_hidden_states : bool, optional
             Whether to return the final hidden states for each layer in addition to
             the prediction. Default: False
+        keep_states_padded : bool, optional
+            If True, treats provided init_hidden_states as already padded (no additional
+            padding applied) and returns final_hidden_states in padded form (no unpadding).
+            Use this for efficient autoregressive loops to preserve boundary information.
+            Default: False
 
         Returns
         -------
         pred : torch.Tensor
-            Output tensor with shape (batch, out_channels, *spatial_dims_out). If
-            resolution scaling is used across layers, spatial_dims_out are scaled by
-            the product of the per-layer factors; otherwise they match the input
-            spatial dimensions.
+            Output tensor with shape (batch, out_channels, *spatial_dims_out).
         final_hidden_states : list[torch.Tensor], optional
-            Returned only if return_hidden_states=True. List of final hidden states for
-            each layer in unpadded format (matching the format of init_hidden_states input).
-            For layers 0..n_layers-2 these are the last-time-step states with shape
-            (batch, hidden_channels, *spatial_dims_out_i). The last entry corresponds to
-            the final hidden representation before the projection layer.
-
-        Notes
-        -----
-        Due to resolution invariance of spectral convolutions, tensors with incorrectly
-        ordered dimensions may "work" without raising low-level errors yet produce
-        nonsensical outputs. To guard against this, this method performs strict checks
-        on the input rank and the channel dimension position.
+            Returned only if return_hidden_states=True.
         """
         # Strict input validation to avoid silent errors from resolution invariance
         expected_rank = 3 + self.n_dim
@@ -405,9 +403,8 @@ class RNO(BaseModel, name='RNO'):
         if init_hidden_states is None:
             init_hidden_states = [None] * self.n_layers
         else:
-            # If domain padding is enabled, pad the provided hidden states
-            # This ensures consistency with how x is handled (user provides unpadded, we pad internally)
-            if self.domain_padding:
+            # If domain padding is enabled AND states are not already padded (keep_states_padded=False), pad them.
+            if self.domain_padding and not keep_states_padded:
                 padded_hidden_states = []
                 for h in init_hidden_states:
                     if h is not None:
@@ -447,13 +444,14 @@ class RNO(BaseModel, name='RNO'):
 
         if self.domain_padding:
             # DomainPadding.unpad expects (batch, channels, *spatial_dims)
-            # h should already be in this format from the last layer
+            # h comes from the last layer, which is padded logic.
+
+            # We ALWAYS unpad 'h' for the projection layer to produce the correct output size
+            # (Computation continues in unpadded space for projection)
             h = self.domain_padding.unpad(h)
 
-            # Also unpad all final_hidden_states to return them in unpadded format
-            # This maintains consistency: user provides unpadded, receives unpadded
-            # This prevents double-padding when states are passed back (e.g., in predict())
-            if return_hidden_states:
+            # For final_hidden_states, we unpad UNLESS explicitly told to keep them padded
+            if return_hidden_states and not keep_states_padded:
                 unpadded_states = []
                 for state in final_hidden_states:
                     unpadded_states.append(self.domain_padding.unpad(state))
@@ -502,11 +500,19 @@ class RNO(BaseModel, name='RNO'):
         states = [None] * self.n_layers
 
         for _ in range(num_steps):
-            pred, states = self.forward(x, states, return_hidden_states=True)
+            pred, states = self.forward(
+                x, 
+                states, 
+                return_hidden_states=True, 
+                keep_states_padded=True
+            )
+            
             output.append(pred)
             x = pred.reshape((pred.shape[0], 1, *pred.shape[1:]))
             if grid_function:
-                grid = grid_function((x.shape[0], x.shape[1], 1, x.shape[-2], x.shape[-1]), x.device)
+                grid = grid_function(
+                    (x.shape[0], x.shape[1], 1, x.shape[-2], x.shape[-1]), x.device
+                )
                 x = torch.cat((x, grid), dim=2)
 
         return torch.stack(output, dim=1)


rwhan:
@vduruiss @mliuschi

config/burgers_rno_config.py
vduruiss, line 1: I don't think you need to specify the values of the arguments which are the default ones in the classes

neuralop/models/tests/test_rno.py
```diff
@@ -0,0 +1,265 @@
...
+    factor = prod(resolution_scaling_factor)
+
+    assert list(out.shape) == [batch_size, 1] + [int(round(factor * s)) for s in size]
+
+
+@pytest.mark.parametrize("norm", [None, "group_norm", "instance_norm"])
+@pytest.mark.parametrize("complex_data", [False, True])
+@pytest.mark.parametrize("use_channel_mlp", [True, False])
+@pytest.mark.parametrize("channel_mlp_skip", ["linear", "identity", "soft-gating", None])
+@pytest.mark.parametrize("fno_skip", ["linear", "identity", "soft-gating", None])
```
vduruiss: Might be good to split it in 2 tests. Your version is more comprehensive, but it might take a lot of time to run the ~200 combinations in this test

neuralop/models/tests/test_rno.py
```diff
@@ -0,0 +1,265 @@
...
+
+    # Test forward pass
+    out, _ = model(in_data)
+
+    # Check output size
+    assert list(out.shape) == [batch_size, 1, *size]
+
+
+@pytest.mark.parametrize("channel_mlp_dropout", [0.0, 0.1, 0.5])
+@pytest.mark.parametrize("channel_mlp_expansion", [0.25, 0.5, 1.0])
```
vduruiss: 2 values are probably enough for both of these

neuralop/models/tests/test_rno.py
```diff
@@ -0,0 +1,265 @@
...
+    # Check output size
+    factor = prod(resolution_scaling_factor)
+
+    assert list(out.shape) == [batch_size, 1] + [int(round(factor * s)) for s in size]
+
+
+@pytest.mark.parametrize("norm", [None, "group_norm", "instance_norm"])
+@pytest.mark.parametrize("complex_data", [False, True])
+@pytest.mark.parametrize("use_channel_mlp", [True, False])
+@pytest.mark.parametrize("channel_mlp_skip", ["linear", "identity", "soft-gating", None])
```
vduruiss: When `use_channel_mlp=False`, the `channel_mlp_skip` is not used, so you are running the same tests 4 times

scripts/train_burgers_rno.py
```diff
@@ -0,0 +1,220 @@
...
+from neuralop.losses import LpLoss
+from neuralop.models import get_model
+from neuralop.data.datasets import load_mini_burgers_1dtime
+
+
+class RNOTimeAdapter(nn.Module):
+    """Wraps an RNO to accept Burgers 1D-time batches from datasets.
+
+    Expects inputs with shape (B, C, T, X). Permutes to (B, T, C, X) for RNO.
+    Returns prediction of the last time step with shape (B, C, X).
```
vduruiss: Maybe clarify this docstring. it's not very clear why it's needed and how it works

scripts/train_burgers_rno.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+
+    def __init__(self, core: nn.Module):
+        super().__init__()
+        self.core = core
```
vduruiss, line 62: What is `core` ? I imagine that's the trained model? Probably worth clarifying/defining it in the docstring

scripts/train_burgers_rno.py
```diff
@@ -0,0 +1,220 @@
...
+
+from config.burgers_rno_config import Default
+from neuralop.utils import get_wandb_api_key, count_model_params
+from neuralop.training import AdamW, Trainer
+from neuralop.losses import LpLoss
+from neuralop.models import get_model
+from neuralop.data.datasets import load_mini_burgers_1dtime
+
+
+class RNOTimeAdapter(nn.Module):
```
vduruiss, line 31: Explain why is this class needed. I imagine it's necessary because of the way the `Trainer` is setup?

Otherwise, it seems like all it's doing is permuting the data and evaluating the model

scripts/train_burgers_rno.py
```diff
@@ -0,0 +1,220 @@
...
+        # (B, C, T, X) -> (B, T, C, X)
+        x_rno = x.permute(0, 2, 1, 3).contiguous()
+        pred, _ = self.core(x_rno)
+        return pred
+
+
+class LastFrameDataProcessor(nn.Module):
+    """Thin wrapper to adapt y -> last frame for L2 supervision.
+
+    Delegates to an underlying data_processor; slices y to y[:, :, -1, :].
```
vduruiss: Clarify docstring

scripts/train_burgers_rno.py
```diff
@@ -0,0 +1,220 @@
...
+    else:
+        wandb_init_args = None
+
+    # Print configuration details
+    if config.verbose:
+        print("##### CONFIG ######")
+        print(config)
+        sys.stdout.flush()
+
+    # Data loading: resolve dataset path relative to the repository root
```
vduruiss:
```suggestion
    # Data loading
```

neuralop/layers/rno_block.py
vduruiss, line 1: Maybe we can call this file `rno_block.py` ?

neuralop/layers/recurrent_layers.py
```diff
@@ -0,0 +1,315 @@
...
+            "factorization": factorization,
+            "rank": rank,
+            "conv_module": conv_module,
+            "fixed_rank_modes": fixed_rank_modes,
+            "implementation": implementation,
+            "decomposition_kwargs": decomposition_kwargs,
+        }
+
+        # Some resolution_scaling_factors are None to super-resolution purposes. We use the hidden representation in the scaled size always (it's initialized that way),
+        # so we only need to scale the dimensions of f1, f3, and f5, which act on x (original dimensionality).
```
vduruiss: Let's clarify this, I didn't understand what that meant

neuralop/layers/rno_block.py
```diff
@@ -0,0 +1,315 @@
...
+        complex_data=False,
+        separable=False,
+        factorization=None,
+        rank=1.0,
+        conv_module=SpectralConv,
+        fixed_rank_modes=False,
+        implementation="factorized",
+        decomposition_kwargs=dict(),
+    ):
+        # resolution_scaling_factor is provided here as an integer or float
```
vduruiss:
```suggestion

```

neuralop/layers/recurrent_layers.py
vduruiss: Looking at the forward I understood, but we can clarify the comment still.

So the hidden states need to always be passed at the output resolution of the cell?

Might be good to clarify that in the forward docstring

neuralop/layers/rno_block.py
```diff
@@ -0,0 +1,315 @@
...
+        if complex_data:
+            self.b1 = nn.Parameter(torch.normal(torch.tensor(0.), torch.tensor(1.)) + 1j * torch.normal(torch.tensor(0.), torch.tensor(1.))) # constant bias terms
+            self.b2 = nn.Parameter(torch.normal(torch.tensor(0.), torch.tensor(1.)) + 1j * torch.normal(torch.tensor(0.), torch.tensor(1.)))
+            self.b3 = nn.Parameter(torch.normal(torch.tensor(0.), torch.tensor(1.)) + 1j * torch.normal(torch.tensor(0.), torch.tensor(1.)))
+        else:
+            self.b1 = nn.Parameter(torch.normal(torch.tensor(0.), torch.tensor(1.))) # constant bias terms
+            self.b2 = nn.Parameter(torch.normal(torch.tensor(0.), torch.tensor(1.)))
+            self.b3 = nn.Parameter(torch.normal(torch.tensor(0.), torch.tensor(1.)))
+    
+    def forward(self, x, h):
```
vduruiss, line 174: Add a comment or docstring to specify what the expected shapes are for x and h

neuralop/layers/recurrent_layers.py
```diff
@@ -0,0 +1,315 @@
...
+    Parameters
+    ------------------
+    n_modes : int tuple
+        number of modes to keep in Fourier Layer, along each dimension
+        The dimensionality of the RNO is inferred from ``len(n_modes)``
+    width : int
+        width of the RNO (i.e. number of channels)
+    return_sequences : boolean, optional
+        Whether to return the sequence of hidden states associated with processing
+        the inputs sequence of functions.
```
vduruiss: Add that default is False

neuralop/layers/recurrent_layers.py
```diff
@@ -0,0 +1,315 @@
...
+        if x.dtype == torch.cfloat:
+            h_hat = cselu(h_combined)  # complex SELU for complex data
+        else:
+            h_hat = F.selu(h_combined)  # regular SELU for real data
+
+        h_next = (1. - z) * h + z * h_hat
+
+        return h_next
+
+class RNO_layer(nn.Module):
```
vduruiss: Maybe we call this `RNOBlock` ?

neuralop/layers/recurrent_layers.py
```diff
@@ -0,0 +1,315 @@
...
+    """N-Dimensional Recurrent Neural Operator layer. The RNO layer extends the
+    action of the RNO cell to take in some sequence of time-steps as input
+    and output the next output function. 
+
+    Paper: https://arxiv.org/abs/2308.08794
```
vduruiss: Would be helpful to write a little bit more. Maybe a quick mention of the formulas. Sometimes when we maintain codes, we update the codes and then it's not exactly the same as in the original paper. Or sometimes we update the papers. So it can be helpful to also have the equations for the code in the docstring

neuralop/layers/recurrent_layers.py
```diff
@@ -0,0 +1,315 @@
...
+
+        # Some resolution_scaling_factors are None to super-resolution purposes. We use the hidden representation in the scaled size always (it's initialized that way),
+        # so we only need to scale the dimensions of f1, f3, and f5, which act on x (original dimensionality).
+        self.f1 = FNOBlocks(width, width, n_modes, resolution_scaling_factor=scaling_factor, 
+                            **fno_kwargs)
+        self.f2 = FNOBlocks(width, width, n_modes, resolution_scaling_factor=None, **fno_kwargs)
+        self.f3 = FNOBlocks(width, width, n_modes, resolution_scaling_factor=scaling_factor, **fno_kwargs)
+        self.f4 = FNOBlocks(width, width, n_modes, resolution_scaling_factor=None, **fno_kwargs)
+        self.f5 = FNOBlocks(width, width, n_modes, resolution_scaling_factor=scaling_factor, **fno_kwargs)
+        self.f6 = FNOBlocks(width, width, n_modes, resolution_scaling_factor=None, **fno_kwargs)
```
vduruiss: Let's write thise with the argument names so it's easier to follow, 

so FNOBlocks(arg1=arg1value, arg2-arg2value, ...)

neuralop/layers/recurrent_layers.py
```diff
@@ -0,0 +1,315 @@
...
+        Fixed_rank_modes parameter for SpectralConv, by default False
+    implementation : str, optional
+        Implementation parameter for SpectralConv. Options: "factorized", "reconstructed", by default "factorized"
+    decomposition_kwargs : dict, optional
+        Kwargs for tensor decomposition in SpectralConv, by default dict()
+    """
+    def __init__(
+        self, 
+        n_modes, 
+        width,
```
vduruiss: I think we should switch that to `hidden_channels` to be more consistent with the other codes from the library

neuralop/layers/recurrent_layers.py
```diff
@@ -0,0 +1,315 @@
...
+    and output the next output function. 
+
+    Paper: https://arxiv.org/abs/2308.08794 
+
+    Parameters
+    ------------------
+    n_modes : int tuple
+        number of modes to keep in Fourier Layer, along each dimension
+        The dimensionality of the RNO is inferred from ``len(n_modes)``
+    width : int
```
vduruiss: I think we should switch that to `hidden_channels` to be more consistent with the other codes from the library

neuralop/layers/rno_block.py
```diff
@@ -0,0 +1,315 @@
...
+            fixed_rank_modes=fixed_rank_modes,
+            implementation=implementation,
+            decomposition_kwargs=decomposition_kwargs,
+        )
+        if complex_data:
+            self.bias_h = nn.Parameter(torch.normal(torch.tensor(0.), torch.tensor(1.)) + 1j * torch.normal(torch.tensor(0.), torch.tensor(1.)))
+        else:
+            self.bias_h = nn.Parameter(torch.normal(torch.tensor(0.), torch.tensor(1.)))
+
+    def forward(self, x, h=None):
```
vduruiss, line 351: Add comment or docstring about expected shapes

neuralop/layers/rno_block.py
```diff
@@ -0,0 +1,315 @@
...
+        outputs = []
+        for i in range(timesteps):
+            h = self.cell(x[:, i], h)
+            if self.return_sequences:
+                outputs.append(h)
+
+        if self.return_sequences:
+            return torch.stack(outputs, dim=1)
+        else:
+            return h
```
vduruiss: Add one or two small in-line comments

neuralop/models/rno.py
```diff
@@ -0,0 +1,468 @@
...
+        n_modes must be larger enough but smaller than max_resolution//2 (Nyquist frequency)
+    in_channels : int
+        Number of input channels in input function. Determined by the problem.
+    out_channels : int
+        Number of output channels in output function. Determined by the problem.
+    hidden_channels : int
+        Width of the RNO (i.e. number of channels).
+        This significantly affects the number of parameters of the RNO.
+        Good starting point can be 64, and then increased if more expressivity is needed.
+        Update lifting_channel_ratio and projection_channel_ratio accordingly since they are proportional to hidden_channels.
```
vduruiss, line 63: 64 as a recommendation might be large as a default, especially for 2D, 3D if you don't use TFNO blocks, no?

neuralop/models/rno.py
```diff
@@ -0,0 +1,468 @@
...
+    n_layers : int, optional
+        Number of RNO layers to use. Default: 4
+    residual : bool
+        Whether to use residual connections between RNO layers.
```
vduruiss: Might be good to mention more precisely where the residual connection is added. 

Also we have been using `_skip` for the names of other skip connections, so maybe we rename it as `rno_skip` ?

vduruiss: Also mention default value of False. 

Out of curiosity, any reason why it is set to False by default?

The other skip connections in the FNO blocks are active by default

neuralop/models/rno.py
```diff
@@ -0,0 +1,468 @@
...
+from ..layers.recurrent_layers import RNO_layer
+from ..layers.padding import DomainPadding
+from ..layers.fno_block import FNOBlocks
+from ..layers.channel_mlp import ChannelMLP
+from ..layers.spectral_convolution import SpectralConv
+from ..layers.complex import ComplexValued
+from ..layers.embeddings import GridEmbeddingND, GridEmbedding2D
+from .base_model import BaseModel
+
+class RNO(BaseModel, name='RNO'):
```
vduruiss, line 23: Given that so much of its structure is identical the the FNO class, I wonder if it might not be better to have it as a subclass of the FNO class

neuralop/models/rno.py
```diff
@@ -0,0 +1,468 @@
...
+                fno_block_precision=fno_block_precision,
+                rank=rank,
+                fixed_rank_modes=fixed_rank_modes,
+                implementation=implementation,
+                separable=separable,
+                factorization=factorization,
+                decomposition_kwargs=decomposition_kwargs,
+                conv_module=conv_module,
+            )
+        for i in range(n_layers - 1)]
```
vduruiss: Let's do that more cleanly. Let's pass the `return_sequence` as a list of booleans (True/False for each hidden state), so we don't need to separate the instantiations of the RNOBlocks

neuralop/models/rno.py
```diff
@@ -0,0 +1,468 @@
...
+            out_channels=out_channels,
+            hidden_channels=self.projection_channels,
+            n_layers=2,
+            n_dim=self.n_dim,
+            non_linearity=non_linearity,
+        )
+        if self.complex_data:
+            self.projection = ComplexValued(self.projection)
+
+    def forward(self, x, init_hidden_states=None): # h must be padded if using padding
```
vduruiss: Why? Could you not also pass `h` through the domain_padding function just like `x` ?

neuralop/models/rno.py
```diff
@@ -0,0 +1,468 @@
...
+            raise ValueError(
+                f"RNO.forward expected input of rank {expected_rank} = "
+                f"(batch, timesteps, channels, {self.n_dim} spatial dims), got rank {x.ndim} with shape {tuple(x.shape)}"
+            )
+        if x.shape[2] != self.in_channels:
+            raise ValueError(
+                f"RNO.forward expected x.shape[2] == in_channels ({self.in_channels}); "
+                f"got {x.shape[2]}. Input must be shaped as (batch, timesteps, in_channels, *spatial_dims)."
+            )
+        # x shape (batch, timesteps, dim, dom_size1, dom_size2, ..., dom_sizen)
```
vduruiss:
```suggestion

```

neuralop/models/rno.py
```diff
@@ -0,0 +1,468 @@
...
+            # DomainPadding.unpad expects (batch, channels, *spatial_dims)
+            # h should already be in this format from the last layer
+            h = self.domain_padding.unpad(h)
+
+        pred = self.projection(h)
+
+        return pred, final_hidden_states
+
+    def predict(self, x, num_steps, grid_function=None): # num_steps is the number of steps ahead to predict
+        # grid_function is assumed to take in a shape and a device and return the grid
```
vduruiss: Add more detailed docstring

neuralop/models/rno.py
```diff
@@ -0,0 +1,468 @@
...
+        pred : torch.Tensor
+            Output tensor with shape (batch, out_channels, *spatial_dims_out). If
+            resolution scaling is used across layers, spatial_dims_out are scaled by
+            the product of the per-layer factors; otherwise they match the input
+            spatial dimensions.
+        final_hidden_states : list[torch.Tensor]
+            List of final hidden states for each layer. For layers 0..n_layers-2 these
+            are the last-time-step states with shape (batch, hidden_channels, *spatial_dims_out_i).
+            The last entry corresponds to the final hidden representation before the
+            projection layer.
```
vduruiss: Do we really need to return them every time? Maybe we can have that as a quick boolean for the forward that is False by default?

neuralop/models/rno.py
```diff
@@ -0,0 +1,468 @@
...
+        x = self.lifting(x)
+        x = x.reshape(batch_size, timesteps, *x.shape[1:])
+
+        if self.domain_padding:
+            # DomainPadding expects (batch, channels, *spatial_dims), so reshape to remove timestep dim
+            x = x.reshape(batch_size * timesteps, *x.shape[2:])
+            x = self.domain_padding.pad(x)
+            x = x.reshape(batch_size, timesteps, *x.shape[1:])
+
+        final_hidden_states = []
```
vduruiss, line 431: See other comment on whether we really need to save those every time. Could have it false by default

neuralop/models/rno.py
```diff
@@ -0,0 +1,468 @@
...
+        if self.domain_padding:
+            # DomainPadding.unpad expects (batch, channels, *spatial_dims)
+            # h should already be in this format from the last layer
+            h = self.domain_padding.unpad(h)
+
+        pred = self.projection(h)
+
+        return pred, final_hidden_states
+
+    def predict(self, x, num_steps, grid_function=None): # num_steps is the number of steps ahead to predict
```
vduruiss: Did not really understand what grid_function is

Is this positional embedding?

neuralop/models/rno.py
rwhan: Unlike FNO where skip connections improve gradient flow, RNO's recurrent structure already provides temporal gradient pathways. e.g. in the tipping point paper, the implementation used  fno_skip="linear" but rno_skip=False

neuralop/models/rno.py
rwhan: I considered making RNO a subclass of FNO, but I think keeping them separate is the better design choice here:

1. Semantic difference: RNO and FNO solve fundamentally different problems. FNO is a 
   feedforward operator, while RNO is recurrent. The "is-a" relationship doesn't really hold.

2. Different interfaces: 
   - RNO.forward expects (batch, timesteps, channels, *spatial) with time dimension
   - FNO.forward expects (batch, channels, *spatial) without time
   - RNO returns hidden states, FNO doesn't
   
3. Composition over inheritance* While they share components (lifting, projection, 
   domain padding), this is a classic case where composition is better than inheritance. 
   They use similar building blocks but are architecturally distinct.

4. Maintainability: Inheritance could create coupling issues - changes to FNO for 
   feedforward use cases might not make sense for RNO's recurrent use cases.

I think the current approach of sharing low-level components (FNOBlocks, SpectralConv, etc.) 
while keeping the models separate is the right balance.

neuralop/models/rno.py
rwhan: I considered automatic padding, but decided against it for these reasons:

1. Per-layer complexity: Each RNO layer can have different `resolution_scaling_factor`, 
   so each layer's hidden state would need different padding. 

2. User error masking: If a user provides wrong-sized hidden states, automatic padding 
   could silently "fix" it in unexpected ways rather than failing clearly.

That said, if you feel strongly about this or see a clean way to implement it, I'm open 
to revisiting. The current approach prioritizes explicitness over convenience for this 
edge case.

vduruiss:
@rwhan Can you also add RNO and RNOBlocks to the API reference file?

`neuraloperator/doc/source/modules/api.rst`

vduruiss:
Looks good to me, thank you @rwhan !

@JeanKossaifi if you want to have a look

doc/source/modules/api.rst
```diff
@@ -177,6 +190,23 @@ FNO Blocks
...
+.. automodule:: neuralop.layers.rno_block
+    :no-members:
+    :no-inherited-members:
+
+.. autosummary::
+    :toctree: generated
+    :template: class.rst
+
+    RNOBlock
+
```
JeanKossaifi, line 207: Pep8 issues, there should be no space between title and ~, and ~ should match the finish of the line.

neuralop/layers/rno_block.py
```diff
@@ -0,0 +1,404 @@
+import torch
+import numpy as np
+import torch.nn as nn
+import torch.nn.functional as F
+
+from ..layers.spectral_convolution import SpectralConv
+from ..layers.fno_block import FNOBlocks
+from ..layers.complex import cselu
+
+class RNO_cell(nn.Module):
```
JeanKossaifi: Should probably be RNOCell to not mix naming conventions

neuralop/layers/rno_block.py
```diff
@@ -0,0 +1,404 @@
...
+    The RNO cell implements the GRU-like recurrence relation with Fourier layers:
+        z_t = σ(f1(x_t) + f2(h_{t-1}) + b1)            [update gate]
+        r_t = σ(f3(x_t) + f4(h_{t-1}) + b2)            [reset gate]
+        h̃_t = selu(f5(x_t) + f6(r_t ⊙ h_{t-1}) + b3)  [candidate state]
+        h_t = (1 - z_t) ⊙ h_{t-1} + z_t ⊙ h̃_t         [next state]
+
+    where σ is the sigmoid function, ⊙ is element-wise multiplication,
+    and f1-f6 are FNO blocks (Fourier neural operators).
+
+    Paper: https://arxiv.org/abs/2308.08794
```
JeanKossaifi: Let's make this a reference in a reference section in the doctoring.

neuralop/layers/rno_block.py
```diff
@@ -0,0 +1,404 @@
...
+
+    Parameters
+    ------------------
```
JeanKossaifi: Lines lengths should match

neuralop/layers/rno_block.py
```diff
@@ -0,0 +1,404 @@
...
+            resolution_scaling_factor=scaling_factor,
+            **fno_kwargs
+        )
+        self.f6 = FNOBlocks(
+            in_channels=hidden_channels,
+            out_channels=hidden_channels,
+            n_modes=n_modes,
+            resolution_scaling_factor=None,
+            **fno_kwargs
+        )
```
JeanKossaifi: This could be a ModuleList and created in a for loop instead of hard coding.

neuralop/layers/rno_block.py
```diff
@@ -0,0 +1,404 @@
...
+
+        if complex_data:
+            self.b1 = nn.Parameter(torch.normal(torch.tensor(0.), torch.tensor(1.)) + 1j * torch.normal(torch.tensor(0.), torch.tensor(1.))) # constant bias terms
+            self.b2 = nn.Parameter(torch.normal(torch.tensor(0.), torch.tensor(1.)) + 1j * torch.normal(torch.tensor(0.), torch.tensor(1.)))
+            self.b3 = nn.Parameter(torch.normal(torch.tensor(0.), torch.tensor(1.)) + 1j * torch.normal(torch.tensor(0.), torch.tensor(1.)))
+        else:
+            self.b1 = nn.Parameter(torch.normal(torch.tensor(0.), torch.tensor(1.))) # constant bias terms
+            self.b2 = nn.Parameter(torch.normal(torch.tensor(0.), torch.tensor(1.)))
+            self.b3 = nn.Parameter(torch.normal(torch.tensor(0.), torch.tensor(1.)))
```
JeanKossaifi: Same here - it actually obfuscates what is going on a little

neuralop/layers/rno_block.py
```diff
@@ -0,0 +1,404 @@
...
+        h : torch.Tensor
+            Hidden state from previous timestep with shape (batch, hidden_channels, *spatial_dims_h)
+            where spatial_dims_h may differ from spatial_dims if resolution_scaling_factor is set
+
+        Returns
+        -------
+        torch.Tensor
+            Updated hidden state with shape (batch, hidden_channels, *spatial_dims_h)
+        """
+        z = torch.sigmoid(self.f1(x) + self.f2(h) + self.b1)
```
JeanKossaifi: Can we name the variables explicitly instead of single letters? e.g. `hidden_state` ? h_hat, r, z isn't super helpful for someone trying to understand the code or the method.

PR #702, opened 2026-01-22
Enforce Hermitian symmetry conditions before IFFT in SpectralConv

vduruiss:
Break up the irfftn into its ifftn and irfft components and zero out appropriate imaginary parts in between

This is to fix the line artifacts that are sometimes observed at powers of 2 on certain GPUs

@JeanKossaifi If you have implementation efficiency concerns/suggestions

### DIFF
--- a/neuralop/models/tests/test_sfno.py
+++ b/neuralop/models/tests/test_sfno.py
@@ -1,0 +1,93 @@
+"""Quick tests for SFNO (Spherical FNO). SFNO is FNO with SphericalConv; 2D only."""
+import pytest
+import torch
+
+try:
+    from neuralop.models import SFNO
+except ImportError:
+    SFNO = None
+
+@pytest.mark.skipif(SFNO is None, reason="SFNO not available (torch_harmonics required)")
+@pytest.mark.parametrize("n_modes", [(4, 4), (5, 5), (4, 6), (5, 6)])
+@pytest.mark.parametrize(
+    "implementation,factorization,resolution_scaling_factor,rank,fixed_rank_modes",
+    [
+        ("factorized", "dense", None, 0.3, False),
+        ("reconstructed", "cp", 1.0, 0.5, False),
+        ("factorized", "tucker", None, 0.5, True),
+        ("reconstructed", "tt", 1.0, 0.2, True),
+    ],
+)
+def test_sfno_forward(
+    n_modes, implementation, factorization, resolution_scaling_factor, rank, fixed_rank_modes
+):
+    """SFNO forward over n_modes, implementation, factorization, resolution scaling, rank, fixed_rank_modes."""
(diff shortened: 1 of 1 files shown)


neuralop/layers/fno_block.py
```diff
@@ -153,31 +161,40 @@ def __init__(
...
+                implementation=implementation,
+                separable=separable,
+                factorization=factorization,
+                fno_block_precision=fno_block_precision,
+                decomposition_kwargs=decomposition_kwargs,
+                complex_data=complex_data,
+            )
+            if isinstance(conv_module, type) and issubclass(conv_module, SpectralConv):
+                kwargs["enforce_hermitian_symmetry"] = enforce_hermitian_symmetry
+            return kwargs
```
JeanKossaifi: I'm not sure I see a need for this function? Can we not assume conv_module *is* a module, i.e. a class? Why not just always pass enforce_hermitian_symmetry?

neuralop/layers/local_no_block.py
```diff
@@ -257,20 +266,28 @@ def __init__(
...
+                rank=rank,
+                fixed_rank_modes=fixed_rank_modes,
+                implementation=implementation,
+                separable=separable,
+                factorization=factorization,
+                decomposition_kwargs=decomposition_kwargs,
+            )
+            if isinstance(conv_module, type) and issubclass(conv_module, SpectralConv):
+                kwargs["enforce_hermitian_symmetry"] = enforce_hermitian_symmetry
+            return kwargs
```
JeanKossaifi: Same here, i'd just always pass it and let the child class just ignore it if not used? 

Or if we really want that either:
i) define a dict and appedn the new key 
ii) **({'enforce_hermitian_symmetry': value} if condition else {})

neuralop/layers/spectral_convolution.py
```diff
@@ -522,17 +533,37 @@ def forward(self, x: torch.Tensor, output_shape: Optional[Tuple[int]] = None):
...
+            # and as a result causes line artifacts.  
+            # To fix this, we split the ifftn into a ifftn in (n-1) dimensions and a irfft in the last dimension,
+            # although it incurs a small additional computational cost.
+            
+            if self.enforce_hermitian_symmetry:
+                out_fft = torch.fft.ifftn(out_fft, s=mode_sizes[:-1], dim=fft_dims[:-1], norm=self.fft_norm)
+                # Enforce Hermitian symmetry conditions for irfft
+                # 0th frequency must be real
+                out_fft[..., 0].imag.zero_()
+                # Nyquist frequency must be real if the spatial size is even
```
JeanKossaifi, line 560: nitpick - can we leave a blank line before the comments so it is clear what line they apply to? It's fairly clear if you read the code but doesn't cost much to be extra clear :)

PR #715, opened 2026-03-16
feat: add norm_groups for GroupNorm in FNO

MauriceDHanisch:
This PR lets you set the number of groups for `GroupNorm` in the `FNO` model. Before, it was hardcoded to 1.

### Changes
- **`FNOBlocks`**: Added `norm_groups` to the constructor and passed it to `nn.GroupNorm`.
- **`FNO` model**: Added `norm_groups` so it can be set from the main model.
- **Tests**: Added tests in `test_fno_block.py` and `test_fno.py` to check that the layers are set up correctly.

*Note: This PR only updates the `FNO` model. `LocalNO` and `UNO` are not changed.*


vduruiss:
Looks good to me @MauriceDHanisch, thank you!

Could you propagate the same changes to the following files, with a small accompanying test? 

```
otno.py 
rno.py
fnogno.py
gino.py
local_no.py
uno.py
local_no_block.py
rno_block.py
```

I don't anticipate there would be any difference in the way it needs to be implemented for these other files.

MauriceDHanisch:
I added it everywhere + added corresponding tests

PR #718, opened 2026-03-27
Fix shape comparison bug in SpectralConv.transform

ehsanngh:
I believe there is a bug in SpectralConv.transform where in_shape is created as a list while out_shape is a tuple, causing the equality check to always fail.


JeanKossaifi:
Thanks @ehsanngh, good catch - perhaps we should then also add a test to cast the output_shape to tuple if it isn't one already?

PR #723, opened 2026-05-10
Temporal embed

yashpincha:
## Summary

Adds time-conditioned FNO variants by extending FNO, SpectralConv, SphericalConv, and FNOBlocks with three optional dict kwargs: embed, mode_modulation, norm_modulation. The default FNO behavior is completely unchanged -- all new kwargs default to None.

### Architecture

Time conditioning is implemented through two independent, optional pathways:

1. **Mode modulation** (`mode_modulation` dict) — a small MLP maps a learned embedding of `(t, k)` to a per-mode multiplier applied to spectral coefficients before the convolution contraction. Supports `real`, `complex`, and `polar` output forms.

2. **Norm modulation** (`norm_modulation` dict) — a per-layer MLP of an embedding of `t` produces AdaIN-style scale/shift parameters and sigmoid gates for the two residual sums in each FNO block (after the spectral branch and after the ChannelMLP branch). Each of the four flags (`modulate1`, `modulate1_gate`, `modulate2`, `modulate2_gate`) is independently toggleable.

Both pathways share a common time/mode embedding (`embed` dict) configurable as sinusoidal or power-law for both `t` and `k`.

**Models (`neuralop.models`):**
- `t_emb_FNO` — FNO with `time_conditioned=True`
- `t_emb_TFNO` — same with Tucker factorization defaults
- `t_emb_SFNO` — spherical variant using `SphericalConvModulated` (requires `torch-harmonics`)

**Layers (`neuralop.layers`):**
- `SpectralConv` — extended with (t, k) mode modulation via embed + mode_modulation kwargs
- `SphericalConv`— same, with additional share_m / pre_modulate flags
- `FNOBlocks` — threads t through convs and adds AdaGN-style norm +  modulation via the norm_modulation kwarg

Enabling time conditioning on `FNO` directly:

```python
model = FNO(
    n_modes=(12, 12), in_channels=1, out_channels=1, hidden_channels=64,
    n_params=1,
    embed={"type_t": "sinusoidal", "type_k": "power", "dim": 32},
    mode_modulation={"enabled": True, "type": "polar", "hidden_channels": 64},
    norm_modulation={"enabled": True, "modulate1": True, "modulate1_gate": True,
                     "modulate2": False, "modulate2_gate": False},
)
out = model(x, t=0.5)          # t: scalar, 0-d tensor, (B,), (B, 1), or (B, n_params)
```

### Backward compatibility

- `FNO(...)` with no new kwargs constructs the exact same module tree as before. `time_conditioned` is only True if modulator dicts have been passed.
- `forward(x)` (no `t`) continues to work unchanged.
- `TFNO`, `SFNO`, and all existing classes are unaffected.
- `state_dict().keys()` of a default `FNO` is identical before and after this PR (asserted in `test_fno_default_unchanged_no_modulated_modules`).


cwwangcal (approved):

neuralop/layers/spectral_convolution.py
```diff
@@ -18,6 +19,34 @@
 einsum_symbols = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
 
 
+class _ContiguousBackward(torch.autograd.Function):
+    """Identity in the forward pass; returns a contiguous gradient in backward.
+
+    PyTorch's native FFT autograd dispatches the transform's backward (an
+    ``_fft_r2c`` / ``_fft_c2r`` / ``_fft_c2c`` ATen op) on whatever gradient
+    tensor it receives. On x86 CPU with Intel oneMKL, a gradient with
```
JeanKossaifi, line 27: Do we really need this? I don't think we ever want to train FNOs on CPU?

JeanKossaifi (changes requested):
Thanks @yashpincha - this is a lot of changes in the core blocks, with very specific changes. Could we try to revisit the API and make it more flexible? Does the embedding need to be done in the spectral cone? Could it be done in the FNO Block only? Could we pass the processed embedding directly? Or even a preprocessed input `x` directly?

Do you have a simple writeup / summary of the main changes you make, in mathematical terms?

PR #725, opened 2026-05-31
fix: align GINOConfig parameter names with GINO.__init__ signature

harshaa765:
## Summary

`GINOConfig` (and its subclass `GINO_Small3d`) defined four config fields whose names do not match any parameter in `GINO.__init__`. This caused an immediate `TypeError` when running `scripts/train_gino_carcfd.py`, because `get_model` passes all config fields as `**kwargs` directly to the model constructor.

**Changes in `config/models.py`:**

| Old field (config) | Correct field (GINO `__init__`) |
|---|---|
| `gno_coord_embed_dim` | `gno_embed_channels` |
| `gno_radius` | `in_gno_radius` + `out_gno_radius` (split) |
| `gno_pos_embed_type` | `in_gno_pos_embed_type` + `out_gno_pos_embed_type` (split) |
| `fno_domain_padding` | *(removed — not a GINO parameter)* |

The same stale `fno_domain_padding` field is also removed from `GINO_Poisson2d`, which inherits from `GINOConfig`.

Note: `GINO_Poisson2d` already used the correct split names (`in_gno_radius`, `out_gno_radius`, `in_gno_pos_embed_type`, `out_gno_pos_embed_type`, `gno_embed_channels`) as its own overrides — those are preserved unchanged.

## Test plan

- [x] Ran `scripts/train_gino_carcfd.py` — model now instantiates and trains without error (verified epoch 0 and epoch 1 with decreasing loss)
- [x] `black config/models.py` — no formatting changes needed

### DIFF
--- a/config/models.py
+++ b/config/models.py
@@ -149,11 +149,13 @@ class GINOConfig(ModelConfig):
     out_channels: int
     latent_feature_channels: Optional[int] = None
     gno_coord_dim: int = 3
-    gno_coord_embed_dim: int = 16
-    gno_radius: float = 0.033
+    gno_embed_channels: int = 16
+    in_gno_radius: float = 0.033
+    out_gno_radius: float = 0.033
     in_gno_transform_type: str = "linear"
     out_gno_transform_type: str = "linear"
-    gno_pos_embed_type: str = "nerf"
+    in_gno_pos_embed_type: str = "nerf"
+    out_gno_pos_embed_type: str = "nerf"
     gno_weighting_function: Optional[str] = None
     gno_weight_function_scale: Optional[float] = None
     fno_n_modes: List[int] = [16, 16, 16]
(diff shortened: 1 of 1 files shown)


JeanKossaifi:
Thanks for finding and fixing these issues @harshaa765! Was there a reason for changing the default from nerf to transformer? If not I'd leave it to not create a new surprising behavior for existing users. Otherwise it looks good to me!

harshaa765:
> Thanks for finding and fixing these issues @harshaa765! Was there a reason for changing the default from nerf to transformer? If not I'd leave it to not create a new surprising behavior for existing users. Otherwise it looks good to me!

Thanks @JeanKossaifi  for the review! You are right, changing the default wasn't intentional. I'll push an update a bit later to change it back to nerf. Glad you caught that!

harshaa765:
@JeanKossaifi reverted the default back to `nerf` for both `in_gno_pos_embed_type` and `out_gno_pos_embed_type` to preserve existing behavior, as you suggested. (Note: `GINO.__init__` itself defaults to `transformer`, but I've kept the config on `nerf` so there's no surprising change for existing users.) Also rebased onto the latest `main`. Thanks again for the review!

PR #727, opened 2026-06-06
fix(PTDataset): preserve input dtype instead of forcing float32

### LINKED ISSUE #724: Complex data processing
Hi! I've got confused using FNO for complex data.
I want to map real data to complex field, so I set FNO parameter "complex_data" to True.
When I try to train a model it I got an error "imag is not implemented for tensors with non-complex dtypes".
If I deliberately change the input dtype from "float32" to "complex64" the error is the same.
Please, explain, how to process complex data correctly.

Upd.
I use PTDataset to load data and as far as I understand the code the "x" vector is deliberately converted to "float32" (not "y" vector though).
Should I use smth else to load my data?

harshaa765:
## Summary

Closes #724.

`PTDataset` was unconditionally casting `x` tensors to `float32` on load:

```python
# before
x_train = data["x"].type(torch.float32).clone()
```

This broke `FNO` with `complex_data=True` because `ComplexValued` layers (used by `FNOBlocks` when `complex_data=True`) access `.real` and `.imag` on the input tensor, which raises:

> `imag is not implemented for tensors with non-complex dtypes`

The fix adds an optional `dtype` parameter (default `None`) to `PTDataset.__init__`:

- `dtype=None` (default) — preserves the original dtype of the saved tensor; complex data stays complex
- `dtype=torch.float32` — explicit opt-in for callers that need the previous behaviour

The cast on `y` is untouched (it was never force-cast).

## Changes

- `neuralop/data/datasets/pt_dataset.py` — add `dtype=None` parameter; replace the two hard-coded `.type(torch.float32)` casts
- `neuralop/data/datasets/tests/test_ptdatasets.py` — three new unit tests covering: float32 preservation, complex64 preservation (the bug), and explicit float32 cast (backward compat)

## Test plan

- [x] `test_dtype_none_preserves_float32` — default path does not change float32 data
- [x] `test_dtype_none_preserves_complex64` — default path preserves complex64 (regression for #724)
- [x] `test_dtype_explicit_float32_casts_complex64` — `dtype=torch.float32` still casts as before
- [x] All three new tests pass locally
- [x] Existing subclasses (`DarcyDataset`, `NavierStokesDataset`, `Burgers1dTimeDataset`) unaffected — none pass `dtype`, so they receive `None` and behaviour is unchanged

### DIFF
--- a/neuralop/data/datasets/darcy.py
+++ b/neuralop/data/datasets/darcy.py
@@ -3,6 +3,7 @@ import os
 from pathlib import Path
 from typing import Union, List
 
+import torch
 from torch.utils.data import DataLoader
 
 from .pt_dataset import PTDataset
@@ -50,6 +51,11 @@ class DarcyDataset(PTDataset):
         dimension of saved tensors to index data channels, by default 1
     subsampling_rate : int or List[int], optional
         rate at which to subsample each input dimension, by default None
+    dtype : torch.dtype, optional
+        dtype to cast input tensors to after loading. The bundled Darcy
+        coefficient fields are stored as torch.bool, so this defaults to
+        torch.float32 to keep them usable by float models. Pass None to
+        preserve the saved dtype. Default is torch.float32.
     download : bool, optional
         whether to download data if not present, by default True
 
@@ -75,6 +81,7 @@ class DarcyDataset(PTDataset):
         encoding="channel-wise",
         channel_dim=1,
         subsampling_rate=None,
+        dtype: torch.dtype = torch.float32,
         download: bool = True,
     ):
         """Initialize the DarcyDataset.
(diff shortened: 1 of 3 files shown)


JeanKossaifi:
Thank you @harshaa765!
I don't think we need the additional unit-test here. Let's just leave dtype=None by default and not handle it as a special case: in that case `tensor.to(dtype)` becomes `tensor.to(None)` which acts as a no-op.

harshaa765:
> Thank you @harshaa765! I don't think we need the additional unit-test here. Let's just leave dtype=None by default and not handle it as a special case: in that case `tensor.to(dtype)` becomes `tensor.to(None)` which acts as a no-op.

Thanks for the review @JeanKossaifi! Good call. I've simplified both assignments to data["x"].to(dtype).clone() directly, letting tensor.to(None) act as the no-op. Removed the extra unit tests and their helper functions as well. Please take another look when you get a chance!

PR #730, opened 2026-06-22
Add tensorgrad optimizer

sebulo:
Add TensorGRaD optimizer and Darcy training example

## Summary
This PR adds TensorGRaD [1] support for memory-efficient neural operator training.

  Changes include:
  - Add `TensorGRaD` optimizer with low-rank and low-rank + sparse
  gradient updates.
  - Add `fno_tensorgrad_param_groups` helper for applying TensorGRaD to
  FNO spectral convolution weights.
  - Update the existing AdamW low-rank projection path to use the
  TensorGRaD projector instead of TensorGalore
  - Add optimizer tests for projection shape, projection refresh
  cadence, sparse projection, compressed optimizer state, and FNO
  parameter grouping.
  - Add a Darcy-Flow FNO example under "Training and Meta-Learning
  Algorithms" comparing AdamW, low-rank TensorGRaD, and low-rank + sparse
  TensorGRaD.

  ## Validation
  - `python -m py_compile examples/training/plot_tensorgrad_fno_darcy.py
  neuralop/training/tensorgrad.py neuralop/training/adamw.py neuralop/
  training/tests/test_optim.py`
  
[1] Loeschcke, S., Pitt, D., George, R., Zhao, J., Luo, C., Tian, Y., Kossaifi, J., Anandkumar, A. (2025). TensorGRaD: Tensor Gradient Robust Decomposition for Memory-Efficient Neural Operator Training. arXiv preprint, https://arxiv.org/abs/2501.02379.


neuralop/training/tensorgrad.py
```diff
@@ -0,0 +1,475 @@
...
+        else:
+            matrix = weights.data
+
+        if self.warm_restart and self.proj_tensor is not None:
+            init_factors = [factor.to(matrix.device) for factor in self.proj_tensor]
+            init_core = self.transform(init_factors, matrix)
+            init = (init_core, init_factors)
+        else:
+            init = "svd"
+        if self.activation_checkpoint:
```
Robertboy18, line 83: I think this path is broken when `activation_checkpoint=True`.

`checkpoint(tucker, matrix, rank, init)` passes `init` as a positional arg, so TensorLy treats it as `fixed_factors` instead of the `init=` kwarg. I hit:

```text
ValueError: Got fixed_factor=svd but no appropriate Tucker tensor was passed for "init".
```

There’s also a second issue in the checkpointed transform: `Transpose=True` should be `transpose=True`.

Could we add a small test for `activation_checkpoint=True` and wrap these calls so the TensorLy kwargs get passed correctly?

neuralop/training/tensorgrad.py
```diff
@@ -0,0 +1,475 @@
...
+            core, factors = checkpoint(tucker, matrix, rank, init)
+        else:
+            core, factors = tucker(matrix, rank=rank, init=init)
```
Robertboy18: Small thing: `tucker_n_iter_max` gets accepted/stored, but I don’t think it ever makes it into the actual `tucker(...)` call. Should this pass `n_iter_max=self.tucker_n_iter_max` here?

Robertboy18 (changes requested):
Nice addition overall. I pulled this locally and the new optimizer tests pass for the default paths, but I think one config path needs fixing before merge: `activation_checkpoint=True` currently crashes in the TensorLy checkpoint calls.

Main issue is that `init` is being passed positionally into `tucker`, so TensorLy sees it as `fixed_factors`; plus the checkpointed `multi_mode_dot` call uses `Transpose=True` instead of `transpose=True`.

Also noticed `tucker_n_iter_max` is exposed but not actually wired into the Tucker decomposition.

Could you add focused tests for those two options?

neuralop/training/tensorgrad.py
sebulo: Thanks for the comments.

Fixed. `tucker_n_iter_max` is now passed as `n_iter_max=self.tucker_n_iter_max` in both the checkpointed and non-checkpointed Tucker calls.

I also added a focused test covering propagation of `tucker_n_iter_max`.

neuralop/training/tensorgrad.py
sebulo: Thanks for the comments!

Fixed. The checkpointed Tucker call now passes `rank` and `init` as kwargs, so `init` is not interpreted as `fixed_factors`. I also changed the checkpointed transform call from `Transpose=True` to `transpose=True`.

Added a focused `activation_checkpoint=True` projection test.

sebulo:
Thanks for the review. I pushed fixes for both TensorGRaD projector issues:

- fixed `activation_checkpoint=True` TensorLy calls by passing Tucker args as kwargs and using `transpose=True`
- wired `tucker_n_iter_max` into Tucker decomposition
- added focused tests for `activation_checkpoint=True` and `tucker_n_iter_max`

Validation:
`conda run -n tensorgrad python -m pytest neuralop/training/tests/test_optim.py -q`

neuralop/training/tensorgrad.py
```diff
@@ -0,0 +1,475 @@
...
+        self.warm_restart = warm_restart
+        self.tucker_n_iter_max = tucker_n_iter_max
+        self.activation_checkpoint = activation_checkpoint
+        self.proj_tensor = None
+
+    def project(self, full_rank_grad, iter):
+        if self.proj_tensor is None or iter % self.update_proj_gap == 0:
+            self.proj_tensor = self.get_projection_tensor(full_rank_grad, self.rank)
+        self.proj_tensor = [
+            factor.to(full_rank_grad.device) for factor in self.proj_tensor
```
JeanKossaifi: Wouldn't we keep them on the right device? I guess it just defaults to a no-op then so not a big deal.

JeanKossaifi (commented):
This looks great, thank you @sebulo and @Robertboy18 for the review!

neuralop/training/tensorgrad.py
sebulo: Good point. I removed it to keep the projection path simpler.

sebulo (commented):
Resolved. The factors should already be on the gradient device in the normal path, so I removed the extra `.to(full_rank_grad.device)` remap to keep the projection path simpler.

JeanKossaifi:
Thank you, merging!

Issue #731, opened 2026-06-25
Aliasing induced in `resample` and problem with complex valued input for `spectralConv`


JakobDilen:
The `resample` method contains two bugs:
 - When resampling input of dimension 3 or higher, there are cases that the mode selection is off, resulting in a wraparound of the indices and inducing aliasing as a result. An example code snippet that triggers this bug is
    ```
    import torch
    import numpy as np
    from neuralop.layers.resample import resample

    # Define discretizations
    res = 15
    resample_factor = 2
    res_increased = int(res * resample_factor)
    ticks = torch.tensor(np.linspace(0, 1, res_increased, endpoint=False), dtype=torch.float32)
    X_increased, _, _ = torch.meshgrid((ticks, ticks, ticks), indexing="ij")
    X = X_increased[::resample_factor, ::resample_factor, ::resample_factor]

    # Calculate test function for both original and increased resolution
    h = (res-1)//2
    f_eval = torch.cos(2 * torch.pi * (h * X))
    f_eval_increased = torch.cos(2 * torch.pi * (h * X_increased))

    # Resample test function and check that the error is close to zero.
    f_eval_resampled = resample(f_eval.unsqueeze(0).unsqueeze(0), res_scale=1.0, axis=list(range(2, 5)), output_shape=(res_increased, res_increased, res_increased)).squeeze(0).squeeze(0)
    relative_error = (torch.linalg.norm(f_eval_increased - f_eval_resampled) / torch.linalg.norm(f_eval_increased))
    assert(relative_error < 1e-5)
    ```

 - Resampling complex valued input is not supported at the moment, as `torch.nn.functional.interpolate` cannot take complex valued tensors as input and for the Fourier interpolation for higher dimensional input the use of the `rfft` is hardcoded. An example code snippet that triggers this bug is
    ```
    import torch
    import numpy as np
    from neuralop.layers.resample import resample

    # Define discretizations
    res = 15
    resample_factor = 2
    res_increased = int(res * resample_factor)
    ticks = torch.tensor(np.linspace(0, 1, res_increased, endpoint=False), dtype=torch.float32)
    X_increased, _, _ = torch.meshgrid((ticks, ticks, ticks), indexing="ij")
    X = X_increased[::resample_factor, ::resample_factor, ::resample_factor]

    # Calculate test function for both original and increased resolution
    h = (res-1)//2
[...]

PR #732, opened 2026-06-25
Fix aliasing induced in resample and problem with complex valued input for spectralConv and add rank-1 lattice and flexible index set support

MrGoomb0:
## Summary

Adds two fixes and two features:
 - Fix: Makes sure that the `resample` function correctly indexes the fourier modes during Fourier interpolation to prevent aliasing from occurring.
 - Fix: Makes sure that complex valued functions are handled correctly by both the `resample` and `spectralConv` methods.
 - Feature: Adds the possibility for more flexibility in the index sets, e.g. a hyperbolic cross index set. 
 - Feature: Adds the possibility to use a rank-1 lattice discretization of the physical domain.

The fixes solve the bugs denoted in Issue #731 .
The features allow for the instantiation of the FNO-LAT, FNO-HC and FNO-HC-LAT models as described in *[Dilen, J., Keller, A., Kuo, F. Y., Nuyens, D. "Fourier Neural Operators with rank-1 lattice points and hyperbolic cross"
](https://arxiv.org/abs/2606.08871)*.

### Fixes

`neuralop/layers/resample.py`
- Correction in the definition of the `mode_resampling` variable, to prevent aliasing
- Adds Fourier interpolation for complex valued input.

`neuralop/layers/tests/test_resample.py`
- Add `test_resampling_odd_resolution` (and `test_resampling_even_resolution`) to test wether real valued input with an odd (respectively even) resolution is upsample correctly. The current `resample` handles the even case correctly, but contains a bug for the odd case.

- Add `test_resampling_complex_odd_resolution` (and `test_resampling_complex_even_resolution`) to test wether complex  valied input with an odd (respectively even) resolution is upsample correctly. The current `resample` does not allow resampling of complex valued input.

`neuralop/layers/spectral_convolution.py`

- Make sure that the negative Fourier frequencies are only dropped for real valued input.

`neuralop/layers/tests/test_spectral_convolution.py`

- Add `test_SpectralConv_complex_output_shape_and_dtype` and `test_SpectralConv_complex_centers_last_fft_dimension` to test wether complex valued input is handled correctly.

### Features

`neuralop/layers/index_sets.py`

- Add the `IndexSet` base class and `RadialIndexSet` extension to allow for index sets based on a given radius. Two examples of a `RadialIndexSet` are `HyperRectangleIndexSet` (which implements the index set as it is defined in the current branch), and `HyperbolicCrossIndexSet` (which is a new index set variant based on the hyperbolic cross).

`neuralop/layers/spectral_transforms.py`

- Factors out the calculation parts of the `SpectralConv` class in `neuralop/layers/spectral_convolution.py` and places them into `RegularGridFFT`. This allows for a modular implementation of the different spectral transforms e.g. `Rank1LatticeFFT`, that fits within the FNO framework with minimal changes. 

`neuralop/layers/fno_block.py` & `neuralop/models/fno.py`

- Add the named arguments `index_set` and `spectral_transform` to allow the use of different index sets as implemented in `neuralop/layers/index_sets.py` and spectral transforms as implemented in `neuralop/layers/spectral_transforms.py`. 

`neuralop/layers/lattice.py`

- Add functionality to use rank-1 lattices and to easily switch between rank-1 lattices and regular tensor grid discretizations.


JeanKossaifi:
@MrGoomb0 thank you for the PR! The bug fixes and lattice/index-set support are very useful. This currently refactors the default SpectralConv substantially and adds lattice-specific transform in FNO and FNOBlocks.

Would it make sense to introduce a dedicated LatticeSpectralConv implementing BaseSpectralConv, selected through conv_module, like I did with the SphericalConv in the SFNO? The contraction and index-set utilities could still be shared. 

It may also be worth splitting the resampling/complex-data fixes into a smaller PR separate from the lattice/index-set feature.

MrGoomb0:
@JeanKossaifi thank you for your feadback. I get that this PR refactors too much, so we opted for splitting as you suggest. I made a new PR #737 that resolves the issues and we will try to see if we can do something more in line with the SphericalConv for the lattices in a follow-up PR.

PR #734, opened 2026-07-09
 Time and parameter conditioning (1/2): precomputed-input in SpectralConv & FNOBlocks

yashpincha:
## Summary

First of a two-part split of #723, restructured per review feedback into smaller, independently reviewable changes with a clear separation of concerns.

This PR adds the layer-level machinery for time and parameter conditioning. Layers consume *precomputed* conditioning tensors only; they do not compute embeddings themselves. The embedding module (`ConditioningEmbedding`) and FNO model wiring follow in PR 2/2.

### `SpectralConv`

This adds two constructor arguments, `mode_modulation: dict | None = None` and `cond_embed_dim: int | None = None`, along with a new forward argument, `mode_embedding: Tensor | None = None`. When mode modulation is enabled, retained Fourier coefficients are scaled by a learned per-mode factor before the spectral weight contraction:

$$\hat{y}_k = W_k \cdot \left(M(e, k) \odot \hat{x}_k\right)$$

Here `e` is the conditioning embedding, `phi_k` is a learned embedding of frequency mode `k`, and `M(e, k)` is a lightweight projection of `[e, phi_k]`. Three parameterizations are supported: real, complex, and polar (phase-only via `exp(i*theta)`, which preserves magnitude). Only `phi_k` and its projection live inside `SpectralConv`, since they're specific to the layer's own mode grid.

### `FNOBlocks`

This adds two constructor arguments, `mode_modulation` and `cond_embed_dim`, both forwarded directly to the underlying `SpectralConv`, plus two forward arguments, `mode_embedding` and `mod_params: dict | None`. The `mod_params` argument holds precomputed FiLM parameters that the block simply applies — no embedding computation happens in the block. FiLM is applied around the two normalization sites, with the presence of dictionary keys selecting which modulation to apply, and only on the post-activation path. Two forms are supported: additive, $x \cdot (1 + \gamma) + \beta$, and gating, $x \cdot \sigma(g)$.

## Implementation notes

For real-valued spatial signals, the modulation factor `M(e, k)` is generally non-Hermitian. The existing DC/Nyquist imaginary-zeroing logic removes any imaginary modulation at those frequencies by design, so no special handling is needed.

## Backward compatibility

All new arguments default to `None`. With defaults, both `SpectralConv` and `FNOBlocks` behave identically to the current implementation. All changes are additive.

## Testing

The full `neuralop` suite passes unchanged (1420 passed, 56 skipped). New tests in `test_spectral_convolution.py` and `test_fno_block.py` bring the layer suite to 273 tests, covering forward-pass shapes across dimensions, all modulation parameterizations, `type_k` variants, gradient propagation, finite-difference gradient verification (the layer is affine in `x`), FiLM identity/no-op invariants, and error paths.

```bash
python3.12 -m pytest neuralop/layers/tests
```

## Open question

Naming of the conditioning forward argument: `mode_embedding` (current, consistent with `mode_modulation`) vs. `t_embedding`. Happy to rename to match reviewer preference.


neuralop/layers/fno_block.py
```diff
@@ -206,6 +208,20 @@ def __init__(
         else:
             self.non_linearity = non_linearity
 
+        # Track whether the per-mode modulation pathway is active, so forward
+        # only threads the precomputed `mode_embedding` to the conv when it is.
+        self._mode_mod_enabled = mode_modulation is not None and mode_modulation.get(
```
JeanKossaifi, line 214: Can we just have a single parameter _mode_modulation: bool? 
The conditional embedding is somewhat independent from the modulation, no?

neuralop/layers/fno_block.py
```diff
@@ -368,13 +385,38 @@ def set_ada_in_embeddings(self, *embeddings):
...
         if self.preactivation:
-            return self.forward_with_preactivation(x, index, output_shape)
+            return self.forward_with_preactivation(
+                x, index, output_shape, mode_embedding=mode_embedding
+            )
         else:
-            return self.forward_with_postactivation(x, index, output_shape)
+            return self.forward_with_postactivation(
+                x, index, output_shape,
+                mode_embedding=mode_embedding, mod_params=mod_params,
```
JeanKossaifi: I'm not super keen on `mod_params` as it is very opaque for users / devs what it does.

JeanKossaifi: I'd rather e.g. `modulation_dims: Optional[int]` that is None or the n_dims as an example

neuralop/layers/fno_block.py
```diff
@@ -389,31 +431,54 @@ def forward_with_postactivation(self, x, index=0, output_shape=None):
...
+            x_fno = self.convs[index](x, output_shape=output_shape)
 
         if self.norm is not None:
             x_fno = self.norm[self.n_norms * index](x_fno)
 
+        # Apply precomputed FiLM around the first norm. The block only applies
+        # whichever params are present; they are computed upstream.
+        mp = mod_params or {}
+        if "scale1" in mp:
+            x_fno = x_fno * (1 + mp["scale1"]) + mp["shift1"]
```
JeanKossaifi: Why not store explicitly the scale and shift and have them default to None? 
e.g. `modulation_scale=None, modulation_shift=None`?

neuralop/layers/fno_block.py
```diff
@@ -389,31 +431,54 @@ def forward_with_postactivation(self, x, index=0, output_shape=None):
...
+        if "gate1" in mp:
+            x_fno = x_fno * torch.sigmoid(mp["gate1"])
```
JeanKossaifi: I didn't realize we also had these two gates. Are they channel wise here?

neuralop/layers/spectral_convolution.py
```diff
@@ -517,9 +697,44 @@ def forward(self, x: torch.Tensor, output_shape: Optional[Tuple[int]] = None):
...
+                )
+            kept_shape = tuple(weight.shape[weight_start_idx:])
+            # Broadcast the per-sample embedding (B, cond_embed_dim) over the
+            # kept-mode grid; phi_t itself is computed upstream.
+            batch_size = mode_embedding.shape[0]
+            t_feature = mode_embedding.reshape(
+                batch_size, -1, *([1] * len(kept_shape))
+            ).expand(batch_size, -1, *kept_shape)
+            k_embed = self.embed_k(shape=kept_shape, device=x.device)
+            mod_factor = self._modulation_factor(t_feature, k_embed)
```
JeanKossaifi, line 732: Could we pass the precomputed value instead of having the computation done in the spectral conv?

JeanKossaifi (changes requested):
Thanks @yashpincha  - added some comments.

neuralop/layers/fno_block.py
yashpincha: i am not sure i see how a single bool can work: `cond_embed_dim` has to stay regardless (the modulator's input projection is sized to the embedding width, which the layer can't infer; in PR2 it's `n_params*embed_dim`); the dict is just internal plumbing. users get t_emb_FNO(..., n_params=5, type_t="power"), which builds it and threads it down? lmk what your thoughts are!

neuralop/layers/fno_block.py
yashpincha: yeah i agree the dict is too opaque. on `modulation_dims`: it can't quite stand in for `mod_params`, since that's carrying six precomputed per-channel FiLM tensors (scale/shift/gate at two sites), not an `int` dim count.

another fix in the spirit of your original separation point: move the FiLM heads into `FNOBlocks`. the block already knows its own channels, so it can have the small `e --> (scale, shift, gate)` head and just receive the conditioning embedding e? lmk

neuralop/layers/fno_block.py
yashpincha: there are two FiLM sites (around each norm) plus two gates, so scale1/shift1/gate1/scale2/shift2/gate2; six tensors. as forward args that's a six-param signature, which is why i bundled them. please see the prev [comment](https://github.com/neuraloperator/neuraloperator/pull/734#discussion_r3574540640):  with the heads moved into FNOBlocks, this resolves cleanly: the block receives just e, computes these internally, and stores + applies them under explicit names

JeanKossaifi:
I like your idea of moving the embedding into the FNO Block and add a small module e -> (scale, shift, gate)!

neuralop/layers/fno_block.py
yashpincha: they're asymmetric. scale/shift are channel-wise (width out_channels), but the two gates are width-1: a single scalar per sample, broadcast over both channels and space, not channel-wise.

neuralop/layers/spectral_convolution.py
yashpincha: the precomputed part already is passed in; that's `mode_embedding (e)`, computed once upstream. what's left in the `conv` is the part that's intrinsically tied to this conv: `\phi_k` is a function of the kept-mode grid (`weight.shape[...]`), which varies per layer and changes dynamically via the `n_modes` setter and `resolution_scaling_factor`, and it's already cached so it's computed once per grid. the `mod_factor` itself is `self.modulator([e, \phi_k])`, where the modulator is a conv-local learned parameter.

yashpincha: so i think it's the mirror of the FiLM point we just discussed: there the head depends only on channels, so it belongs in the block; here it depends on the mode grid, so it seems most natural to keep in the conv. precomputing `mod_facto`r upstream would mean the caller has to track each conv's mode geometry and own the modulator weights, which pulls those internals out of the conv. that said, if you'd prefer a consistent "everything precomputed upstream" situation across both paths, im happy to explore it; we'd just need to decide where the modulator weights and grid bookkeeping should live. thoughts?

PR #738, opened 2026-07-26
fix typos in Flattened1dConv class

kmario23:
Fixed typos to represent the actual operations (**Conv1d** instead of *Conv3d*) performed in the `Flattened1dConv` class.


JeanKossaifi:
Thanks @kmario23, looks good, merging

PR #739, opened 2026-08-14
Conditioning Refactored

yashpincha:
## Add spectral-mode and FiLM conditioning via Conditional{SpectralConv, FNOBlocks, FNO}

This PR adds two complementary conditioning mechanisms (spectral per-mode modulation and FiLM) usable together or independently. Following the pattern discussed with @JeanKossaifi , each is introduced as a subclass.
 
- ConditionalSpectralConv in `spectral_convolution.py`: a subclass of SpectralConv that multiplies the kept Fourier modes by a learned per-mode factor derived from a conditioning embedding. The factor comes from a small ChannelMLP taking a concatenation of the condition embedding and a frequency-position embedding (power-law or sinusoidal). Three modulation types: polar (phase rotation, non-Hermitian), real, and complex. The base SpectralConv forward is untouched.
- ConditionalFNOBlocks in `fno_block.py`: a subclass of FNOBlocks that fills in two inert hooks the base now exposes: `_apply_conv` routes the embedding into a `ConditionalSpectralConv` when `mode_modulation=True`, and `_condition` applies a per-layer FiLM affine `(1 + scale) * x + shift` after each norm site when `film=True`. Embedding is supplied per forward via `cond_emb` or once via `set_cond_emb`.
- ConditionalFNO in `fno.py`: FNO gains a `fno_block_module` so a subclass can pick its block class. `ConditionalFNO` uses it to build `ConditionalFNOBlocks` and to select the conv class `ConditionalSpectralConv` when `mode_modulation`, else `SpectralConv`.

### Minimal usage
```
model = ConditionalFNO(
    n_modes=(12, 12), in_channels=1, out_channels=1, hidden_channels=64,
    condition_embedding_channels=16,   # width of e; enables the FiLM heads
    mode_modulation=True,              # also routes e into ConditionalSpectralConv
)

e = embed(params)  # (B, 16)
y = model(x, cond_emb=e)
```
---
The principles behind this mechanism and mathematical formulation can be viewed [here](https://www.overleaf.com/read/tdkxstkygfhj#27a0b1).


JeanKossaifi:
Thanks @yashpincha, this is much cleaner, especially making `ConditionalSpectralConv` inherit from `SpectralConv`. We should apply the same pattern to the blocks, so ConditionalFNOBlocks inherits from FNOBlocks and ConditionalFNO from FNO.

The regular `FNO` and `FNOBlocks` remain unchanged and know nothing about conditioning.

To let `ConditionalFNO` construct `ConditionalFNOBlocks` without first creating and then overriding `self.fno_blocks`, `FNO` could accept a general block class and its arguments:

```python
class FNO:
    def __init__(
        self,
        ...,
        fno_block_module=FNOBlocks,
        fno_block_kwargs=None,
    ):
        self.fno_blocks = fno_block_module(
            ...,
            **(fno_block_kwargs or {}),
        )
```

Then the conditional model can select its block class during construction:

```python
class ConditionalFNO(FNO):
    def __init__(self, ..., condition_embedding_channels):
        super().__init__(
            ...,
            fno_block_module=ConditionalFNOBlocks,
            fno_block_kwargs={
                "condition_embedding_channels":
                    condition_embedding_channels,
            },
        )
```

This avoids modifying the regular FNO, and it avoids constructing an `FNOBlocks` instance only to overwrite it afterward.

Likewise, `ConditionalFNOBlocks` should receive the convolution class explicitly. It should not silently replace `conv_module`, since that would break models such as SFNO that use `SphericalConv`.

That seems like the clearest path to me:

- `FNO` builds `FNOBlocks`.
- `ConditionalFNO` builds `ConditionalFNOBlocks`.
- `ConditionalFNOBlocks` can build `ConditionalSpectralConv` when spectral modulation is requested.
- Existing models remain unchanged.

JeanKossaifi:
Thanks, it's such better!

I see just an issue with the change: the embedding stored with set_cond_emb. This is brittle as it breaks activation checkpointing during backward and risks reusing a previous call’s condition. I think we should pass it explicitly through the shared forward implementation and hooks:
```python
def forward(self, x, output_shape=None, *, condition_embedding):
    return self._forward_impl(
        x,
        output_shape=output_shape,
        condition_embedding=condition_embedding,
    )
```
    
I am a little torn on the arguments passed as dict (fno_block_kwargs and conv_kwargs): it's convenient and I don't see necessarily a better alternative but it make the available options harder to follow and discover. 

What about explicitly passing as arguments and simply having a factory that reroutes the right params?

Would look something like this:
```python
from functools import partial

class FNO(nn.Module):
    def __init__(self, n_modes, hidden_channels, *,
                 block_factory=FNOBlocks):
        super().__init__()
        self.fno_blocks = block_factory(
            in_channels=hidden_channels,
            out_channels=hidden_channels,
            n_modes=n_modes,
        )


class ConditionalFNO(FNO):
    def __init__(self, n_modes, hidden_channels, *,
                 condition_embedding_channels):
        block_factory = partial(
            ConditionalFNOBlocks,
            condition_embedding_channels=condition_embedding_channels,
        )
        super().__init__(
            n_modes,
            hidden_channels,
            block_factory=block_factory,
        )
```

JeanKossaifi:
Thank you for all the work on this @yashpincha!

It looks good now, I just realized we have two breaking changes: 

1. ConditionalFNO fails with our Trainer: 

    The trainer passes the whole batch (including target y), but the new forward signature rejects it:
   ```TypeError: ConditionalFNO.forward() got an unexpected keyword argument 'y'```
   
   Can we keep the required embedding and restore FNO’s handling of extra batch arguments:
   ```def forward(self, x, output_shape=None, *, condition_embedding, **kwargs):```
   
   We could also add a small unit test of a Trainer with x, y, and condition_embedding.
   
2. Two small failure cases: 

    odd sinusoidal embedding sizes cause a channel mismatch, and blocks[i] cannot forward the condition. Would be good to validate/support odd sizes and update the indexed-block wrapper.

Issue #720, opened 2026-04-13
PINO Reproducibility Issues Between GPU Architectures


ryanliu30:
I have seen a large discrepancy between identical PINO training running on 4090 + torch 2.6.0 and 5090 + torch 2.8.0. After some investigation, I found the cause to be the tensor core optimization that cuDNN uses. For my case, I was able to fix this (and gain substantial improvement in accuracy) by setting:
```
torch.backends.cudnn.allow_tf32 = False
```
for torch < 2.9.0, and 
```
torch.backends.fp32_precision = "ieee"
```
for torch >= 2.9.0

JeanKossaifi:
Thank you - what setting did you find this in? You would loose out quite a bit in speed by turning off tf32 and I would only expect it to make a different for high precision (which I guess you would need for the PDE losses).

ryanliu30:
I was training PINO on the 1+1-D KS equation, and the root cause of the issue is that the 1x1 convolution layers in FNO dispatch to the cuDNN backend and will use tf32 by default. In a high-precision setting like PINO, this will make a big difference. I have personally benchmarked this on 4090/5090, and disabling tf32 did make a big difference for training throughput since it only affects the 1x1 convolutions.

PR #742, opened 2026-09-03
Fix TF32 settings for reproducible PINO training

abhs21:
This adds an explicit `allow_tf32` option to the shared training setup and routes the Burgers PINO script through it. The option defaults to `True` in both the shared and Burgers PINO configurations, preserving existing behavior; users can set it to `False` when IEEE float32 is needed.

On newer PyTorch versions, changing only the global precision setting does not override an existing backend setting. The helper sets the global, CUDA matmul, and cuDNN precision controls explicitly. Older versions use the legacy flags, without mixing the two APIs.

Validation of the review changes on Python 3.12.5 / PyTorch 2.14.0:
- Five focused precision-configuration tests passed, covering both precision APIs and invalid input.
- A configuration smoke check confirmed the `True` default and explicit `False` opt-out.
- The string-executing subprocess test was removed.
- `git diff --check` passed.

Broader local configuration-test collection is blocked by the existing missing `config.datasets` module. CUDA execution and a matched PINO training comparison across GPUs remain unverified; these checks validate configuration behavior, not cross-GPU training reproducibility.

Related to #720.

### DIFF
--- a/neuralop/training/tests/test_torch_setup.py
+++ b/neuralop/training/tests/test_torch_setup.py
@@ -1,0 +1,54 @@
+from types import SimpleNamespace
+
+import pytest
+
+from neuralop.training import torch_setup
+
+
+class LegacyTorch:
+    def __init__(self):
+        self.precision = None
+        self.backends = SimpleNamespace(
+            cuda=SimpleNamespace(matmul=SimpleNamespace(allow_tf32=None)),
+            cudnn=SimpleNamespace(allow_tf32=None),
+        )
+
+    def set_float32_matmul_precision(self, precision):
+        self.precision = precision
+
+
+class NewTorch:
+    def __init__(self):
+        self.backends = SimpleNamespace(fp32_precision="none")
+
+
+@pytest.mark.parametrize(
+    ("allow_tf32", "expected_precision"),
+    [(False, "highest"), (True, "high")],
+)
+def test_set_tf32_legacy_api(monkeypatch, allow_tf32, expected_precision):
+    fake_torch = LegacyTorch()
+    monkeypatch.setattr(torch_setup, "torch", fake_torch)
+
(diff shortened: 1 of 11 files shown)


neuralop/training/tests/test_torch_setup.py
```diff
@@ -0,0 +1,102 @@
...
+    assert fake_torch.backends.cudnn.conv.fp32_precision == expected_precision
+    assert fake_torch.backends.cudnn.rnn.fp32_precision == expected_precision
+
+
+def test_set_tf32_rejects_non_boolean_values():
+    with pytest.raises(TypeError, match="allow_tf32 must be a bool"):
+        torch_setup.set_tf32("false")
+
+
+def test_set_tf32_overrides_existing_backend_settings():
```
JeanKossaifi, line 71: This part is a little odd - can we avoid executing a string? It will be hard to debug and we could easily end up with inserted, problematic code. 

DO we need all of these unit-tests?

JeanKossaifi (changes requested):
Thank you for the PR @abhs21, this seems reasonable to me. Let's make the new option just true by default and optionally turn it off so that there won't be any surprises for existing users. Would be good to remove or adapt the unit-test that is executed from the str.

neuralop/training/tests/test_torch_setup.py
abhs21: Addressed in 55735b4: removed the string-executing subprocess test and its imports. The remaining five focused tests cover the legacy and newer precision APIs plus invalid input, and all five pass. I also made the Burgers PINO option default to `True`, as requested in the review; setting `allow_tf32=False` remains supported. The configuration smoke check confirms both the default and opt-out.

The new workflows are awaiting maintainer approval.

PR #743, opened 2026-09-04
Use native segment reduction during compilation

### LINKED ISSUE #733: Proposal: Replace torch_scatter.segment_csr with native torch.segment_reduce in GINO to allow compile and remove dependency.
Hi everyone, 

thank you guys again for still maintaining this repository, it is immensely helpful for my research. 

GINO models currently cannot be compiled using `torch.compile(model)` due to the external `torch_scatter.segment_csr` function. Since (I believe) torch 2.11, there is `torch.segment_reduce` (or more specialized scatter functions), which would be a torch-native alternative to `segment_csr`, which can also be used sparsely using the _offset_ parameter as an index pointer.

This issue is also discussed in the torch_scatter repo (https://github.com/rusty1s/pytorch_scatter/issues/440), where the maintainers recommend either disabling compilation for the specific operation or using the PyG utils "scatter", which dynamically switches between `torch_scatter` and native `torch`, depending on the device and whether or not the model is compiled. This might align with the current idea of switching the backend to PyG https://github.com/neuraloperator/neuraloperator/issues/726 .

Aside from keeping everything as is, the options would be 
1. Set `@torch.compiler.disable` over `def segment_csr` in _neuralop/layers/segment_csr_ (least invasive change).
2. Switch to native `torch.segment_reduce`, which is slower on CPU, but can be compiled (removes `torch_scatter` dependency).
3. Use the dynamic solution that PyG suggests, either implemented by ourselves or imported from PyG.

I am happy for any feedback, maybe I overlooked a crucial detail. If not, I can also provide some experiments or make a pull request with any of the options.

Best, 
Josh

P.s.: Minimal example for completeness (ran on an A100 with Cuda 12.9, neuraloperator 2.0.0, torch 2.12.0, torch-scatter 2.1.2)

```
import torch
from neuralop.models import GINO

device = "cuda"
grid = 16
model = GINO(
    in_channels=1, out_channels=1,
    gno_coord_dim=2,
    fno_n_modes=(8, 8),  # length must equal gno_coord_dim
    in_gno_radius=0.3, out_gno_radius=0.3,
)
model = model.to(device=device)

n_in, n_out = 25, 10
input_geom = torch.rand(1, n_in, 2, device=device)
x = torch.rand(1, n_in, 1, device=device)

lin = torch.linspace(0, 1, grid, device=device)
gx, gy = torch.meshgrid(lin, lin, indexing="ij")
latent_queries = torch.stack([gx, gy], dim=-1).unsqueeze(0)

output_queries = torch.rand(1, n_out, 2, device=device)

out = model(input_geom, latent_queries, output_queries, x=x)
print(f"Uncompiled forward works: {out.sum()}")

model.compile()

try:
    out_compiled = model(input_geom, latent_queries, output_queries, x=x)
except Exception as e:
    print("Forward has thrown an error here")
    print("\n\n\n")
    print(e)
```

abhs21:
Fixes #733

## Summary
- use native `torch.segment_reduce` when compiling or when scatter-free execution is requested
- preserve the eager `torch_scatter` path
- preserve zero output for empty mean neighborhoods
- document the backend selection behavior

## Validation
- 93 targeted tests passed locally

### DIFF
--- a/neuralop/layers/gno_block.py
+++ b/neuralop/layers/gno_block.py
@@ -80,7 +80,9 @@ class GNOBlock(nn.Module):
         Default None, by default None
     use_torch_scatter_reduce : bool, optional
         Whether to use torch-scatter to perform grouped reductions in the IntegralTransform.
-        If False, uses native Python reduction in neuralop.layers.segment_csr, by default True
+        If False, uses native PyTorch segment reduction when available, by default True.
+        Native reduction is also selected automatically while tracing with
+        ``torch.compile``.
 
         .. warning::
 
--- a/neuralop/layers/integral_transform.py
+++ b/neuralop/layers/integral_transform.py
@@ -58,7 +58,9 @@ class IntegralTransform(nn.Module):
         by default independently of this parameter, by default 'linear'
     use_torch_scatter : bool, optional
         Whether to use torch-scatter to perform grouped reductions in the IntegralTransform. 
-        If False, uses native Python reduction in neuralop.layers.segment_csr, by default True
+        If False, uses native PyTorch segment reduction when available, by default True.
+        Native reduction is also selected automatically while tracing with
+        ``torch.compile``.
 
         .. warning:: 
 
--- a/neuralop/layers/segment_csr.py
+++ b/neuralop/layers/segment_csr.py
@@ -5,6 +5,36 @@ import torch
 from torch import einsum
 
 
+def _is_compiling():
+    """Return whether this function is being traced by ``torch.compile``."""
+    compiler = getattr(torch, "compiler", None)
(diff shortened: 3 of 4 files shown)


abhs21:
CI follow-up (22 September): I checked the failed Python 3.9 job. It finished with 1,453 passed, 56 skipped, and one failure: test_incremental receives HTTP 504 while requesting Zenodo record 12784353.

Job log: https://github.com/neuraloperator/neuraloperator/actions/runs/33824966915/job/104486889481

The eight focused segment-reduction tests also pass locally. Could a maintainer rerun the failed job when convenient? I do not have permission to restart this upstream workflow.

PR #744, opened 2026-09-04
Keep native neighbor CSR counts exact

### LINKED ISSUE #740: native_neighbor_search returns inconsistent neighbors_index / neighbors_row_splits above 2**24 neighbor pairs
`neuralop.layers.neighbor_search.native_neighbor_search` builds its two outputs by different arithmetic: `neighbors_index` comes from an exact `nonzero()`, while `neighbors_row_splits` is accumulated in **float32** and only cast to `int64` afterwards. float32 represents integers exactly only up to `2**24` (16,777,216), so once a search returns more pairs than that, the splits stop agreeing with the index list they are supposed to describe.

The two are consumed together in `IntegralTransform.forward`, so the inconsistency surfaces as a shape error in `torch.cat`, far from its cause and only on larger inputs:

```
File "neuralop/layers/integral_transform.py", line 179, in forward
    agg_features = torch.cat([rep_features, self_features], dim=-1)
RuntimeError: Sizes of tensors must match except in dimension 1.
             Expected size 55436424 but got size 55436432 for tensor number 1 in the list.
```

`neighbors_row_splits[-1]` is documented as "the total number of neighbors", and downstream code relies on that: `IntegralTransform.forward` derives `num_reps` from the splits and `repeat_interleave`s the query points by it, so any drift makes `self_features` a different length from `rep_features`.

Affects 2.0.0 and current `main`. Only the native PyTorch path; the open3d `FixedRadiusSearch` branch returns exact integer counts and is unaffected.

## Reproduction

No GPU and no large mesh required. Coincident 1-D points make the pair count large cheaply. The total must be **odd**, so that it is not representable in float32 at that magnitude (the spacing there is 2) — with an even total the float accumulation can land on the correct answer by luck and the bug hides:

```python
import torch
from neuralop.layers.neighbor_search import native_neighbor_search

# 4201 x 4199 coincident points, all within the radius -> 17,639,999 pairs (odd)
data, queries = torch.zeros(4201, 1), torch.zeros(4199, 1)
out = native_neighbor_search(data, queries, radius=1.0)

print("true pairs               :", 4201 * 4199)
print("neighbors_index length   :", out["neighbors_index"].numel())
print("neighbors_row_splits[-1] :", int(out["neighbors_row_splits"][-1]))
```

Observed on neuralop 2.0.0 / torch 2.12.0+cu130, CPU, about a second:

```
true pairs               : 17639999
neighbors_index length   : 17639999
neighbors_row_splits[-1] : 17640000     <-- one more than the indices it describes
```

Below `2**24` pairs the same call is consistent, which is why smaller meshes never show this.

## Cause

In `native_neighbor_search`:

```python
nbr_indices = dists.nonzero()[:,1:].reshape(-1,)              # exact
...
in_nbr = torch.where(dists > 0, 1., 0.,)                      # float32 indicator
nbrhd_sizes = torch.cumsum(torch.sum(in_nbr, dim=1), dim=0)   # float32 sum AND cumsum
splits = torch.cat((torch.tensor([0.]).to(queries.device), nbrhd_sizes))
nbr_dict["neighbors_row_splits"] = splits.long()              # cast after the precision is gone
```

The running total is accumulated in float32 and quantizes once it exceeds `2**24`. The `.long()` then faithfully converts an already-incorrect value.

## Fix suggestion

The neighbor counts are integers, so accumulating them in an integer dtype rather than float32 would make the splits exact at any size: the indicator can be summed as a boolean/integer count before the cumulative sum, leaving the neighbour set, its ordering and the `return_norm` weights untouched. Casting later, as the current code does, cannot recover precision already lost in the accumulation.

## Side note
`native_neighbor_search` materializes the full `[m, n]` distance matrix (and, as written, a few copies of it) on every forward with no caching, so a GNO over a large mesh allocates it repeatedly. That is a performance concern rather than a correctness one, but it is part of what makes the >`2**24` regime easy to reach in practice.

abhs21:
Fixes #740

## Summary
- accumulate native neighbor counts in int64 instead of float32
- keep `neighbors_index` and `neighbors_row_splits` consistent above 2**24 pairs
- add a regression with 4,097 × 4,097 coincident points

## Validation
- 2 neighbor-search tests passed
- 85 GNO/GINO downstream tests passed

### DIFF
--- a/neuralop/layers/neighbor_search.py
+++ b/neuralop/layers/neighbor_search.py
@@ -109,9 +109,18 @@ def native_neighbor_search(
     if return_norm:
         weights = dists[dists.nonzero(as_tuple=True)]
         nbr_dict["weights"] = weights **2 # weighting function computed on squared norms
-    in_nbr = torch.where(dists > 0, 1., 0.,)
-    nbrhd_sizes = torch.cumsum(torch.sum(in_nbr, dim=1), dim=0) # num points in each neighborhood, summed cumulatively
-    splits = torch.cat((torch.tensor([0.]).to(queries.device), nbrhd_sizes))
+    # Keep neighborhood counts in an integer dtype. Accumulating them in
+    # float32 loses exactness once the total exceeds 2**24, which can make
+    # neighbors_row_splits disagree with neighbors_index.
+    nbrhd_sizes = torch.cumsum(
+        torch.sum(dists > 0, dim=1, dtype=torch.long), dim=0
+    )
+    splits = torch.cat(
+        (
+            torch.zeros(1, dtype=torch.long, device=queries.device),
+            nbrhd_sizes,
+        )
+    )
     
     nbr_dict["neighbors_index"] = nbr_indices.long().to(queries.device)
     nbr_dict["neighbors_row_splits"] = splits.long()
--- a/neuralop/layers/tests/test_neighbor_search.py
+++ b/neuralop/layers/tests/test_neighbor_search.py
@@ -53,3 +53,21 @@ def test_fallback_nb_search():
 
     return_dict = compute_norm_separate(return_dict, coords, coords)
     print(return_dict["squared_norm"])
+
+
+def test_large_neighbor_counts_keep_csr_consistent():
+    # This creates just over 2**24 coincident pairs, the point where float32
+    # cumulative counts can no longer represent every integer exactly.
+    n_points = 4097
+    data = torch.zeros(n_points, 1)
+    queries = torch.zeros(n_points, 1)
+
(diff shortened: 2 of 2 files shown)


abhs21:
CI follow-up (22 September): I checked the failed Python 3.9 job. It finished with 1,448 passed, 56 skipped, and two failures: test_DarcyDatasetDownload[16] and test_NSDownload both receive HTTP 504 from Zenodo.

Job log: https://github.com/neuraloperator/neuraloperator/actions/runs/33824969403/job/104872603270

Both focused neighbor-search tests also pass locally. Could a maintainer rerun the failed job when convenient? I do not have permission to restart this upstream workflow.

PR #745, opened 2026-09-04
Accept alternate Navier-Stokes archive names and explicit PT paths

abhs21:
The Navier-Stokes archive files reported in #708 use names such as `nsforcing_1024_train.pt`, while the loader expects `nsforcing_train_1024.pt`. With no path override, this patch accepts either filename order. It checks both before deciding to download and resolves the names again after extraction; the existing standard name takes precedence when both exist.

The optional `train_path` and `test_paths` arguments remain available for custom locations. Supplying only one split still lets the other split download.

Validation on local Python 3.12 and PyTorch 2.14.0+cpu: five focused tests passed. Two new regression cases cover alternate names already on disk and names produced by a replacement downloader; three cover explicit and partial path overrides. `git diff --check` passed. The large live Zenodo archives were not downloaded for this check. New GitHub workflows await maintainer approval.

Related to #708.

### DIFF
--- a/neuralop/data/datasets/navier_stokes.py
+++ b/neuralop/data/datasets/navier_stokes.py
@@ -1,7 +1,7 @@
 import logging
 import os
 from pathlib import Path
-from typing import Union, List
+from typing import Union, List, Optional
 
 from torch.utils.data import DataLoader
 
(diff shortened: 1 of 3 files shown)


abhs21:
The current PR lets callers pass the alternate filenames reported in #708 explicitly, and both CI workflows pass. It does not change the default filename lookup when those arguments are omitted. Would a maintainer prefer an automatic fallback for the archive’s `nsforcing_{resolution}_{split}.pt` names? I can adjust the patch if that better addresses the issue.

PR #747, opened 2026-09-04
Validate spherical SWE dataset indices

### LINKED ISSUE #328: Handle `index` in `SphericalSWEDataset::__getitem__` correctly.
The `SphericalSWEDataset` class currently implements the `__getitem__` as follows:

```py
   def __getitem__(self, index):

        with torch.inference_mode():
            with torch.no_grad():
                inp, tar = self._get_sample()

                if self.normalize:
                    inp = (inp - self.inp_mean) / torch.sqrt(self.inp_var)
                    tar = (tar - self.inp_mean) / torch.sqrt(self.inp_var)

        return {'x': inp.clone(), 'y': tar.clone()}
```

which ignores the index completely. This allows to access the dataset with any arbritary key:

```py
dataset = test_loader.dataset
dataset['foo'] # returns a sample
dataset['anotherkey'] # returns another sample
```

Expected behaviour is: Just not be able to use keys alltogether or receive a KeyError or something like that.

abhs21:
Fixes #328

Validates that `SphericalSWEDataset` receives integer indices and raises `IndexError` outside `range(len(dataset))`.

Validation on Python 3.12.13 and PyTorch 2.14.0+cpu:
- All 6 index-validation tests passed with the real `torch_harmonics` 0.7.4 package installed; its solver import was not stubbed.
- The tests cover valid integer access, invalid index types, and out-of-range indices. They replace sample generation with small tensors to isolate indexing behavior.
- Compilation and diff checks passed.

The newer torch_harmonics 0.9.2 wheel failed to import with this PyTorch build because of an undefined binary symbol, so this result is specific to the working 0.7.4 environment.

### DIFF
--- a/neuralop/data/datasets/spherical_swe.py
+++ b/neuralop/data/datasets/spherical_swe.py
@@ -1,4 +1,5 @@
 from math import ceil, floor
+from operator import index as operator_index
 
 import torch
 from torch.utils.data import DataLoader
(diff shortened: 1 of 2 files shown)


abhs21:
This addresses the invalid-key behavior in #328, and both CI workflows are passing. When a maintainer has time, could you review whether the index-type and range checks match the intended dataset API? I can follow up on any requested changes.

